PR #929, opened 2015-01-03
Allow utf-8 page names when not using grit. Resolves #830 (when using rugged adapter).

### LINKED ISSUE #830: Encoding::CompatibilityError at search Japanese words
when trying to query words with specific Japanese character such as "トランス", "ト" or  "ミ", gollum throws  bellow the exception.
```
Encoding::CompatibilityError at /search
incompatible character encodings: UTF-8 and ASCII-8BIT
file: template.rb location: render line: 39
```

```
ruby 1.9.3p545 (2014-02-24 revision 45159) [x86_64-linux]
```

But the query  of "ア” (http:localhost:4557/search?q=ア) do not throws the Encoding error,  I'm complicated.

dometto:


lib/gollum/app.rb
```diff
@@ -22,6 +22,13 @@
 # Fix to_url
 class String
   alias :upstream_to_url :to_url
+
+  unless defined?(Grit)
```
bartkamphorst: Please don't introduce a reference to the Grit class in gollum. How about checking which git adapter gollum is using?

dometto: That's not possible at the moment: although we can check whether `GIT_ADAPTER != 'grit'`, that doesn't ensure that gollum-lib will not fall back on grit anyway (when `GIT_ADAPTER` cannot be required). See [here](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib.rb#L9-L13).

bartkamphorst: > that doesn't ensure that gollum-lib will not fall back on grit anyway (when GIT_ADAPTER cannot be required)

Well, it looks like that code will throw a LoadError if `Gollum::GIT_ADAPTER` is defined but requiring the corresponding adapter fails. But you are right that if `GIT_ADAPTER` isn't defined, then grit will be used (for now). But what about 
```ruby
unless ! is_defined?(Gollum::GIT_ADAPTER) || Gollum::GIT_ADAPTER == 'grit'
```

dometto: :+1:

dometto:
@bartkamphorst: pushed the change.

PR #949, opened 2015-01-22
M: update readme, set wiki_options to allow editing in rackapp

milligramme:
by defaut, mount in Rack app disallowed editing
so, add `:allow_editing` option to Rack example in README


dometto:
Thanks for chipping in @milligramme! However, I thought/hoped that https://github.com/gollum/gollum/commit/09149592b55d4a0e87296bb8a3acc3f9015a6c79 would have taken care of this problem. If you use the latest gollum (`3.1.1`) explicitly setting `@allow_editing` should be unnecessary. Have you tried that?

dometto:
@milligramme, sorry for the confusion. I tested and noticed that the issue was not in fact fixed. See PR #950. I just merged the fix and released a new version of the gem that doesn't require setting the `:allow_editing` option to true manually.

Issue #911, opened 2014-12-03
"allow_editing" is false by default?


bennibu:
When I upgraded to v3.1.0 I had to set the wiki_option `allow_editing` to true to re-enable editing. We use gollum as a rack application with the following config.ru:

``` ruby
....
Precious::App.set(:wiki_options, {
    universal_toc: false,
    allow_editing: true,
    live_preview: false,
    allow_uploads: false
  })

run Precious::App
```

If so, please update the README.md or HISTORY.md.

dometto:
It turns out that this was not resolved completely by 0914959  (read all about it [here](https://github.com/gollum/gollum/pull/950)). I just merged and released the proper fix in `v3.1.2`.

arr2036:
Thanks for keeping this issue updated :)

PR #950, opened 2015-01-22
Set the allow_editing wiki option to true by default.

dometto:
There was a bug when mounting gollum as rack app: the new `:allow_editing` option in the wiki's option hash would default to false. See https://github.com/gollum/gollum/issues/911. However, https://github.com/gollum/gollum/commit/09149592b55d4a0e87296bb8a3acc3f9015a6c79 did not fix the issue completely. More specifically, it fixed it for GET requests, but not for POST requests.

This is because my previous fix only made the `@allow_editing` instance variable, and not the corresponding key in the options hash, default to true. So [this line](https://github.com/gollum/gollum/blob/master/lib/gollum/editing_auth.rb#L10) would stop any POST request.

My apologies. ;)

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -102,7 +102,8 @@ module Precious
       @css = settings.wiki_options[:css]
       @js  = settings.wiki_options[:js]
       @mathjax_config = settings.wiki_options[:mathjax_config]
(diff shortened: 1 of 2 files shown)


dometto:
Unfortunately we miss tests for the rack-mount use case. (And my manual test of the fix was not thorough enough) Perhaps having a live instance of gollum to run tests again could be a solution here.

Issue #952, opened 2015-01-25
Left scrollbar in left preview unselectable in firefox


simonzack:
Unselected in firefox in both windows & linux. On windows, selecting the very left edge works, but not the full scrollbar. I think this is due to the overlap between the left and right panes.

PR #953, opened 2015-01-25
pixel tweaks so the left scrollbar is selectable and the viewport takes the whole page

simonzack:
Happens on firefox for both windows & linux. See #952.

Before
before

After
after

### DIFF
--- a/lib/gollum/public/gollum/livepreview/js/livepreview.js
+++ b/lib/gollum/public/gollum/livepreview/js/livepreview.js
@@ -453,18 +453,16 @@ var applyTimeout = function () {
     var height = $( win ).height();
     var heightHalf = height / 2;
 
-    // height minus 50 so the end of document text doesn't flow off the page.
-    // + 15 for scroll bar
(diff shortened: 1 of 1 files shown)


lib/gollum/public/gollum/livepreview/js/livepreview.js
```diff
@@ -453,18 +453,17 @@ var applyTimeout = function () {
     var height = $( win ).height();
     var heightHalf = height / 2;
 
-    // height minus 50 so the end of document text doesn't flow off the page.
+    // height minus 40 so the end of document text doesn't flow off the page.
     // + 15 for scroll bar
```
dometto: This seems to be no longer applicable. :) Would be wise to remove the comment to avoid confusion.

dometto:
:+1: Thanks for spotting the problem and fixing. Please remove the comment I indicated and I'll merge (also, please rebase into one commit).

simonzack:
Sure, I've removed that now and merged into one commit.

dometto:
ty!

Issue #957, opened 2015-01-27
Expose editor to allow configuring the live editor.


simonzack:
Can the `editor` and `editorSession` variables be exposed in window to allow the live editor to be configured? In the code I see that `$.editorSession` is currently exported "for testing".

Another approach I can think of would be to fire some event or call some function in `initAce`. This has the advantage of killing any flickers when setting a different theme.

dometto:
As I said in the PR this seems reasonable to me. Note that we're discussing moving away from ACE in #826, the idea being to integrate live preview into gollum's default editor. If you have any thoughts on that please let us know.

dometto:
Resolved by #965.

PR #958, opened 2015-01-27
added an initAce hook to allow configuration of the live editor

simonzack:
See #957. I've used a function here instead of an event so the editor can be configured before it's initialized.

The theme is in the else block to avoid flickering between black & white, if a light theme is chosen.


dometto:
Seems reasonable to me, but I don't know much about JS. @sunny?

sunny:
Hey @simon. I dont see any code in the PR anymore, but the idea sounds good.

Beware that we are thinking about moving away from ACE. If you feel strongly against it we would like to hear about your use for it.

simonzack:
@sunny I changed the branch to submit another PR, that must've cleared the code, can't seem to edit the branches so I submitted another one at #965.

I don't really mind whether ACE is included or not, but I really need markdown syntax highlighting.

PR #959, opened 2015-01-27
Updated ace so markdown highlights better

simonzack:
I've updated ace to get it's better markdown highlighting since 17 Jun 2012 where it was last copied here. It's able to highlight inline code in bullet points.

There's the plan to possibly drop ace but I think this would still be helpful in the meantime.


dometto:
I think it's a good idea to update ACE. As we discussed, we might have to keep it in anyway for syntax highlighting. Would merging this into `master` break backwards compatibility in any way?

simonzack:
I don't think so, I've been using it for the past hour without any problems.

dometto:
What browser are you on? I'll test a bit with other browsers later. (Also curious whether this makes ACE compatible with Safari)

simonzack:
I'm on firefox 35. Haven't tested it with other browsers. Compatibility with more browsers would be nice indeed.

dometto:
My findings:
* On Firefox, [fenced code blocks don't work](https://cloud.githubusercontent.com/assets/147111/5920308/0dd6cf56-a63c-11e4-8d4f-5aa8ad9f830b.png). However, this is already the case on `master`, so this update doesn't break it. 
* Same on Safari, except with the ACE update, we get a [nicer result](https://cloud.githubusercontent.com/assets/147111/5920326/34fc5178-a63c-11e4-95eb-a5b6fc42d1eb.png).
* Chrome does the same as Safari with the ACE update

Otherwise I can't find anything broken (also not on Safari), so I'll merge this later.

Any thoughts on how to fix fenced blocks (especially the weird way Firefox renders them) are very welcome.

simonzack:
> Any thoughts on how to fix fenced blocks (especially the weird way Firefox renders them) are very welcome.
Not sure, it works on my firefox screenshot. I've tested on both windows & linux.

dometto:
Strange, must be a FF on OS X issue then.

dometto:
Congratulations, you are now a [top contributor](https://github.com/gollum/gollum/graphs/contributors?from=2010-03-28&to=2015-01-27&type=a) ;)

simonzack:
That's a nice side-effect :)

PR #965, opened 2015-01-28
added an initAce hook to allow configuration of the live editor

simonzack:


sunny:
Thanks @simonzack! :sparkles:

PR #973, opened 2015-02-04
Add option to configure PlantUML endpoint

hsanson:
Allow users to change the PlantUML server endpoint when running gollum via a new --plantuml-url option. This requires gollum-lib rc version to work.

I guess is better to wait the next release of gollum-lib with PlantUML support and modify the Gemfile in gollum to use that version before this gets merged.

Is there a roadmap or an estimate on when the next gollum-lib version with PlantUML support is to be released?


dometto:
Restarted the travis jobs and everything is green, but can't get the lights to update in this PR.

dometto:
...just merging then. Thanks @hsanson!

PR #985, opened 2015-03-13
Fixed two broken links in README.md

uraimo:
Both working now, the textile site was shut down a few years ago, better refer to wiki for format details.


dometto:
Thanks!

Issue #984, opened 2015-03-12
wiki_options[:base_path] setting ignored


craibuc:
It appears that the `wiki_options[:base_path]` setting is ignored when starting Gollum as a Rack app.

Starting Gollum via command:

`$ gollum --config config.rb --allow-uploads page --user-icons identicon --base-path foo/bar`

Gollum served at `http://localhost:4567/foo/bar/`

Starting Gollum via Rackup:

`$ rackup -p 4567`

Gollum served at `http://localhost:4567/`

config.ru:

```
#!/usr/bin/env ruby
require 'rubygems'
require 'gollum/app'

wiki_options = {}
wiki_options[:universal_toc] = false
wiki_options[:base_path] = 'foo/bar' # Equivalent to --base-path
wiki_options[:allow_editing] = true # # Equivalent to --no-edit
wiki_options[:live_preview] = true # Equivalent to --live-preview
wiki_options[:allow_uploads] = true # Equivalent to --allow-uploads
wiki_options[:per_page_uploads] = true # When :allow_uploads is set, store uploads under a directory named after the page, as when using --allow-uploads page
wiki_options[:user_icons] = "identicon" # Equivalent to --user-icons ; gravatar, identicon, none. Default: none.
wiki_options[:show_all] = true # Equivalent to --show-all
wiki_options[:collapse_tree] = true # Equivalent to --collapse-tree
wiki_options[:h1_title] = true # Equivalent to --h1-title

gollum_path = File.expand_path(File.dirname(__FILE__)) # CHANGE THIS TO POINT TO YOUR OWN WIKI REPO

Precious::App.set(:gollum_path, gollum_path)
Precious::App.set(:default_markup, :asciidoc) # set your favorite markup language
Precious::App.set(:wiki_options, wiki_options)
run Precious::App
```

Using Gollum 3.1.2.

bartkamphorst:
Thanks for reporting the issue @craibuc . We're pretty swamped at the moment, so I can't give you a timeline for a fix. But if you could help out by investigating the cause, that would certainly speed things up!

dometto:
@craibuc, it seems that base_path is unavailable as an option when running gollum as a rack app. This is because the base_path option is implemented via a simple [piece of middleware](https://github.com/gollum/gollum/blob/master/bin/gollum#L209-L229) when running gollum via `bin/gollum`. You'll have to add the same middleware to your `config.ru` to get the desired result.

I'll update the wiki accordingly.

[...]

PR #988, opened 2015-03-27
adding template dir logic to app.rb vs. just the bin

andrewarrow:
I'm running gollum from a `config.ru` file:

    Precious::App.set(:wiki_options, {:universal_toc => false, :template_dir => 'templates'})
    run Precious::App

and without this fix, it never reads my custom templates directory.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -96,6 +96,7 @@ module Precious
     end
 
     before do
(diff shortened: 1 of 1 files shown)


dometto:
Thanks @andrewarrow! Looks good to me.

bartkamphorst:
Thanks @andrewarrow . If you have time, would you consider producing a similar fix for this [similar issue](https://github.com/gollum/gollum/issues/984)?

PR #992, opened 2015-04-05
prepend baseUrl to /uploadFile

kirat-singh:
Hi folks,

Uploading a file into the editor fails when the gollum wiki is running with base_path != '/'.   The fix is pretty simple.  By the way, I noticed that there is a post to /wiki/help in this same js file.  I'm not sure what that's for, but I imagine it needs a similar treatment.

best,
-Kirat


dometto:
Thanks!

PR #995, opened 2015-04-06
Remove extraneous li tags

tfogo:
I found a couple of unclosed `li` tags that appear to do nothing (correct me if I'm wrong).

### DIFF
--- a/lib/gollum/templates/page.mustache
+++ b/lib/gollum/templates/page.mustache
@@ -41,10 +41,8 @@ Mousetrap.bind(['e'], function( e ) {
       {{/editable}}
     {{/allow_editing}}
     {{#page_exists}}
-    <li class="minibutton jaws">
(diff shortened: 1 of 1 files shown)


SkyCrawl:
This is one of Gollum's many puzzling features :). Some time ago, I stumbled upon this as well when refactoring Gollum to be somewhat better and thought it had no meaning... Well, eventually it turned out it HAD a meaning, after I refactored all of this away :p. So it was hidden very craftily!

In any case, **DON'T MERGE THIS**.

Some pieces you didn't notice:
* https://github.com/gollum/gollum/blob/master/lib/gollum/public/gollum/css/editor.css
* https://github.com/gollum/gollum/blob/master/lib/gollum/public/gollum/javascript/gollum.js

SkyCrawl:
CSS should prevent them from being displayed but yes, they do seem to be pointless :).

dometto:
Merging then :) Thanks @tfogo!

PR #1006, opened 2015-05-05
Fix basepath redirect

grigory51:
If run gollum with custom basepath, then in Location header no present schema, host and port.
Browser change location think basepath as host

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -206,9 +206,9 @@ else
 
     class MapGollum
       def initialize base_path
-        @mg = Rack::Builder.new do
+        @mg = Rack::Builder.new do |env|
           map '/' do
-            run Proc.new { [302, { 'Location' => "/#{base_path}" }, []] }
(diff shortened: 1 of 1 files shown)


SkyCrawl:
> If run gollum with custom basepath, then in Location header no present schema, host and port.

Well, this is probably very much by design. An exerpt from [here](http://en.wikipedia.org/wiki/HTTP_location):
> The IETF HTTP working group found that the most popular web browsers tolerate the passing of a relative URL and, consequently, the updated HTTP 1.1 specifications (IETF RFC 7231) relaxed the original constraint, allowing the use of relative URLs in Location headers.

Relative URLs are surely the easiest and probably also the safest way to go. If a browser has a problem with them, I'd rather advise to redirect this issue to its developers :).

> Browser change location think basepath as host

I never noticed this with Chrome and Firefox, at the very least. Since relative URLs in Location header are allowed in HTTP 1.1 specification... which browser (and version) did you have in mind?

dometto:
Closing this since no response and tests are failing.

PR #1014, opened 2015-05-20
Fix #1012: /Pages View should list folders first, then files, all alphabetically

### LINKED ISSUE #1012: All Pages should sort Folders first, then Pages
Like a file manager, the All Pages view should list Folders alphabetically first, then the pages in the current folder. Currently, it looks like this:

image

Notice that the pages are sorted alphabetically, but the folders are not. Additionally, the folders are interspersed with the pages.

techwiz24:
This pull request fixes #1012 that I opened earlier this week. The `/pages` view will now list all folders in the given base path alphabetically first, and then pages. I have also updated the tests for `/pages` to account for the changes.

While this should technically cover the changes in 1200afe, I think we may want to consider re-writing these tests or re-naming them, as they seem to only provide coverage for the `/Mordor` sub-folder. A better test would cover the root of the test repository because it contains more folders and a mix of upper- and lower-case pages.

I can write up a separate test for that situation if requested.

Sorry for the delay, Sinatra has been behaving funny on my arch install.

### DIFF
--- a/lib/gollum/views/pages.rb
+++ b/lib/gollum/views/pages.rb
@@ -31,9 +31,11 @@ module Precious
 
       def files_folders
         if has_results
-          folder_links = []
+          folders = {}
+          page_files = {}
 
-          @results.map { |page|
+          # 1012: Folders and Pages need to be separated
+          @results.each do |page|
             page_path = page.path.sub(/^#{@path}\//, '')
 
             if page_path.include?('/')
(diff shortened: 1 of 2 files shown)


lib/gollum/views/pages.rb
```diff
@@ -41,19 +43,23 @@ def files_folders
               folder_path = @path ? "#{@path}/#{folder}" : folder
               folder_link = %{<li><a href="#{@base_url}/pages/#{folder_path}/" class="folder">#{folder}</a></li>}
 
-              unless folder_links.include?(folder_link)
-                folder_links << folder_link
-
-                folder_link
-              end
+              folders[folder] = folder_link unless folders.key?(folder_link)
```
bartkamphorst: Shouldn't it be `folders.value?(folder_link)` ?

lib/gollum/views/pages.rb
```diff
@@ -41,19 +43,23 @@ def files_folders
...
-                %{<li><a href="#{@base_url}/#{page.escaped_url_path}#{page.name}" class="file">#{page.name}</a></li>}
+                page_link = %{<li><a href="#{@base_url}/#{page.escaped_url_path}#{page.name}" class="file">#{page.name}</a></li>}
+                page_files[page.name] = page_link
               end
             end
-          }.compact.join("\n")
+          end
+
+          # 1012: All Pages should be rendered as Folders first, then Pages, each sorted alphabetically
+          result = folders.sort_by{ |k, v|  k.downcase }.compact.map{ |f| f[1] }.join("\n") + "\n"
```
bartkamphorst: I think you can accomplish the same with the following, which maybe a little easier to read:
```ruby
result = Hash[folders.sort_by{| key, value | key.downcase} ].values.join("\n") + "\n"
```

bartkamphorst:
@techwiz24 Thanks for the PR!

> A better test would cover the root of the test repository because it contains more folders and a mix of upper- and lower-case pages. I can write up a separate test for that situation if requested.

Better, more sensible tests are always welcome. :+1:

lib/gollum/views/pages.rb
techwiz24: Ok, so I think I changed that at the last second before the commit and mixed it up. I think it makes more sense for it to be:

```ruby
folders[folder] = folder_link unless folders.key?(folder)
```

That way, each `folder` is only rendered once. I think I would have caught this with a more thorough test case (which I intend on writing)

lib/gollum/views/pages.rb
techwiz24: Agreed

lib/gollum/views/pages.rb
bartkamphorst: Sounds good. :+1:

techwiz24:
@bartkamphorst @dometto I've made the suggested style changes and have added a test that I feel provides better coverage. The tests pass locally, but it seems Travis-CI is slow today.

Also, as with my previous PR, will you want these commits rebased?

bartkamphorst:
Looks great @techwiz24 .

> Also, as with my previous PR, will you want these commits rebased?

Please. You can squash them all together, or make one for the logic and one for the test change. Thanks!

techwiz24:
@bartkamphorst Done. Thanks for all the feedback!

:white_check_mark: - Ready to merge this PR at your leisure

bartkamphorst:
Happy to help. Thank you for taking the initiative on this one!

PR #1032, opened 2015-07-03
Fix #1030: Disable AJAX Cach in Live Editor

### LINKED ISSUE #1030: Live Preview Editor should override cache settings for IE compatibility
According to this [blog post](http://www.dashbay.com/2011/05/internet-explorer-caches-ajax/) and [this stackoverflow post](https://stackoverflow.com/questions/4303829/how-to-prevent-a-jquery-ajax-request-from-caching-in-internet-explorer), IE always caches AJAX requests:

> In IE9′s new F12 Developer Console (finally!), cached requests show up as 304 Not Modified with a very suspicious < 1 ms response time.  The 304 is a red herring:  the server is **never** contacted.  It simply means IE in its infinite wisdom has decided the response wouldn’t have been modified IF it had asked the server, and serves it from browser cache.  Debugging on the server side shows that IE is really and truly *not* hitting the server.

This causes IE to load the wrong page content for the Live Preview Editor.

This can be fixed by altering [this ajax call](https://github.com/gollum/gollum/blob/3642370f116e66daecd3f68de351948d5d0bb4a2/lib/gollum/public/gollum/livepreview/js/livepreview.js#L394-L400) to disable the cache:

```js
jQuery.ajax( {
      type: 'GET',
      url: baseUrl + '/data/' + path + key_no_leading_slash( 'page' ),
      cache: false,
      success: function( data ) {
         editorSession.setValue( data );
      }
    });
```

I am unsure if this problem appears elsewhere in the source. The Live Preview Editor is the only place I have noticed it so far.

I've kept the original text of the issue below:
 
---

It seems that the Live Preview Editor is giving the wrong http status code for an XMLHttpRequest at `/data/<Page-Name>` for Internet Explorer, which causes IE to load incorrect page data from the cache:

URL | Protocol | Method | Result | Type | Received | Taken | Initiator
----- | ---- | ---- | ---- | ---- | ---- | ---- | ----
/data/Cache-Testing | HTTP | GET | 304 | text/html | 174 B | < 1 ms | XMLHttpRequest

and a `200` in chrome:

Name | Method | Status | Type | Initiator | Size | Time
-------- | -------- | -------- | ------- | ---------- | ------ | -------
Cache-Testing | GET | 200 | xhr | jquery-1.7.2.min.js:4 | 429B | 28ms

Our wiki instance is behind an nginx proxy (for ssl termination suppport), but this problem also occurs when accessing gollum directly on port `4567`. Clearing the cache in IE corrects the problem until the live preview editor is used to edit the page.

In the IE F12 Developer Tools under the network tab, selecting "Always refresh from server" appears to disable the cache, however it looks like IE thinks that gollum responds with a `200` for the previously mentioned XHR, so I'm not sure if this is a client side issue or if it's with gollum or somewhere further up the stack.

This problem does **not** occur with the standard editor.

### Steps to reproduce
1. Start `gollum` with the Live Preview Editor enabled
2. Open a page in both Chrome and IE
3. In both browsers, edit the page
4. In IE, make a change to the page and hit save
5. In chrome, simply press the back button to return to the page. After the change in IE has been made, refresh the page
6. Edit the page again in both browsers

Notice that IE is displaying the old content for the page, while chrome is displaying the newer content.

**EDIT**: To clarify: This is with IE11.

techwiz24:
Internet Explorer caches all AJAX get requests and decides when to expire
them in the cache. This commit disables caching on the GET request to
/data/<PAGE NAME>, which fixes IE displaying old page data when using the
live editor.

See http://www.dashbay.com/2011/05/internet-explorer-caches-ajax/ and
https://stackoverflow.com/questions/4303829/how-to-prevent-a-jquery-ajax-request-from-caching-in-internet-explorer

### DIFF
--- a/lib/gollum/public/gollum/livepreview/js/livepreview.js
+++ b/lib/gollum/public/gollum/livepreview/js/livepreview.js
@@ -394,6 +394,7 @@ var applyTimeout = function () {
     jQuery.ajax( {
       type: 'GET',
       url: baseUrl + '/data/' + path + key_no_leading_slash( 'page' ),
+      cache: false,
       success: function( data ) {
          editorSession.setValue( data );
       }


bartkamphorst:
Thanks @techwiz24 ! Sorry for the wait, will merge now.

PR #1035, opened 2015-07-21
Search

rgroux:
This is 2 commit for searching in gollum.

1. add empty querry (q='') as default 
2. remove empty results.

may be it is not interesting to show and return file with 0 matches.
I hope you will accept this request.

Thanks
/Richard

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -438,7 +438,7 @@ module Precious
     end
 
     get '/search' do
-      @query   = params[:q]
+      @query   = params[:q] || ''
       wiki     = wiki_new
       # Sort wiki search results by count (desc) and then by name (asc)
       @results = wiki.search(@query).sort { |a, b| (a[:count] <=> b[:count]).nonzero? || b[:name] <=> a[:name] }.reverse


dometto:
Thanks for the PR, I'll have a more detailed look at it later. But in what cases are results with count == 0 returned at the moment? Aren't those results hits based on the filename rather than the file's content?

rgroux:
when the count == 0, i check `a[:count].zero? == True`  where `a` is an occurence of `result`.

So if there are any matches, it not necessary to send the empty results. but after you send me the message, i thought it can be interesting to return the filename if it match the query.

May be an another PR for this. the search in gollum can be very powerful, it can be interesting to do something clever with it.

SkyCrawl:
@rgroux 

> add empty querry (q='') as default 

Thank you :).

@dometto 
For the record, this should eliminate an error when `http://localhost:4567/search` HTTP request is made. I think I stumbled upon it in the past as well, only forgot to report it... maybe because I fixed it in my forks.

@rgroux 
> Remove empty results.

Like @dometto indicated, **search should never return empty results with the current implementation, unless we have a bug**. The `search` route that you tried to mend is lacking an empty results check by design :).

If, however, empty results are somehow returned from Gollum, would you mind posting an example? Like I said, in such an event we have a bug and we should fix it, not surpress it (like you suggest) :).

rgroux:
@SkyCrawl 

I send you the html query/result not a picture

The query `http://127.0.0.1:9292/search?q=tutut` return this

```html
....
  <ul>
    <li>
      <a href="/Home">Home</a>
      <span class="count">(0 matches)</span>
    </li>
    <li>
      <a href="/Help/Home">Help/Home</a>
      <span class="count">(0 matches)</span>
    </li>
    <li>
      <a href="/_Footer.md">_Footer.md</a>
      <span class="count">(0 matches)</span>
    </li>
    <li>
      <a href="/_Header.md">_Header.md</a>
      <span class="count">(0 matches)</span>
    </li>
....
 </ul>
</div>
<div id="footer">
  <ul class="actions">
    <li class="minibutton"><a href="#">Back to Top</a></li>
  </ul>
</div>
</div>
```html

it' s what you expected?

**Comment by SkyCrawl** [comment]
> it' s what you expected?

Yes, thank you :). In that case, we should delve deeper into why this occurs (it is not supposed to).

What are your circumstances? Which system do you have? Which Ruby and version do you use? Which git adapter?

Also, could you please post the content of one of the returned files so we can attempt to reproduce the issue and find the bug? :)

**Comment by rgroux** [comment]
@SkyCrawl 
~~~~
ruby --version
ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
~~~~

~~~~
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.2 LTS
Release:	14.04
Codename:	trusty
~~~~

~~~~
bundle show
Gems included by the bundle:
  * RedCloth (4.2.9)
  * addressable (2.3.8)
  * asciidoctor (1.5.2)
  * builder (3.2.2)
  * bundler (1.3.5)
  * charlock_holmes (0.7.3)
  * creole (0.5.0)
  * diff-lcs (1.2.5)
  * expression_parser (0.9.0)
  * github-markup (1.3.3)
  * gitlab-grit (2.7.3)
  * gollum (4.0.0)
  * gollum-grit_adapter (1.0.0)
  * gollum-lib (4.0.3)
  * gollum-rugged_adapter (0.3b)
  * htmlentities (4.3.4)
  * kramdown (1.6.0)
  * mime-types (1.25.1)
  * mini_portile (0.6.2)
  * mustache (0.99.8)
  * nokogiri (1.6.6.2)
  * org-ruby (0.9.12)
  * posix-spawn (0.3.11)
  * rack (1.6.4)
  * rack-cas (0.11.0)
  * rack-protection (1.5.3)
  * redcarpet (3.3.2)
  * rouge (1.7.7)
  * rubypants (0.2.0)
  * rugged (0.21.4)
  * sanitize (2.1.0)
  * sinatra (1.4.6)
  * stringex (2.5.2)
  * tilt (2.0.1)
  * twitter-text (1.12.0)
  * unf (0.1.4)
  * unf_ext (0.0.7.1)
  * useragent (0.10.0)
  * wikicloth (0.8.3)
~~~~

in config.ru i have this git adapter

```ruby
module Gollum
  Gollum::GIT_ADAPTER = 'rugged'
end
require "gollum/app"
```

techwiz24:
To comment on the search results turning up with zero matches, I would double check the encoding and line ending style on those files. We've had a few false positives turn up in our environment a while ago because the pages were originally converted in PowerShell which encoded them with an encoding that git/gollum didn't like.

SkyCrawl:
Thanks, @techwiz24 . So you experienced this too? E.g. `/wiki/my-page.md ... (0 matches)` search results?

Let's look at the code for a moment:

1. The [search routine](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/wiki.rb#L601) in `gollum-lib`  
I can only conclude that the wrong results (in case of @rgroux ) must come from the `@repo.git.grep(query, options)` call because otherwise, `results[file_name].to_i` must evaluate to `-1` in the `ls-files` call below. So, theoretically... yes. Encodings might play a part in this :).
2. But, here comes the [grep call in rugged adapter](https://github.com/gollum/rugged_adapter/blob/master/lib/rugged_adapter/git_layer_rugged.rb#L174)  
It explicitly states that it doesn't want empty results. The only way for potential "zero" to make it through is for `count` to be `nil`, which it can not be :/.

This is one weird issue...

rgroux:
Thanks to you
and sorry for this bug. 

I use rugged, because  bare repo is mandatory in my case, and I didnt find a way to use `repo_is_bare = true` with grit adapter.

gollum is started as rake app (dev) and apache passenger (prod)

techwiz24:
@SkyCrawl Keep in mind that [the line](https://github.com/gollum/rugged_adapter/blob/master/lib/rugged_adapter/git_layer_rugged.rb#L174) in the rugged adapter you mentioned is not released yet.

@rgroux can you try installing the adapter from master and either revert 6f2fb3 or try the search in the master branch? That should fix the false positives (results with 0 matches)

SkyCrawl:
@techwiz24 
Yes, just noticed your update which, right now, seems to have disappeared again :). Yup, seems that a bump to rubygems has not yet been made.

SkyCrawl:
@rgroux 
For the sake of completeness, it should be easy to run Gollum via rack with grit adapter, on a bare repo.

This is the example `config.ru` taken from Gollum's home page:
```
#!/usr/bin/env ruby
require 'rubygems'
require 'gollum/app'

gollum_path = File.expand_path(File.dirname(__FILE__)) # CHANGE THIS TO POINT TO YOUR OWN WIKI REPO
Precious::App.set(:gollum_path, gollum_path)
Precious::App.set(:default_markup, :markdown) # set your favorite markup language
Precious::App.set(:wiki_options, {:universal_toc => false})
run Precious::App
```

The `{:universal_toc => false}` hash holds all the major options to run Gollum with, including whether the repository is bare or not:

```
Precious::App.set(:wiki_options, {
    :universal_toc => false,
    :repo_is_bare => true
})
```

Did you try it like this? :)

rgroux:
@SkyCrawl 

yes i try
my config
~~~~~
Precious::App.set(:wiki_options, {
        :mathjax => true,
        :live_preview => false,
        :universal_toc => false,
        :repo_is_bare => true
})

~~~~~

mesage is 
~~~~
2015-07-24 18:04:29 - Grit::InvalidGitRepositoryError - /data/Wiki-bare:
	/data/gollum/gem/ruby/1.9.1/gems/gitlab-grit-2.7.3/lib/grit/repo.rb:52:in `initialize'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-grit_adapter-1.0.0/lib/grit_adapter/git_layer_grit.rb:226:in `new'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-grit_adapter-1.0.0/lib/grit_adapter/git_layer_grit.rb:226:in `repo'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-grit_adapter-1.0.0/lib/grit_adapter/git_layer_grit.rb:203:in `versions_for_path'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-lib-4.0.3/lib/gollum-lib/page.rb:281:in `versions'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-4.0.0/lib/gollum/app.rb:498:in `show_page_or_file'
	/data/gollum/gem/ruby/1.9.1/gems/gollum-4.0.0/lib/gollum/app.rb:481:in `block in <class:App>'
....
~~~~

SkyCrawl:
Hmm, the message seems to indicate that you were using Ruby v1.9.1? Gollum requires the minimum of 1.9.3 :).

rgroux:
@SkyCrawl 
yes i did, 
ubuntu default ruby/bundler make strange things

SkyCrawl:
@rgroux 
Thanks again, seems like this is another bug which I just managed to reproduce :(. Onwards to create a new issue then...

Anyway, back to your PR.

@dometto 
Eventually, it came out that:
* @rgroux stumbled upon a `rugged` searching issue,
* the very same one that @techwiz24  also stumbled upon and which you personally fixed,
* but those fixes have not yet been pushed to rubygems.

When it comes to the PR, the first proposed fix is perfectly valid and effective. But personally, I think it's better not to surpress issues like these with the second proposed fix. Maybe it's better to know that something is "still wrong"? :)

dometto:
> When it comes to the PR, the first proposed fix is perfectly valid and effective. But personally, I think it's better not to surpress issues like these with the second proposed fix.

I agree. But do we need the default query = `''` fix? If so, @rgroux, could you modify this PR (or open a new one) that only includes the default query fix?

I will do a new release of the rugged adapter so the empty search results issue is resolved.

rgroux:
@dometto 

I did it.
you can continue 

Thanks a lot. 
gollum is awesome

PR #1047, opened 2015-08-05
Changed layout template, to handle custom.x with page-file-dir

benubird:
Because currently, gollum is only able to read custom files from the root directory - but when page-file-dir is set, the root directory is not accessible, making it impossible to load a custom.js or custom.css file.

### DIFF
--- a/lib/gollum/views/layout.rb
+++ b/lib/gollum/views/layout.rb
@@ -20,8 +20,16 @@ module Precious
         !@path.nil?
       end
 
+      def page_dir
+        @page_dir
+      end
+
       def base_url
         @base_url
+      end
+
+      def custom_path
+        "#{@base_url}#{@page_dir.nil? ? '' : '/'}#{@page_dir}"
       end
 
       def css # custom css
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -98,6 +98,7 @@ module Precious
       forbid unless @allow_editing || request.request_method == "GET"
       Precious::App.set(:mustache, {:templates => settings.wiki_options[:template_dir]}) if settings.wiki_options[:template_dir]
       @base_url = url('/', false).chomp('/')
+      @page_dir = settings.wiki_options[:page_file_dir].to_s
       # above will detect base_path when it's used with map in a config.ru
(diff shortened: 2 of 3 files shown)


dometto:
:+1: Looks good to me. Thanks for reporting and providing the quick fix, @joshua-paddle  and @benubird!

lib/gollum/views/layout.rb
```diff
@@ -20,10 +20,18 @@ def has_path
...
+      def page_dir
+        @page_dir
+      end
+
       def base_url
         @base_url
       end
 
+      def custom_path
+        @base_url + (@page_dir.nil? ? '' : '/') + @page_dir
```
dometto: I think we want to encourage using string interpolation wherever possible:
`"#{@base_url}#{@page_dir.nil? ? '' : '/'}#{@page_dir}"`

dometto:
@bartkamphorst, any further suggestions/worries?

lib/gollum/views/layout.rb
bartkamphorst: Yes, string interpolation is the preferred way.

bartkamphorst:
Not that I can think of. Seems like a sensible contribution. Thanks @benubird !

dometto:
Please incorporate the suggestion about interpolation (and squash the commits if necessary) and I'll merge it ASAP. :-)

benubird:
Yeah, I don't really know anything about ruby... or gollum... :p 

Not sure who you meant should be adding the interpolation, so I assumed me?

dometto:
Yeah, thanks. :+1: Interpolation is more canonical in ruby than using `+', and also more efficient (`+' initializes a new `String` object).

mchill:
Not sure if commenting on a closed pull request is proper, but this breaks custom.css/custom.js when page_file_dir isn't set. See my comment on the commit.

bartkamphorst:
Thanks @mchill ; this is very helpful. Care to try a PR of your own with a fix?

dometto:
Ugh, we really ought to have tests for all the frontend stuff. :-( Thanks for pointing this out @mchill

mchill:
@bartkamphorst I could give it a whirl after work. I'll see about adding a test, too.

Issue #826, opened 2014-05-14
Live Preview user experience roadmap?


rtrvrtg:
Hi,

I work for a company that uses Gollum for internal documentation. I'm interested in working on user interface enhancements for Live Preview, and am interested to see what intentions the Gollum community had in mind for it.

I've been mulling over a few ideas for improvements to the user experience, and wanted to see if anyone else was interested, or that the developers had a particular user experience roadmap in place, before I started hacking away on these ideas on my own branch.
### Shortcut key combinations

There seem to be a rather limited set of key combinations used in live preview at present, and they don't feel particularly discoverable.  Would anyone be interested in any additions to what there is at present?

One candidate that seems logical is Ctrl+S for Save (and maybe Ctrl+Shift+S for Save with Comment). I know that some people might feel leery of this idea given that it hijacks common browser commands, but I feel that there's sufficient precedent for this with Google Docs and other web-based editors that this might change people's expectations, allowing this to work in a different way from the default. 
### Implementation of fragments in Live Preview

I've done a little exploring and think it might be possible to render custom fragments of Markdown server-side, allowing things like links and custom macros to be depicted accurately on the client side. This might not be the ideal way to integrate custom markup, but it should be a way of doing so without having to hand off too much to Gollum's internal preview.

Would anyone else be interested in such a feature?
### Toolbar icons

In terms of the toolbar icons that are used in Gollum's Live Preview mode, we figured that they should be brought in line with the major features from the non-live editor.

This would mean that it'd contain buttons such as:
- Link
- Bold / Italic / Underline
- Lists (ordered and unordered)
- Quotes
- Horizontal Rule
- Headings
- Links
- Images
- Contextual Markdown help

It may make sense to retain Preview, but only if we weren't able to preview every single bit of the page using Live Preview.

Separating Save and Save with Comment don't make any sense; it might be useful to include some sort of contextual switch between them, or to expose the comment as an option when you save.

Also, the pane-switching icon seems kind of superfluous. I'm wondering if this isn't something we could conceal behind something else.
### Per-user settings

The bit about the pane-swapping icon made me wonder if there's any scope for global user preferences in this project to cover features such as this. The reason behind this is that most users won't need to swap the order of editing and preview panes more than once, but they'll probably want to do so on a global basis.

It might be possible to implement something trivial on the front-end only using localStorage, but since this is highly dependent on browser capabilities, I figured we ought to explore other possibilities first.
### Front-end performance

[...]

PR #1048, opened 2015-08-07
Editor to ace

dometto:
Working PR to merge @rtrvrtg's new livepreview mode (incorporated in the default editor). See discussion in #826. In addition, I:
1. Deleted the (old) livepreview subdirectory
2. Changed the default editor to use Ace. Benefits:
   - We can use syntax highlighting for e.g. markdown
   - Uses latest ace, easy to update, less to maintain
   - Simplifies `gollum.editor.js` significantly (no more manual routines to replace text, set selections, add keyboard shortcuts)
   - Possible benefits of ace features such as line numbers etc. (if the user wants them)

Currently the only thing I can find that's not yet working is drag-drop upload.

Huge PR because of the long history and importing/deleting of JS files. Should probably squash this into something more readable. Would definitely like some feedback on my commits (not @rtrvrtg's :)) as I'm not at all experienced with JS -- @sunny? Especially thinking about my changes to `gollum.editor.js` and `edit.mustache`.


dometto:
I had to basically revert some of @rtrvrtg's changes, like [cooler function buttons](https://github.com/gollum/gollum/commit/2c8d00f57b76a76dd214aa9306a237d4630e317e), because I couldn't get the jquery textManipulator to work with ace. :-(

SkyCrawl:
Just a thought on the `Use Ace as default editor` commit:
I think we should avoid replacing all the old code (so many files) with a lot of new code. Specifically, when I was discovering Gollum for myself, I really didn't like the way it bloated the `gemspec` and what efforts were needed to upgrade the thing. Imho, it would be much nicer to convert the Gollum frontend app to Rails and use gems like `ace-rails-ap`. Gollum will not only benefit from this regarding Ace, but also MathJax and maybe some more stuff. Gollum's offline mode would also improve significantly, since MathJax is pulled from CDN, if I recall correctly :).

dometto:
Moving gollum to rails is, realistically, not going to happen I think. :) Plus it's not really a good idea I think, given Rails's database-centeredness, which I've discovered doesn't really play well with a git-based backend.

I'm all for using a gem to require ace, but there isn't one for Sinatra I think. Might it help to extract the editor into a separate gem, like gollum-lib?

I appreciate that the code base is bloated atm. In fact the whole point of this operation, for me, is to get rid of the dual maintenance load of two editors, and have as little custom code as possible. The new code is mostly new versions of ace and some JS libraries. There's a lot of CSS stuff because those commits were deep in the branch so it wasn't easy to cherry pick only the livepreview commits, but as I said I'm going to try and come up with a squashed version, so hopefully I can remove the stuff that's not strictly necessary (I think most of the CSS stuff is alrealry in the css-to-sass branch).

SkyCrawl:
> I'm all for using a gem to require ace, but there isn't one for Sinatra I think. Might it help to extract the editor into a separate gem, like gollum-lib?

Might be a better idea idd. And do the same for the mentioned MathJax? But then we also need to properly select and integrate an asset technology for Sinatra?

> I think most of the CSS stuff is alrealry in the css-to-sass branch

In case you're referring to mine, then yes. Most of the work is done, mostly only beside the editors actually :(. The bad thing is (for the sake of integration) I had to "mess" (although I rather call it "clean" or "introduce structure to") with mustache templates and some other things too, mostly to unify all the generated HTML and CSS and make it editable via `config.rb`. And then I got caught up in changing `gollum-lib`, be it generated HTML and CSS content or the code itself and eventually ran out of time :(. So the current state is still pretty unfinished, unfortunately, despite that the work thus far is done well I hope (according to my, usually high, standards) :).

dometto:
> In case you're referring to mine, then yes.

Actually, I was referring to @rtrvrtg's commits: https://github.com/gollum/gollum/commits/css-to-sass . Not sure if they already include his work on making the gollum theme more responsive etc. (the CSS commits in this PR).

techwiz24:
Sweet! Quick thing I noticed (Chrome 44.0.2403.130):

In it's default location, the live preview split overlaps certain elements:

image

It must be moved a considerable amount, which makes the preview window sort of small:

image

Same problem when the editor is on the other side:

image

Naatan:
Not to undermine the awesome work being done here, but I was able to get [SimpleMDE](http://nextstepwebs.github.io/simplemde-markdown-editor/) running on Gollum with very minimal effort. Basically just had to follow its installation steps and ensure the Gollum editor wasnt trying to load. I also removed everything but the Gollum texteditor area because it had been made redundant by SimpleMDE.

I love the way the end result looks, and think it could possibly make sense for Gollum.

Again I don't mean to undermine the work being done here, just weighing in with an alternative so that you may weigh your options accordingly.

dometto:
Thanks for the suggestion. SimpleMDE is really nice, but it looks like it only supports markdown, where gollum tries to support multiple markups. Ace has support for a couple, though by no means all of them. I think the ideal solution would be to abstract the gollum editor somehow (into a gem), and in a later step to make it really easy for users to enable their own editor, i.e. through letting them configure their own template or some such.

Naatan:
I had not considered that, being as I only intend to use gollum for Markdown. That's a good point. I like the idea of allowing the user to choose their preferred editor (override the default). That may prove challenging though as so far gollum is easily hacked but not easily modded.

dometto:
For those interested, I've started to abstract the editor out to its own gem, hopefully making it easier for endusers to define their own editor (and making it possible to use the gollum editor in other projects).

See the [changes to gollum](https://github.com/dometto/gollum/tree/abstract_editor) and the [editor gem](https://github.com/dometto/gollum-editor)

The idea is that you can subclass `Gollum::Editor::Default` and set `Gollum::Editor::ACTIVE_EDITOR` to override the default editor.

SkyCrawl:
@dometto 

Wish I could join in the efforts but I probably took on too much work :(. Well, at least I may be able to help out a little in a month's time, with some testing and revision.

> For those interested, I've started an experiment to abstract the editor out to its own gem, hopefully making it easier for endusers to define their own editor (and making it possible to use the gollum editor in other projects).

Like @Naatan , I have no intent to undermine the great work being done here but still... I can't help but express a few thoughts (please forgive me for producing so long texts usually) and maybe present another way to look at this :).

What I don't seem to understand at the moment are the benefits of abstracting the whole editor into a separate gem. Especially since you mentioned:

> 1. The idea is that you can subclass Gollum::Editor::Default and set Gollum::Editor::ACTIVE_EDITOR to override the default editor.
2. One of the many problems with the current setup is that the editor uses the functionality from gollum.dialog.js. So that has to be packaged with the editor gem. But other parts of gollum use the dialog functionality too, so they untransparently depend on the editor gem.

The first point is, as I understand, a definition of basic interface - where to look for assets most of all. The second point then seems to indicate what I was getting at in the lines above and for one, I think finding a good solution to it is going to be extremely difficult if larger code duplication is out of question.

As I understand (correct me if I'm wrong), the editor's functionality (and CSS/JS) is and will be (at least a little) tied to Gollum. Also the requirement for a mustache template? However I look at it, it's difficult for me to imagine somebody using the extracted editor for their own purposes. And should they really want to, it won't be difficult to download/filter the assets related to it, if we define and separate them clearly in the code (maybe with a bit of associated documentation). After all, it is almost an independent unit that may inherit from Gollum's templates, views, JS and SASS.

Also, it is not quite easy (for me, personally) to imagine somebody wanting to program their own editors for Gollum. I think such cases will be extremely rare, as long as Gollum's default editor is good enough (which I hope it will be when we all give it some effort). It is likely that most of them will be slight customizations (SC) of Gollum's default editor in which case it doesn't make much sense to me to integrate them into Gollum via gems, as this would probably require either to launch Gollum from source, install the SC as a gem into the system or publish the SC as separate gems (all of which seem kinda inefficient, if nothing else).

In such cases, the gem will likely be local and it makes more sense (for me) to integrate it into Gollum directly, without deeper knowledge of Ruby and Rubygems. E.g. by subclassing `Gollum::Editor::Default`, defining paths to the required assets and using, e.g. Gollum's configuration file. All the files should be allowed to live outside the Gollum repo anyway so no harm done to individual repositories and if somebody would really like to publish their work for other Gollum users, it only needs to be a basic folder (not a gem) so a plain GitHub repo and (for example) a notice in the Wiki is enough.
Note that Gollum's configuration file is universal and doesn't disallow the use of gems for this purpose - if anyone finds that beneficial.

To me, it makes more sense (and seems more "best-effort") at the moment to:
* First abstract the editor within Gollum itself, on the basis of interfaces and extract all the code related to the default editor into separate files (assets). Example asset structure:
```
gollum_editor_abstract.rb (the main interface that all editors should provide)
|-- default (folder)
|-- -- gollum_editor_default.rb (subclasses the abstract editor, provides assets and information)
|-- -- other files, potentially the required assets but those may be elsewhere
|-- my_editor (folder)
|-- -- gollum_editor_mine.rb (subclasses the abstract editor)
|-- -- other files, potentially the required assets but those may be elsewhere
...
```
* Using Gollum's configuration file (that holds wonderful potential, yet Gollum doesn't use it as much as it would deserve) to dynamically switch between the defined editors, e.g.:
```
# OPTIONAL
# may be a gem or a local Ruby script subclassing Gollum::Editor::Abstract
# may or may not be located within the current Gollum repo (same as this particular configuration file)
require 'my-editor'

# <editor> may be one of the predefined editors (probably only the default) or a custom editor, e.g. 'MyEditor::Editor.new'
Precious::App.set(:editor, <editor>)
```
* And finally, extracting Ace into a separate Sinatra-asset-ready gem, stripping it from the bloated Gollum's gemspec and making an example for anyone who'd like to:
    * Use another technology as a base for Gollum's editor.
    * Extract MathJax to the same kind of gem and improve Gollum's offline mode.
    * Etc.

## Summary

In my humble opinion, Gollum's codebase already spans across enough separate gems:

1. If we add a gem for each major customizable feature, things might be out of hand one day, as Gollum grows and potentially loses maintainers. On the contrary, Gollum needs more maintainers and to me, it doesn't seem to be a good idea to split them among all these gems. Let alone increasing the load for @dometto , who's doing wonderful work at finding enough time to manage them all and still take part in the coding.
2. As somebody who'd like to improve on Gollum and doesn't know it all too well (and possibly Ruby too), I'd personally be quite discouraged if it involved several gems (projects) at once, possibly fundamentally different. For one, I like simple, layered, well designed and documented code which allows me to get to the point quickly and efficiently. In my experience, dividing the code into different projects only helps in some special cases and this is definitely not one of them.

So... as for me, I'd avoid the fragmentation and complexity as much as possible and use a good-looking, well designed interface/layer/folder structure within a single repository as much as possible (where something else is not strictly required or appropriate). When it comes to customizability, we can make usable shortcuts for Gollum's configuration file most of the time.

calid:
I'd love to see which editor people use made pluggable.  I'm personally not a huge fan of Ace.

@SkyCrawl I wouldn't worry about a proliferation of gems.  A plugin based model is much more flexible and sustainable in the long run.  If done right it can help promote an active and healthy community.

dometto:
I also still like the idea of having more or less pluggable adapters. However, my experimental branch ([here](https://github.com/dometto/gollum/tree/abstract_editor)) is much too messy. It just struck me that the right way to do this is by harnessing the [already-present](https://github.com/gollum/gollum/tree/master/lib/gollum/views) powers of mustache view. For example, we could move the editor source to a new `Editor` view, and making the view class overridable.

bartkamphorst:
Interesting idea. :+1:

dometto:
Closing in favor of https://github.com/gollum/gollum/pull/1173

jakob-o:
Hi,

since this got closed in favor of #1173 I wanted to kindly ask if there are any plans to support the live preview functionality as it is in this PR?

Regards

Jakob

dometto:
Yes, we would actually like to get that in the `5.0` release. I haven't looked at the specifics, but I'm guessing it should be quite doable to adopt the approach from this PR and implement in top of the new ACE editor. We're quite busy with the new release so any help is welcome, if you'd like to open a PR! Always happy to review and help out where needed.

jakob-o:
Hi,

thank you very much for your quick answer. Unfortunately I have only been browsing the branches as a user but didn't have the time to actually check out the code. If you have a specific request where I could support you let me know and I / or one of my colleagues could try to help out.

Regards

Jakob

PR #1055, opened 2015-08-14
Add webrick as development dependency. Fixes #1053.

### LINKED ISSUE #1053: content-length warnings
The server log is filled with these:

```
[2015-08-12 10:47:44] WARN  Could not determine content-length of response body. Set content-length of the response or set Response#chunked = true
```

It doesn't seem to affect anything, but it makes the log hard to scan and it seems like it ought to be easy to fix for someone who knows his way around.

bartkamphorst:


dometto:
:+1:

PR #1060, opened 2015-08-21
Fix custom.x path when page-file-dir is not set

mchill:
Addresses problem introduced by https://github.com/gollum/gollum/pull/1047. Edit an existing test and add a new test to ensure that the path to custom.css/custom.js correctly reflects page-file-dir.


dometto:
:+1: Many thanks!

bartkamphorst:
:+1: Merging.

Issue #1049, opened 2015-08-07
Users?


Naatan:
Does Gollum have anything to support user registration of some sort? Atm if I were to make a wiki public then anyone could edit without any sort of authentication and the author would always be the same to Gollum. How does this work? Is there any documentation on the topic?

Ironically I find myself googling a lot of these questions for Gollum, the current documentation is very sparse.

dometto:
There's nothing in the docs about this because gollum doesn't have any user/authorization support. :-) The is, however, a `session[gollum.author]` key that sets the name for commits and such.

Gollum is just a rack app, so you can put any bit of authentication and authorization middleware you want in front of it. Some examples:
* [omnigollum](https://github.com/arr2036/omnigollum)
* @dekimsey's comment  [here](https://github.com/gollum/gollum/issues/702) for how to implement basic auth middleware.
* I've recently implemented something similar in a Rails app [here](https://github.com/repotag/repotag/blob/master/app/classes/repotag_auth_proxy.rb#L64).

dometto:
Also, the docs are mostly the public wiki for this repo. Please feel free to explain anything you run into there!

Naatan:
Thanks @dometto. I figured gollum didnt support users/auth, but since it's fairly vital for public wiki's I expected at least a mention of it. Your answer perfectly addresses the question, I don't mind submitting it to the wiki but I'd feel like I'd be stealing your efforts as it'd just be a copy paste. Consider adding it or let me know if you're ok with my copy pasting it in there.

dometto:
No worries, you're right we should mention this in the wiki. I'll try to type something up. But if you encounter anything else please feel free :)

rgroux:
I use gollum with CAS auth, and i will make (ASAP) a doc/how to. I am actualy in a train in Milano. So i think i will commit on my page and hope for PR soon. 

This is a very basic auth. Just read/write, and no page modification, so everyone has edit button. 

/Richard

dometto:
Please note that there is already a --no-edit command line switch (wiki_options[:allow_editing]). We're not looking to build authentication/authorization  support into gollum, as that can/should be done in a middleware. However, examples for such a middleware a very welcome (in contrib or the wiki).

rgroux:
Indeed. I am titaly agréé with this. I use only standard gems. And if i talk about a PR it s only for docs or readme, notre à features.

In my opinion, it's only a nice must have for few peaple. 

[...]

PR #1061, opened 2015-08-25
README add link to wiki about CAS SSO

rgroux:
Hi,

like i wrote in the Issue #1049, i use gollum with a CAS SSO server, so I made a wiki entry about this.

With this you can have RW access for some user specified in a config file, and everybody else have only RO access to the wiki.

Hope it my help someone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ For more information on Gollum's capabilities and pitfalls:
 1. [Syntax/capability overview for pages](https://github.com/gollum/gollum/wiki).
 2. [Known limitations](https://github.com/gollum/gollum/wiki/Known-limitations).
 3. [Security overview](https://github.com/gollum/gollum/wiki/Security).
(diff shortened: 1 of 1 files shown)


rgroux:
is it ok?

SkyCrawl:
I'm not a maintainer so the decision is not mine but personally, I would rather add this into the [Running section in README](https://github.com/gollum/gollum#running), e.g. something like:
```
### With an authentication server

Gollum can also be ran alongside a CAS (Central Authentication Service) SSO (single sign-on) server. With a bit of tweaking, this adds basic user-support to Gollum. To see an example and an explanation, navigate [over here](https://github.com/gollum/gollum/wiki/Gollum-via-Rack-and-CAS-SSO).
```

rgroux:
Thanks, 

it's a far better idea to write this into the [README's Running Section](https://github.com/gollum/gollum#running), I change my commit an update the pull request.

dometto:
Sorry for the late reply, but thanks for taking the effort to explain how this works in the wiki and Readme, @rgroux!

Issue #1052, opened 2015-08-12
Set page_file_dir for internal use only


Naatan:
I want to set a sub directory for wiki pages, however it seems that when I use `page_file_dir` (or `--page-file-dir` for the CLI) it exposes this directory to end-users. Eg. if I set it to "wiki" then this path will be represented in the url.

I want to set a sub directory that is only internally used. From an end-user perspective the root directory should not change.

Any way of doing this?

SkyCrawl:
Hello @Naatan 

I'm afraid you can not do anything about this at the moment, except forking both `gollum` and `gollum-lib` and fixing all the necessary code, which might prove unexpectedly difficult for certain reasons :(.

I may be wrong but it seems to me that it hasn't been done before either because noone noticed or because it spans both the above gems. I will add it to the list of known limitations, thanks for reporting :).

@dometto 
Let's label this issue?

bartkamphorst:
> I want to set a sub directory for wiki pages

@Naatan Just so I'm clear on what you're asking, could you tell me a little more about your setup? Is it that you have an existing git repository and want to add wiki pages to that repo? Or is it that you don't want the pages being saved to gollum's root directory when you run `./bin/gollum`?

Naatan:
Basically I have a repository that in its root has public, views and templates folders. These should not be picked up by gollum as being part of the wiki.

The only way around it that I've found is to have the wiki itself be a different repository, which I then run gollum on. This is not ideal however, I'd prefer to only have 1 repository that covers the project.

dometto:
This has been implemented in `5.x`. See #1328

dometto:
Oops, wrong issue. Reopening.

dometto:
Implemented in 5.x!

PR #1077, opened 2015-10-04
fixed file renaming issues with page_file_dir

mpnowacki:
This will only work with appropriate changes in gollum-lib, corresponding pull request is here: https://github.com/gollum/gollum-lib/pull/174

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -179,7 +179,7 @@ module Precious
       halt 500 unless tempfile.is_a? Tempfile
 
       # Remove page file dir prefix from upload path if necessary -- committer handles this itself
-      dir      = wiki.per_page_uploads ? params[:upload_dest].match(/^(#{wiki.page_file_dir}\/+)?(.*)/)[2] : 'uploads'
+      dir      = wiki.per_page_uploads ? params[:upload_dest] : ::File.join([wiki.page_file_dir, 'uploads'].compact)
       ext      = ::File.extname(fullname)
       format   = ext.split('.').last || 'txt'
       filename = ::File.basename(fullname, ext)


lib/gollum/app.rb
```diff
@@ -179,7 +179,7 @@ def wiki_new
       halt 500 unless tempfile.is_a? Tempfile
 
       # Remove page file dir prefix from upload path if necessary -- committer handles this itself
-      dir      = wiki.per_page_uploads ? params[:upload_dest].match(/^(#{wiki.page_file_dir}\/+)?(.*)/)[2] : 'uploads'
+      dir      = wiki.per_page_uploads ? params[:upload_dest] : ::File.join([wiki.page_file_dir, 'uploads'].compact)
```
dometto: Now I wonder: shouldn't we make sure that `params[:upload_dest]` is a path under `wiki.page_file_dir`?

mpnowacki: And return a BAD REQUEST if it isn't? I think it's not a good idea, because pages can be not under `page_file_dir`. The same should go for files. I understand, that this is risky, because someone can actually overwrite a project file, but enabling uploads is for internal use only anyway (i guess, or rather hope).

dometto: Also sounds reasonable. Again, the Right Way seems to be to make page-file-dir internal only, so we can just prepend `wiki.page_file_dir` in any case.

mpnowacki: that's precisely what I had in mind. This change seems inevitable, so there is not much sense in doing 'just-for-now' security checks.

SkyCrawl: I agree that `page_file_dir` should be completely internal, just as #1052 suggests. But further than that, I've almost always thought that Gollum is really sloppy with handling internal paths (several code regions seem to each handle them in a separate way) which logically resulted in many issues. If I remember correctly, I also encountered several places in the code that allow for external files (outside the given repository) to be displayed on the wiki. I tried to fix this in my forks but never finished it.

Personally, I think it's really time to revise all path-related features and come up with a unified, documented and enforced model...

dometto: On 05-10-15 12:21, Jiří Smolík wrote:
> Personally, I think it's really time to revise all path-related 
> features and come up with a unified, documented and enforced model...

Handling `page-file-dir` only internally will actually make this much 
easier, since we can then be sure that any request paths from the client 
don't contain the `page-file-dir` prefix yet. So there should just be a 
helper method on the `Wiki` class that converts request paths to in-repo 
paths (and one to do the opposite I guess). This being ruby I think 
there's no way to 'enforce' the use of a particular method to construct 
a path, nor should there be. But yeah, we can document it and make it 
explicit that developers should use the helper method.

For now we have a serious bug, and we should just squash it instead of 
waiting for someone to do the Right Thing Soon(tm).

SkyCrawl: > Handling `page-file-dir` only internally will actually make this much easier, since we can then be sure that any request paths from the client don't contain the `page-file-dir` prefix yet.

Yep.

> So there should just be a helper method on the `Wiki` class that converts request paths to in-repo paths (and one to do the opposite I guess).

Almost what I was suggesting :).

> This being ruby I think there's no way to 'enforce' the use of a particular method to construct a path, nor should there be.

Using only a single method everywhere is not an option of course, there are some "variables". But why couldn't we come up with several methods that do the hardest "internal" work, common to all cases?

> For now we have a serious bug, and we should just squash it instead of waiting for someone to do the Right Thing Soon(tm).

Sure, nothing against fixing a bug or two. I was just trying to get your attention to the big (imo) issue :).

dometto:
Sorry, got distracted from this. @mpnowacki, any changes pending, or shall I merge this?

mpnowacki:
No pending changes from me

dometto:
Thanks again @mpnowacki!

mpnowacki:
No problem, However For this whole thing to work the pull request in gollum-lib has to be fixed as well. Most of the changes were done there.

PR #1086, opened 2015-11-02
Fix position of sidebar in preview

ngyuki:
hi!

In preview of when there is a sidebar, the sidebar does not appear in the correct position.

I tried to fix this.


dometto:
:+1:

ngyuki:
Thank you!

PR #1091, opened 2015-11-26
Use `last_version` instead of `versions` when possible. Fixes #1087.

### LINKED ISSUE #1087: Unnecessary usages of `Gollum::Page#versions`
OK, so, while profiling the regular page view action, I have hit the following performance issue (I have a repository that has been migrated, and because of that I already have 30k revisions, but my new pages only have about 3 or 4).

In the code path to render a page, the `Gollum::Page#versions` method is called three times:

 * Once in [`Precious::App#show_page_or_file`](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L499)
```ruby
@page_exists   = !page.versions.empty?
```
 * Once in [`Precious::Views::Page#author`](https://github.com/gollum/gollum/blob/v4.0.1/lib/gollum/views/page.rb#L25-L26), and once in [`Precious::Views::Page#date`](https://github.com/gollum/gollum/blob/v4.0.1/lib/gollum/views/page.rb#L32-L33):
```ruby
page_versions = @page.versions
first         = page_versions ? page_versions.first : false
```

In all three cases, I feel that the call to `Gollum::Page#versions` is mostly unnecessary, as only the last version is pertinent.
However, these calls provoke a rather enormous performance dip in my repository because of these facts:

 1. `#versions` is requested for my page, with the default pagination setting - so Gollum will look for 30 versions of my page;
 2. Because my page only has 2 or 3 versions in the repository, the entire revision range will have to be examined by Gollum in case more versions are actually available;
 3. Because my repository has 30k revisions, that takes about 300 ms for each call (with Grit - it is a lot worse with Rugged);

I can think of 3 distinct solutions for that issue, in decreasing order of personal preference:

 1. Use the `Gollum::Page#version` attribute - however, I am not fully clear on how to use it in the case of a preview page, as it is then filled with the last commit information;
 2. Define a `Gollum::Page#last_version` method, that calls `Gollum::Git::repo#log` with a `:max_count` of 1 - however, it seems awful to have to add a new method for that;
    * Note: that is the solution I have currently implemented on my instance (which led me to raise [that `rugged_adapter` issue](https://github.com/gollum/rugged_adapter/issues/10));
 3. Cache the result of the invocation of `Gollum::Page#versions` - I do not like it because caching is always an issue for invalidation;

Thank you for your help.

jhominal:
Use Gollum::Page#last_version instead of Gollum::Page#versions in the
cases identified in #1087:
 * In Precious::App#show_page_or_file
 * In Precious::Views::Page#author
 * In Precious::Views::Page#date

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -496,7 +496,7 @@ module Precious
 
         # Extensions and layout data
         @editable      = true
-        @page_exists   = !page.versions.empty?
+        @page_exists   = !page.last_version.nil?
         @toc_content   = wiki.universal_toc ? @page.toc_data : nil
         @mathjax       = wiki.mathjax
         @h1_title      = wiki.h1_title
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -22,15 +22,13 @@ module Precious
       end
 
       def author
-        page_versions = @page.versions
-        first         = page_versions ? page_versions.first : false
+        first = page.last_version
         return DEFAULT_AUTHOR unless first
         first.author.name.respond_to?(:force_encoding) ? first.author.name.force_encoding('UTF-8') : first.author.name
       end
 
       def date
-        page_versions = @page.versions
-        first         = page_versions ? page_versions.first : false
+        first = page.last_version
         return Time.now.strftime(DATE_FORMAT) unless first
         first.authored_date.strftime(DATE_FORMAT)
       end


jhominal:
The Travis CI check fails because the referenced method exists only in gollum-lib's master branch, and the CI runs on packaged versions. How can we handle that?

dometto:
Thanks again, looking fine. The tests will not pass until we release the new gollum-lib gem. Any more forthcoming changes for gollum-lib, or should we release the new gem ASAP?

jhominal:
I am not in a hurry for pushing a new version of gollum-lib, I think I still have one or two performance issues that I would like to profile and fix. Would it be better to retract this PR in the meantime?

dometto:
Just keep it open. There are some macro-related PRs open for gollum-lib now, I'll try to get those merged and then release a new version.

bartkamphorst:
> I think I still have one or two performance issues that I would like to profile and fix.

@jhominal Would you happen to have any thoughts about what else could be causing the slowdown discussed [here](https://github.com/gollum/gollum/issues/1078)?

jhominal:
@bartkamphorst Oh, I should really come back to this. I have one performance issue where, with a large number of files (IIRC, independently from the number of commits), the subpage lookup becomes pretty slow. I seem to remember that that issue is exacerbated by the use of rugged. I will look into it again and file at least an issue.

bartkamphorst:
Thanks for your quick response.

> the subpage lookup becomes pretty slow

Where by subpage you mean link?

bartkamphorst:
Funnily enough subpage lookup is especially bad if no subpages are present, because nil is not cached and therefore it goes through the entire page map twice (once for page extracting and once for page processing) for each subpage. There are three subpages (footer, header, sidebar), so for a repo with 1000 pages, it would perform 6000 checks per request for absolutely nothing.

dometto:
:astonished:

jhominal:
@bartkamphorst : Yes, that is exactly what happens on my wiki instance. I got around it by simply removing the subpages (with the intention of coming back to it later).

bartkamphorst:
Right. Another thing I just noticed is that sometimes gollum looks for `favicon.ico` thinking it might be a page, in which case it again goes through the entire page map twice... In my setup with 4000 pages this takes about half a second. :-1:

dometto:
@jhominal and @bartkamphorst: does either of you have a quick fix for the subpage issues? I'm looking to cut a new `gollum-lib` gem this weekend. No problem if the fix is not in of course, but then I'll just release immediately.

bartkamphorst:
I've been tinkering with a quick fix solution, but that's not ready yet.

jhominal:
@dometto I do not have anything, and will probably not be ready this weekend.

bartkamphorst:
Here's a rough proposal.
**The getters (instead of explicitly defining `#footer`, `#header` and `#sidebar`)**
```ruby
    SUBPAGENAMES.each do |subpagename|
      define_method(subpagename) do
        find_sub_pages if self.instance_variable_get("@#{subpagename}").nil?
        self.instance_variable_get("@#{subpagename}") == :empty ? nil : self.instance_variable_get("@#{subpagename}")
      end
    end
```
(where `SUBPAGENAMES = [:header, :footer, :sidebar]`)

So, the idea is that if no subpage was found, we set it to `:empty` instead of to `nil`. This prevents gollum from going through the loop again during processing.

**The `#find_sub_pages` method**
```ruby
    def find_sub_pages(subpagenames = SUBPAGENAMES, map = nil)
      dir = ::Pathname.new(::File.dirname(self.path))
      map ||= @wiki.tree_map_for(@wiki.ref, true)
      # Remove the bulk of the pages
      map.select!{|entry| entry.name.to_s.match /^_(#{SUBPAGENAMES.map(&:capitalize).join("|")})/}

      if map.empty?
        subpagenames.each{|subpagename| instance_variable_set("@#{subpagename}", :empty)}
      else
        subpagenames.each do |subpagename|
          # Find all entries with a particular subpage name (there can be multiple)
          candidate_entries = map.select{|entry| entry.name.to_s.match subpagename.to_s.capitalize }
          next if candidate_entries.empty?
          selected_entry = nil
          # Look for the most specific (nested) subpage we can find
          loop do
            path = dir.join("_#{subpagename.capitalize}")
            selected_entry = candidate_entries.find{|entry| path == ::Pathname.new(entry.path).sub_ext("") }
            if selected_entry
              instance_variable_set("@#{subpagename}", selected_entry.page(@wiki, @version) )
              break
            else
              # drop down a directory
              dir = dir.dirname
            end
            break if dir.to_s == '.'
          end
          instance_variable_set("@#{subpagename}", :empty) if selected_entry.nil?
        end
      end
    end
```

I'm not sure yet whether this code actually captures all edge case, but I'd be interested to hear your thoughts @jhominal and @dometto .

dometto:
Nice. Have to take a more detailed look later, but here's a quick comment: rather than setting the instance variables to `:empty?`, I think you can just check `defined?(:instance_var)`. That will return false even if the variable was set to `nil`. That may simplify things a bit.

My other thought is that we may want a more generic way of getting a bunch of pages based on a list of pathspecs in one iteration (to improve performance elsewhere). Not sure if there would be any overlap with this.

bartkamphorst:
> I think you can just check defined?(:instance_var)

We could, but that doesn't play nice with the meta-programming magic for the getters (`defined?` thinks `"@#{subpagename}"` is an expression because it sees a string). We'd have to fall back on separate, explicit getters, but they would be sweet and short:
```ruby
    def sidebar
      find_sub_pages unless defined?(@sidebar)
      @sidebar
    end
```
> My other thought is that we may want a more generic way of getting a bunch of pages based on a list of pathspecs in one iteration (to improve performance elsewhere).

Absolutely, especially since each page link currently causes the page map to be searched. So in a repo with 1000 pages and 10 page links to be resolved in a page, the worst case is that 10000 checks are performed. Per request. :disappointed: This absolutely needs to be addressed, but I was focusing first on making subpage lookup more efficient for the short term.

dometto:
Restarted the tests and it works except for a problem on `2.3.0` unrelated to this PR (see https://github.com/gollum/gollum/issues/1110). Merging now. Thanks again for all the effort, @jhominal!

jhominal:
Thank you for the compliment, and thank you in return for Gollum itself.
I guess we should discuss the subpage issue on another ticket. (According to my testing, using grit, I see a 60% reduction in processing time on the sample 4k commits/4k pages wiki)

PR #1094, opened 2015-11-29
Update kramdown and useragent dependencies.

bartkamphorst:


dometto:
Thanks, that was long overdue. Did you also try upgrading the mustache 
dependency? Did that break anything?

bartkamphorst:
I didn't, as I recalled that there was an issue with mustache v.1.0.2 requiring Ruby version ~> 2.0.

dometto:
Ah, I see. We should decide for how long we still want to support ruby 
1.9.x then. Ruby 2.3.0 preview has just been released.

rgroux:
@dometto 

in fact distrib use 
 - rhel 
    - 7 -> ruby 2.0.0
    - 6 -> ruby 1.8.7 => need scl
  - ubuntu
    - LTS -> ruby 1.9+2.0
  - debian
    - oldstable -> ruby 1.9.3
    - stable -> ruby 2.1
    - testing/sid -> ruby 2.2
  - fedora
    - 23 -> ruby 2.2
    - 22 -> ruby 2.2
    - 21 -> ruby 2.1

It seems every "modern" distro has a ruby >2.0 , if rhel 6 you already need software collection, it's just about the oldstable debian distro (ruby2 is not into backport)

I think it's safe, but not my choice :)

enjoy 
/Richard

PR #1097, opened 2015-12-03
Css to sass

dometto:
Let's see whether this branch can still be merged easily.


bartkamphorst:
Ping @rtrvrtg .

rtrvrtg:
I haven't really been working on the Live Preview as of late, so if this gets merged in I don't really mind - probably better to get the SASS conversion in so other people can use it. If I do come back to it, I'll just pick up from this point.

dometto:
Now get this warning when running `bundle install` with this branch:

```
# WARNING : sinatra-assetpack is not maintained anymore! #
```
Should we try to move to something else (e.g. sprockets) immediately, or first merge this branch as is?

bartkamphorst:
Right... That's a deal breaker for me. :disappointed:

dometto:
Well, most of this branch is in generating the SASS files, which we will need anyway. Hopefully we can use sprockets just by modifying the gemspec and `app.rb`.

dometto:
I have a local version working with https://github.com/kalasjocke/sinatra-asset-pipeline. That gem doesn't look super active, but it seems like many people use it with sinatra. It also has some added bonuses in the form of sprockets helpers, which we should be able to use to clean up the layout files at some point (using `stylesheet_tag` helpers and the like).

Would that be ok in your opinion? /cc @sunny

bartkamphorst:
No sprockets 3.x support as of yet.

dometto:
True, but how badly do we need that?

dometto:
Neither sinatra-assetpack nor asset-pipeline seem to be maintained actively. The question for us is: is an unmainained, but more or less functional asset-pipeline better than the manual setup we have now?

bartkamphorst:
We have our answer: `sinatra-asset-pipeline` is alive and well. 😄 It now supports `sprockets 3.x`.

dometto:
Great! I'll open a new PR with my local branch then!

dometto:
I've added the commit that implements `asset-pipeline`. Once a new version of that gem is released we will have to use that. Now getting some mostly minor merge conficts on this PR. The only serious one is this:
```
CONFLICT (modify/delete): lib/gollum/public/gollum/css/gollum.css deleted in css-to-sass and modified in HEAD. Version HEAD of lib/gollum/public/gollum/css/gollum.css left in tree.
```

bartkamphorst:
Do we need an asset pipeline? What about http://sass-lang.com/documentation/Sass/Plugin/Rack.html ?

dometto:
Yes, that is an option as well! Advantage of using `sinatra-asset-pipeline` would be that we could also use CoffeeScript.

bartkamphorst:
Do we want coffeescript?

dometto:
No idea :) If the Rack implementation is the simplest we could go for that. If we want coffeescript we can always switch.

dometto:
Although with asset-pack we get sprockets, and thus also its helper methods like `require_tree` in JS, which is potentially useful. Also out of the box minification. Downside is it looks more complicated to setup, and I don't know how well it really plays outside of rails.

thomaskuntzz:
I don't have any experience with the integration of SASS in a 'production' project, but the rack plugin looks good, simple and to the point. But maybe we could minify the assets (Stylesheets, JS) too? I don't know if i'ts worth the hassle, but it's the only reason I see that could make us use something else...

thomaskuntzz:
I just saw that the SASS rack plug-in can output minified CSS with the appropriate option. So maybe we can get away with just using the rack plug-in, it's the simplest and I think we'll be just fine with it.

See http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style

dometto:
Ah, thanks! :+1:

dometto:
Closing in favor of #1247.

PR #1098, opened 2015-12-06
Sanitize basepath options when adding with more than one / at the start Fixes #1066.

### LINKED ISSUE #1066: Infinite redirect after escaping base path
Now that #1054 is merged, there's a new issue of infinite redirect when a base path is set that needs to be escaped.

To reproduce:
* `gollum -b "//wiki"` (one slash is alright; if two, the latter slash needs to be escaped before being used)
* Open Gollum in browser, e.g. `http://localhost:4567`.

andrehjr:
Just takes away all slashes at the start of a basepath instead of just removing the 'first' one.

Not sure how to add a proper spec for this, as the change was made directly at the cli.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -71,10 +71,8 @@ opts = OptionParser.new do |opts|
     "Example: setting this to '/wiki' will make the wiki accessible under 'http://localhost:4567/wiki/'.") do |base_path|
       
     # first trim a leading slash, if any
-    if base_path.start_with?("/")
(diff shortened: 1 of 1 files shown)


dometto:
The test failure seems to have nothing to do with this code. @bartkamphorst, agree to merge this? I don't have time to fix 1.9.3 on travis.

bartkamphorst:
@dometto All good!

Issue #504, opened 2012-09-01
Fix markdown headings in editor


bootstraponline:
@victorhooi mentioned this on #496

> Also - with the heading shortcuts (Ctrl+1, Ctrl+2 etc.) - it just seems to insert the hash (#) characters wherever your cursor - however, the intelligent behaviour would actually be to insert it at the beginning of the line, since that's where you'd normally put the heading markup.

PR #1101, opened 2015-12-21
Work in progress on Javascript editor

AnvilStriker:
In-progress changes in this branch ("editor-functions-fix"), not yet ready for merging:
- Fix editor misfeatures related to "backreferences," and loss of cursor position when applying functions that append text;
- New editor functions for reStructuredText;
- Introduction of QUnit testing framework, and a handful of tests (incomplete) for markdown and reStructuredText editor functions;
- Sinatra endpoint for invoking tests.

If possible, I would like someone to review my approach and suggest improvements if anything I've done so far clashes with established design philosophy.

The small fixes to `gollum.editor.js`, and the editor functions in `langs/rest.js` are straightforward and should be unremarkable. (Note, though, that in the current commit I have left JS console debugging turned on in the editor, and have not updated the help text in `rest.js` -- it's still the help text for markdown, since I cloned `rest.js` from `markdown.js`, but haven't updated the text yet).

Of greater possible concern is the way I've implemented the invocation of JS tests. In `app.rb` I've created a new endpoint called `jstest`, which takes an additional pathname element identifying the markup language whose test suite is to be run. For example:

```
http://localhost:4567/wiki/jstest/rest
```

runs the test suite for reStructuredText. Note that saving is disabled in the jstest editor view, although I haven't added the CSS necessary to "grey out" the "Save" button.

Running a test suite for a supported markup language that has no test suite defined results in "0 of 0 tests passed." Running a test suite for an unsupported or unknown markup language results in a 404 Not Found. 

This all seems to work quite nicely, but the question is whether this is really the _right_ way to invoke the tests, or if there's a better way that (a) doesn't use up a perfectly good endpoint name, and (b) could possibly run all defined test suites for supported languages at a single go.

---

A second question, completely orthogonal to the first, is: 

Do the editor functions (across all the supported markup languages) work the way that a user would really want them to in terms of (a) whether they produce markup that will render correctly without further input from the user, and (b) where they leave the cursor or highlighted region?

Consider the following cases:
1. The "HR" function for markdown splits the current line at the cursor (or end of the current selection), and puts a sequence of three asterisks on a line by themselves at the split. However, it only puts a single newline before and after the asterisks, meaning that if the user goes immediately to the preview, the asterisks don't get rendered as a horizontal rule if there wasn't a newline immediately before the cursor position (or end of the selection) when the function was invoked. In this case the user needs to manually add the newline before the horizontal rule will render correctly. Should the HR function instead ensure that there is always a blank line before the asterisks? And if so, how sophisticated should it be? Is it okay to blindly add the extra newline even if there had already been a newline before the cursor? (See my implementation of "HR" for reStructuredText for an example of such a "blind" approach.)
2. The "HR" function also has the annoying property that it moves the cursor all the way to the end of the editor buffer. I've tentatively fixed this by changing any "append-only" editor function to put the cursor at the _end_ of the appended fragment of text. But is this the right behavior? Might it make more sense to set the current selection to the entirety of the appended (so, for example, if the user is not happy with it, it's a simple matter of backspacing to get rid of it).
3. See also the behavior of the UL and OL functions for both markdown and reStructuredText, which can also produce results that need the same kind of additional user input as in case 1 above in order to yield correct rendering.

Would be curious to know if there's a "desired behavior" for these aspects of the editor functions, perhaps already implemented by some existing editor, that I should be trying to emulate.

### DIFF
--- a/lib/gollum/public/gollum/javascript/editor/tests/harness.js
+++ b/lib/gollum/public/gollum/javascript/editor/tests/harness.js
@@ -1,0 +1,51 @@
+var gollumEditorTests = {
+    Text: function(text, sel_start, sel_end) {
+        this.text = text;
+
+        if ( typeof sel_start == "number" )
+            this.selStart = sel_start;
+        else
+            this.selStart = this.text.indexOf(sel_start);
+
+        if ( typeof sel_end == "number" )
+            this.selEnd = sel_end;
+        else
+            this.selEnd = this.text.indexOf(sel_end);
+    },    
+
+    editorFunctionTest: function(assert, func, 
+				 text, sel_start, sel_end,
+				 expected, exp_start, exp_end) {
+
+        // Append some text to the buffer.
+        // Select a portion (possibly empty) of the text appended.
+        // Apply an editor function to the selected text, or at point.
+        // Check that the applied function had the expected effect,
+        // i.e., that the resulting text is the same as `expected`,
+        // and that the resulting selection has the expected start
+        // and end indexes (`exp_start` and `exp_end`).
+
+
+	var $editorBody = $("#gollum-editor-body"); // JQuery object
+        var editorBody = $editorBody[0];            // DOM object
+        var content = $editorBody.val();
+        var start = content.length;
+        $editorBody.val(content + text + "\n\n");
+
+        var t0 = new gollumEditorTests.Text(text, sel_start, sel_end);
+        var t1 = new gollumEditorTests.Text(expected, exp_start, exp_end);
+
+	editorBody.focus();
+	editorBody.select();
+	editorBody.setSelectionRange(start + t0.selStart, 
+                                     start + t0.selEnd);
+        $("#function-"+func).click();
+
+        assert.equal( $editorBody.val().slice(start, -2), t1.text, 
+                      "expected markup" );
+        assert.equal( editorBody.selectionStart, start + t1.selStart, 
+                      "selection start should be " + (start + t1.selStart) );
+	assert.equal( editorBody.selectionEnd, start + t1.selEnd, 
+                      "selection end should be " + (start + t1.selEnd) );
+    }
+};
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -281,6 +281,16 @@ module Precious
       redirect to('/')
     end
 
+    get '/jstest/*' do
+      name = params[:splat].first
+      halt 404 if !Gollum::Markup.formats().keys.include?(name.to_sym)
+      wikip = wiki_page(name)
+      @name = wikip.name.to_url
+      
+      mustache :jstest
+    end
+
+
     get '/create/*' do
       forbid unless @allow_editing
       wikip = wiki_page(params[:splat].first.gsub('+', '-'))
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
@@ -13,7 +13,7 @@
     EditorMode: 'code',
     NewFile: false,
     HasFunctionBar: true,
-    Debug: false,
+    Debug: true,
     NoDefinitionsFor: []
   };
   var ActiveOptions = {};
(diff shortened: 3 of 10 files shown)


bartkamphorst:
@AnvilStriker I won't have much time this week to review, but I do want to quickly say that I appreciate your efforts! Especially testing with QUnit is a great addition that was long overdue. I wonder though whether it would make sense to pull the different threads apart in separate PRs. That way we could merge for example the backreferences stuff, even if we are still discussing best testing practices or the proper editor behavior. If this was your plan all along and you were just offering us a glimpse into your work so far, please ignore this comment. :)

AnvilStriker:
@bartkamphorst Ideally, yes, splitting this work into separate issues/branches would be desirable, but it may not be easy to tease apart the various concerns, which are quite wrapped around one another.

I encountered the backreference issue as I was implementing the editor functions for reST (the original topic of this PR), which I couldn't complete without fixing that issue. Yet, fixing the backreference issue *already* changes the behavior of many existing editor functions, whether or not we have the discussion about what's proper editor behavior. 

I suppose we could consider the QUnit testing stuff a truly separable piece. On the other hand, going without test suites for each language's editor functions means that we have no systematic way of assessing the impact of changes (such as the backreference fix or the new cursor-position-after-append fix) made to the editor's common function-application logic. (Maybe this is okay, though, as long  as we lack a statement of what the desired behavior really is.)

Anyway, I'm happy to partition this work in any way that makes sense. I'm just having some trouble seeing which way that is. :-)

bartkamphorst:
>  but the question is whether this is really the right way to invoke the tests, or if there's a better way 

What about a headless test runner like [qunited](https://github.com/aaronroyer/qunited)? Is even compatible with JRuby it seems.

AnvilStriker:
> What about a headless test runner like qunited? Is even compatible with JRuby it seems.

That's pretty nifty. Will definitely look into it, thanks!

lib/gollum/public/gollum/javascript/editor/tests/harness.js
```diff
@@ -0,0 +1,51 @@
...
+        // Select a portion (possibly empty) of the text appended.
+        // Apply an editor function to the selected text, or at point.
+        // Check that the applied function had the expected effect,
+        // i.e., that the resulting text is the same as `expected`,
+        // and that the resulting selection has the expected start
+        // and end indexes (`exp_start` and `exp_end`).
+
+
+	var $editorBody = $("#gollum-editor-body"); // JQuery object
+        var editorBody = $editorBody[0];            // DOM object
```
bartkamphorst: Strange indentation.

bartkamphorst:
> I'm happy to partition this work in any way that makes sense.

And I'm happy to consider the whole thing together. :) I just would hate to see the whole thing delayed because of one component not being ready yet (we've seen that happen before, and it always feels wasteful). But let's just keep going like we are now!

bartkamphorst:
Question: do you see any way to DRY-ify the "standard" editor tests (bold, italics, etc.) so that each language at least goes through the same basic tests without having multiple copies of them? Obviously each input would be different, but the expected outcome would probably be the same, or not?

bartkamphorst:
Regarding your second, three-part question, I think that

1. I'd prefer a slightly more involved approach as I suspect that adding newlines blindly could lead to more white space in the preview than users would expect to find;
2. The cursor behavior you proposed makes a lot sense to me and I would like that to be the default editor behavior;
3. Same as 1.

@dometto , @sunny Any additional thoughts on this?

dometto:
@bartkamphorst, agree with your points. thanks again @AnvilStriker!

dometto:
#1097 makes substantial changes to the way gollum handles css in general, and also to the editor css. However, it doesn't seem to change any JS, and I can merge it with this PR branch locally. To be sure: @AnvilStriker, do you think it would make your life more difficult if I merged #1097 already?

AnvilStriker:
@dometto - I'm inclined to agree that it probably will not affect me much, if at all. All I can see in the near term is that I might have some adjustments to make to the mustache template I created for QUnit testing. And in the longer run, that template may not survive, if we decide to go with a headless test runner (though I still need to do some more reading to be sure).

In any case, I'd say go for merging #1097.

lib/gollum/public/gollum/javascript/editor/tests/harness.js
AnvilStriker: Seems to be an artifact of the GitHub code viewer. That indent is not present in my emacs editor buffer.

bartkamphorst:
@AnvilStriker Would you be able to give an indication of how involved it would be to fix [this bug](https://github.com/gollum/gollum/issues/504) in the editor?

dometto:
What's the status of this PR?

bartkamphorst:
@AnvilStriker I would love to incorporate your changes to the editor into gollum. Perhaps we should split this PR into smaller chunks after all; what do you think? I realize you're probably very busy, but if you could let us know where we stand with this it would be much appreciated!

bartkamphorst:
Closing this since https://github.com/gollum/gollum/commit/58225c7c2025856a70a7df61544309f26dccc2ef adds a rest language definition and https://github.com/gollum/gollum/pull/1348 fixes the cursor issues with markup insertions.

PR #1106, opened 2016-01-10
Adds links to youtube videos in README.

bartkamphorst:
Relates to https://github.com/gollum/gollum/issues/1105.


bartkamphorst:
Alternatively, we could reference them in the wiki, but this is more visible.

dometto:
I think this makes sense!

Issue #1118, opened 2016-03-18
mermaid support


dhalsim:
Hello,

I wanted to add [mermaid](http://knsv.github.io/mermaid) diagrams to wiki pages like this one [Haroopad ](http://pad.haroopress.com/page.html?f=how-to-draw-diagram) does.
But if I add the code snippet of mermaid to a markdown page this HTML is created:

```html
<span class="err">graph TD;
    A--&amp;gt;B;
    A--&amp;gt;C;
    B--&amp;gt;D;
    C--&amp;gt;D;</span>
```

I'm no expert at Ruby and can't understand the exact place to look in the source code (and its dependencies). I need a basic guidance here.

P.S: I know gollum uses PlantUML for this purpose but I like to use mermaid if possible.

Thanks,

dometto:
I suspect this is an issue with html sanitization. You should whitelist whatever elements mermaid uses. Use the --config option to specify a config file and see:

https://github.com/gollum/gollum/wiki/Security#customization

rgroux:
awesome lib, 
very easy to integrate (nothing to do)

you can use it in 6 steps, hope someone (maybe me) will create a gem and use it as gem or maybe as native(far more better)

GIT_REPO=<path to your wiki git repo>

1. download the mermaid.js and memaid.css from github
1. move and rename mermaid.js -> $GIT_REPO/custom.js
1. move and rename mermaid.css -> $GIT_REPO/custom.css
1. git add custom.css custom.js
1. git commit custom.css custom.js -m 'feature(js): add support for memaid diagrams'
1. start gollum with --js --css (bundle exec bin/gollum --js --css $GIT_REPO)
1. enjoy

for example create a new page and add this

```html
<div class="mermaid">
graph RL
    A[Hard edge] -->|Link text| B(Round edge)
    B --> C{Decision}
    C -->|One| D[Result one]
    C -->|Two| E[Result two]
</div>
<div class="mermaid">
graph LR
    A[Hard edge] -->|Link text| B(Round edge)
    B --> C{Decision}
    C -->|One| D[Result one]
    C -->|Two| E[Result two]
</div>
```

dhalsim:
Yeah, it worked like a charm :+1: 
 
Thanks

rgroux:
@dhalsim 

i create a pull request #1119 to add mermaid js inside gollum and use it natively without custom.[js|css].

I hope it will be accepted in the upstream :)

/R

wf5w:
[...]

PR #1119, opened 2016-03-25
Add mermaid js. Issue #1118

rgroux:
Mermaid is under MIT License, imported from [mermaid dist](https://github.com/knsv/mermaid/tree/master/dist)

It can be useful to use mermaid js instead of plantulm or jumly, it's under MIT license and 100% javascript.
With mermaid embed inside gollum, the wiki can server diagram natively without any external tool.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -145,6 +145,9 @@ MSG
     "If not specified, uses the '<git-repo>/mathjax.config.js' file.") do |file|
     wiki_options[:mathjax_config] = file || 'mathjax.config.js'
   end
+  opts.on("--plantuml-url [URL]", "Sets the PlantUML server endpoint.") do |url|
+    wiki_options[:plantuml_url] = url
+  end
   opts.on("--template-dir [PATH]", "Specify custom mustache template directory.") do |path|
     wiki_options[:template_dir] = path
   end
@@ -162,7 +165,6 @@ MSG
   end
   
   opts.separator ""
-  
 end
 
 # Read command line options into `options` hash
@@ -208,6 +210,12 @@ if options[:irb]
     if !wiki.exist? then
       raise Gollum::InvalidGitRepositoryError
     end
+    if wiki_options[:plantuml_url]
+      Gollum::Filter::PlantUML.configure do |config|
+        puts "Using #{wiki_options[:plantuml_url]} as PlantUML endpoint"
+        config.url = wiki_options[:plantuml_url]
+      end
+    end
     puts
     puts "Loaded Gollum wiki at:"
     puts "#{File.expand_path(gollum_path).inspect}"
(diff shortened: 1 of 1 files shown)


bartkamphorst:
@rgroux I like the idea of not having to depend on an external service for diagram rendering. My concern however is the maintainability of the mermaid code. Is there a mermaid gem we could include in our dependencies?

Also, I am curious if mermaid could supplant plantUML (pun intended). @hsanson , @j4hangir, @ace03uec I know you use PlantUML in your wikis. Would there be any principled or technical reason why you couldn't (or wouldn't want to) transition to mermaid diagrams?

hsanson:
First time I hear about mermaid js and having diagrams without need of external services is indeed a plus.
The downside is that PlantUML has far more diagram types that we depend on (State, Activity, UseCase, etc) and that currently we have thousands of diagrams done in PlantUML making a move impractical.

dometto:
Sounds like there is a solid use case for both plantuml and mermaid, the former for the more complicated work and the latter for simpler stuff. Maybe we could support both (but no other diagram software), unless that has performance impact?

I agree with the caveat that the JS should be easy to upgrade, so yes, a simple mermaid gem would be great.

dometto:
To be clear, this would mean dropping support for WSD.

rgroux:
@All 

IMAO i think both are  necessary. may be in future we add new one if this new js is better than plantuml or mermaid js, or can use the spec of both. but actualy i think plantuml is nice for complex diagram with external software, but for 80% of usecase mermaid is good enough.

for create a gem, it can be great but i am not sure how to create a gem to override the 3 mustaches files (livepreview, file_view and layout)

rgroux:
@self, @all 

the PR #762 is not useful because i don't want to change the template dir, only override/add some features.

is there a methode to add some line into mustcahe templates?, or we need to add a process to do that.

like a conditional into mustache file if :extended_js_stuff -> parse an hash ["path_1","path_2] and use the css/js and specified inside these path

actualy i dont know/ how to do it, it s only an open thought

rgroux:
hi

what is the status of this PR?

actually is not possible to override the layout mustache so i don t know how do make a gem plugin for this.

dometto:
@rgroux, we would still like to be able to not have to handle mermaid updates manually, if at all possible. If a gem implementation is not possible, what about adding a rake task to update, perhaps in combination with a submodule? @bartkamphorst. would you find that acceptable?

hsanson:
I found  [asciidoctor-diagram](http://asciidoctor.org/docs/asciidoctor-diagram/) gem that seems to support plantuml
without need to use an external server/service. It seems to use the plantuml.jar file directly to generate the diagrams.

You still need to have java runtime installed and available in your path, but there is no need for setting additional servers. This eliminates
the only downside I have with plantuml.

Just though this information is relevant to the discussion. Not sure how difficult would it be to use this gem with Gollum.

bartkamphorst:
> what about adding a rake task to update, perhaps in combination with a submodule? @bartkamphorst. would you find that acceptable?

I'm not in favor of introducing more maintenance tasks and procedures to gollum.

Mic92:
This is cool! Looking forward to a release!

Issue #1125, opened 2016-04-15
EmojiOne for Gollum


svoop:
I know, not everybody likes emoji. But they do come in handy at times. Here are just a few of my use cases:

* In an international or multilingual context, I like to mark parts with the corresponding flag such as :us: or :fr:.
* Bits of information can easily be highlighted with :warning:, :bulb: or alike.
* Todo lists can be ticked with :white_circle: and :red_circle:

#722 was closed due to the lack of an open license source of emoji, but these days, we have [EmojiOne](http://emojione.com) (licensed under CC).

I've implemented basic support for emoji as PNG images in order to support as many browsers as possible. However, it takes modifications in both gollum and gollum-lib, hence this issue to bundle.

Thanks for considering this addition!

svoop:
Whoops, I've broken some tests, working on it.

dometto:
Seeing as @dekimsey and @bootstraponline already ok'd the feature in the past, I'm not against this. @bartkamphorst?

In general, it would be great to have some benchmarks for the various parts of the filter chain, so we can assess the performance impact of adding a particular filter. It's a fairly expensive regex there, I think. Perhaps we can turn emoji's off by default?

svoop:
> Perhaps we can turn emoji's off by default?

Sure, good idea!

svoop:
I've added `--emoji` (default: false) which kicks emoji from the filter chain unless set.

svoop:
@dometto Is there anything else I can do in order to get this merged?

svoop:
@bartkamphorst Just a heads up: In order to merge the emoji feature, there are two PR:
* #1126
* gollum/gollum-lib#203

PR #1126, opened 2016-04-15
Support for emoji

svoop:
See #1125 for more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ For a quick impression of gollum, see [this video](https://www.youtube.com/watch
 
 ## SYSTEM REQUIREMENTS
 
-| Operating System | Ruby           | Adapters           | Supported | 
+| Operating System | Ruby           | Adapters           | Supported |
 | ---------------- | -------------- | ------------------ | --------- |
 | Unix/Linux-like  | Ruby 1.9.3+    | all except [RJGit](https://github.com/repotag/rjgit) | yes |
 | Unix/Linux-like  | [JRuby](https://github.com/jruby/jruby) (1.9.3+ compatible) | [RJGit](https://github.com/repotag/rjgit) | yes |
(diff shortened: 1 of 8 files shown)


dometto:
Thanks for your work on this! Regarding the switch: in general I want to be careful in adding new switches, as I explained in https://github.com/gollum/gollum/issues/1127#issuecomment-210834450. But since emojis are popular these days, perhaps this merits a user-friendly switch. What do others think?

svoop:
@dometto In case the switch question is blocking the merge: I don't mind removing it for the sake of not having too many switches. The `config.rb` is way more useful anyways :smile:

bartkamphorst:
Can we use any of the existing EmojiOne gems so that we don't have to bundle the images with gollum? I'm worried about the maintenance of these images.

svoop:
I've tried that at first but for some reason I don't remember had to drop it. I can take another look though.

bartkamphorst:
That would be great. 👍

svoop:
@bartkamphorst Here you go: The emoji are now served via their own route `/emoji/:name` which streams the PNG directly from the gem. 

By the way: Any particular reason why you have Travis test against `jruby-19mode`? I had to use a rather ugly workaround `"string".each_byte.to_a` for the Ruby 2 equivalent `"string".bytes` in one of my tests to pass in 1.9 mode.

bartkamphorst:
> The emoji are now served via their own route /emoji/:name which streams the PNG directly from the gem.

Very cool. 👍 I'll try to find time soon to examine the code in detail.
 
> Any particular reason why you have Travis test against jruby-19mode? 

Yes, because when we added jruby it ran in 1.8.7 mode by default... I take it from your comment that that has changed. 😉

bartkamphorst:
PR looks very clean to me. 👍 Okay to merge, @dometto?

Issue #1127, opened 2016-04-16
New option --allow-style


svoop:
For many wikis strict sanitation is the way to go, however, when working e.g. with complex tables, the `style` attribute is very useful. Of course, there's `--css` which is great for wiki-wide styles, but some styles are strictly bound to the content on a particular page.

I've implemented a `--allow-style` option, what do you thing of this addition?

dometto:
I see the use case, but I'm not sure this needs its own switch, also considering that are all already quite a lot of switches. Can't this be accomplished simply by using a `config.rb`? If so, we should put that in the wiki, and keep it at that.

svoop:
Absolutely, I guess the same applies to #1126?

svoop:
@dometto Same here: Is there anything else I can do in order to get this merged?

bartkamphorst:
I'm in agreement with @dometto here that a snippet in `config.rb` is probably sufficient to add `style` to the whitelisted attributes.

svoop:
@bartkamphorst Actually the corresponding PR #204 doesn't feature an option anymore but is `config.ru`-able only. (I'll change the title of this issue accordingly.)

svoop:
Since the option has been dropped, no changes on gollum are necessary, but only on gollum-lib (see https://github.com/gollum/gollum-lib/issues/204). Closing...

PR #1128, opened 2016-04-16
Add --allow-style option to use style attributes in HTML tags

svoop:
see #1127


svoop:
No swich for this as suggested by @dometto

Issue #907, opened 2014-12-02
Tables are not being rendered properly


nlowe:
I just updated to 3.1 to see if that corrects the issue. It does not.

Given the following sample table

``` markdown
| P | Q | P * Q |
| - | - | - |
| T | T | T |
| T | F | F |
| F | T | F |
| F | F | F |
```

Gollum renders it as one line, not as a table:

![gollum-render](https://cloud.githubusercontent.com/assets/794251/5263363/4c557d78-79fe-11e4-95f8-2139437217fd.png)

[atom](http://github.com/atom/atom) renders it like a table:

![image](https://cloud.githubusercontent.com/assets/794251/5263357/2a0dd5b2-79fe-11e4-8b79-65dc2e1c7b1a.png)

According to the [GFM wiki](https://help.github.com/articles/github-flavored-markdown/#tables), tables are supported.

VladUreche:
I'd like to open this again, for gollum 4.0.0 -- it's not even rendering the table above.

bartkamphorst:
@VladUreche Which markdown gem do you have installed?

VladUreche:
@bartkamphorst, thanks for the quick answer! This is what I get with `--version`:
```
$ gollum --version
Gollum 4.0.0
```
Is there a better way to check? (sorry, Ruby noob, I come from the JVM world)

bartkamphorst:
@VladUreche No worries, welcome to the Ruby community. :+1: Gollum depends on [github-markup](https://github.com/github/markup) for rendering markups, which in turn uses [kramdown](https://github.com/gettalong/kramdown) as its default markdown renderer. If I do an out-of-the-box install of gollum with `gem install gollum` the example table renders just fine.
![gollum_table](https://cloud.githubusercontent.com/assets/571173/8125699/7195fcb2-10e8-11e5-8918-bf956dbf1cd4.png)
Which makes me wonder about your particular setup. Could you elaborate a bit, for example on  how you installed gollum? Did you change any of the default settings? Is the page (with the table) that you are trying to render a markdown page?

dometto:
`gem install gollum` should automatically install kramdown as markdown renderer. You could try doing `gem install github-markdown` and see if it renders correctly with that (@bartkamphorst, is it possible that you were accidentally using github-markdown in your test?).

bartkamphorst:
[...]

PR #1132, opened 2016-05-18
rendering markdown

carmen0208:
Proposal to fix issue: "Tables are not being rendered properly" 
see [#907](https://github.com/gollum/gollum/issues/907) for more.


bartkamphorst:
There's no need to introduce `github-markdown` as a gem dependency. Just make sure you have it installed and gollum will automatically use it if it is found. If you have a fix so that kramdown renders the tables from #907 correctly, then please let us know.

PR #1141, opened 2016-06-25
Custom auth

jradd:
I am new to git workflow, so I hope I am doing this right.


bartkamphorst:
@jradd Technically you are doing it right, although typically you'll want to squash your intermediate commits into one big one with a descriptive commit message before opening a PR (so not to clutter the project's commit history once the PR is merged). Also, when opening a PR, it is good practice to explain (in full sentences) what the PR is about. In this case I can guess, but unfortunately I think I will have to disappoint you, as gollum does not provide any type of authentication by design. There is another project called [omnigollum](https://github.com/arr2036/omnigollum) that might be of interest to you. If I misunderstood your intentions with this PR, however, feel free to elaborate.

jradd:
Thank you so much for taking the time to explain this to me. I will be sure to follow your guidlines for any future PR.

bartkamphorst:
You're welcome! 👍

PR #1143, opened 2016-06-27
Fix compare page not accessible in no-edit mode. Resolves #1140.

### LINKED ISSUE #1140: Allow "Compare changes" in no-edit mode
When `no-edit` is set, accessing a `/compare/:page` returns:

> Error
> Forbidden. This wiki is set to no-edit mode.

I don't have a lot of context here, but I don't see why these are tied together.
In particular, my team is starting to use Gollum but we want to disable edit mode to use a peer-review process on changes (therefore adding/editing a page would be made on the git repo itself). We still want to be able to see the history though.

It would be great to have an option, such as `--disallow-compare`, that would control this independently.

I can look to open a PR if you are OK with this idea.

astorije:
Permission checking was spread across `post` action handlers instead of inside the `before` to normalize between `get` and `post` action handlers and be more explicit.

This resolves #1140.

#### Result

screen shot 2016-06-27 at 13 52 24

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -96,7 +96,6 @@ module Precious
     before do
       settings.wiki_options[:allow_editing] = settings.wiki_options.fetch(:allow_editing, true)
       @allow_editing = settings.wiki_options[:allow_editing]
-      forbid unless @allow_editing || request.request_method == "GET"
       Precious::App.set(:mustache, {:templates => settings.wiki_options[:template_dir]}) if settings.wiki_options[:template_dir]
       @base_url = url('/', false).chomp('/')
       @page_dir = settings.wiki_options[:page_file_dir].to_s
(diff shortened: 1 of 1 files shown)


astorije:
@bartkamphorst, ping? :-)

bartkamphorst:
I was actually thinking along the lines of having a `WHITELIST` constant or something and adding a check tot the `before` block to see if the requested page is whitelisted or not. @dometto ?

bartkamphorst:
@dometto What are your thoughts on this?

bartkamphorst:
Ping @dometto .

dometto:
I don't particualrly like having to remember the `forbid` when we create new `POST` actions (nor for some `GET` actions, as is currently the case). But I don't see a much neater way of doing it for now. Perhaps we can refactor later, but just merge this now to fix the bug?

bartkamphorst:
Thanks @astorije .

PR #1148, opened 2016-07-11
Remove deprecated github-markdown markup

astorije:
From [RubyGems](https://rubygems.org/gems/github-markdown/versions/0.6.9):

> THIS GEM IS NOT MAINTAINED AND NOT SUPPORTED.

Also, the GitHub repository, maintained by GitHub, was removed.

The link is now a redirect, it doesn't point to the GFM page anymore.

Do you want this replaced with something else? My understanding is that kramdown comes with everything needed, not entirely sure if something needs to be done to enable GFM support there.


bartkamphorst:
No, this is good. Kramdown is used by default; the list only concerns additional markups. Thanks!

PR #1156, opened 2016-08-07
Merge rc

dometto:
`rc` still contains a commit we missed by @hsanson we haven't merged into `master`. Merging this and then deleting the `rc` branch in favor of using `5.x`.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -124,6 +124,10 @@ opts = OptionParser.new do |opts|
   opts.on("--h1-title", "Sets page title to value of first h1") do
     wiki_options[:h1_title] = true
   end
+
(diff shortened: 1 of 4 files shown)


dometto:
@hsanson sorry that we forgot to merge this file, and thanks for the original PR! If I'm not mistaken the Plantuml page on the wiki does not reflect this new way of configuring plantuml (with the command line switch), correct?

hsanson:
@dometto sorry for the late response.

I cannot find the discussion but my impression was that this PR was rejected because there are other ways to set the plantuml endpoint like using env vars or a config.rb file (see [here](https://github.com/gollum/gollum/wiki/Custom-PlantUML-Server#gollum-configuration)).

Now that it is merged I can update the wiki if required.

PR #1158, opened 2016-08-07
Handle GET /favicon.ico

### LINKED ISSUE #1157: Serve up favicon.ico if it exists
It'd be handy if gollum handled `GET /favicon.ico` specially so that, if `favicon.ico` exists in the wiki root, it serves it up, otherwise returns a `404`.

To do the same thing, I need to add a handler using `config.rb` then run gollum pointed to that config.

stevearm:
Serve up favicon.ico if it exists in the wiki root, otherwise return a 404.

Closes #1157 

I've never written ruby, but I moved this code from my `config.rb` file, where it worked fine. Could someone with a proper dev setup please do a `curl -vo /dev/null http://localhost:4567/favicon.ico` with and without a `favicon.ico` in the wiki root to ensure it works in this context?

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -110,6 +110,15 @@ module Precious
       redirect clean_url(::File.join(@base_url, @page_dir, wiki_new.index_page))
     end
 
+    get '/favicon.ico' do
+      if File.file?('favicon.ico')
+        content_type 'image/x-icon'
+        File.read('favicon.ico')
+      else
+        halt 404
+      end
+    end
+
     # path is set to name if path is nil.
     #   if path is 'a/b' and a and b are dirs, then
     #   path must have a trailing slash 'a/b/' or


stevearm:
Um, I don't know why, but this works out of box on my mac. And on retest, it looks like it's working fine on my Ubuntu box where I originally "fixed" the problem. So never mind, as this looks like it's already implemented, and no need to change anything.

Thanks for the great software.

stevearm:
In a different repo, I'm now having the same problem I had originally. I'm at a loss. The sequence is:

1. Ubuntu machine (wiki 1) did not load favicon
2. Added config.rb to wiki 1, ran with `gollum -c config.rb` and ubuntu machine now loads favicon
3. Checked out wiki 1 to mac, ran with `gollum` and it loaded favicon
4. Deleted config.rb and mac continued to load favicon
5. Tried mac on wiki 2, and it does not load favicon

From this (without any experience or time to learn ruby, setup a dev env, and start to debug) I'm wondering:

* Does `config.rb` load by default somehow (is #1159 irrelevant)?
* Once `config.rb` has been loaded, is the resulting compiled (like how python has `.pyc` files) file cached somewhere, so running a second time with `config.rb` deleted has no effect?

dometto:
Possibly you just need to commit the file to the repo?

dometto:
I meant, commit that to wiki2. And make sure the file is checked out in both wiki's. Config.rb should be irrelevant here I think.

As for now, there are no plans to make gollum serve uncommitted files (although this may be a future feature so that one can preview pages written in a different editor). So serving the image should just work when it's property committed.

stevearm:
Ah, yes that's what I was missing. I thought gollum looked at the files, not at the git index. Makes sense now, thanks.

bartkamphorst:
Glad you got it sorted! 👍

PR #1170, opened 2016-11-09
Replace editor's content editor with Ace Editor

ProgramFan:
Hi, I tried to replace the content editor with the famous Ace editor for better wiki editing experiences. The following are the results:

Before:
before

After:
after

Here are some caveats currently, for example, dragging a file to the original textarea for uploading, snippets inserting, do not work currently.

Could someone give me some pointers whether it can be accepted in revised form to the upstream? And what should I do? I am new to the javascript world but I use gollum wiki a lot and really want to help improving it.

### DIFF
--- a/lib/gollum/templates/layout.mustache
+++ b/lib/gollum/templates/layout.mustache
@@ -30,6 +30,32 @@
   <script type="text/javascript" src="{{base_url}}/javascript/gollum.dialog.js"></script>
   <script type="text/javascript" src="{{base_url}}/javascript/gollum.placeholder.js"></script>
   <script type="text/javascript" src="{{base_url}}/javascript/editor/gollum.editor.js"></script>
+  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/ace/1.2.5/ace.js"></script>
+  <script type="text/javascript">
+$(function () {
+  $('textarea[id="gollum-editor-body"]').each(function () {
+    var textarea = $(this);
+    var mode = textarea.attr("data-markup-lang");
+    var editDiv = $('<div>', {
+      position: 'absolute',
+      height: textarea.height(),
+      'class': textarea.attr('class')
+    }).insertBefore(textarea);
+    textarea.css('display', 'none');
+
+    var editor = ace.edit(editDiv[0]);
+    editor.setTheme("ace/theme/tomorrow");
+    editor.setKeyboardHandler("ace/keyboard/vim");
+    editor.getSession().setUseWrapMode(true);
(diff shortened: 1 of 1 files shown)


victor-torres:
Nice, @ProgramFan.

You can improve your pull request by doing the following things:

- download and version the library into javascript/libs/ace-1.2.5.js
- download and version related theme files (syntax, parser, stylesheets...)
- create another javascript file to place the logic you've added to the mustache file

The two first points will make it possible to use this feature when running offline.

ProgramFan:
Thanks, @victor-torres. I like the idea of bundling ace into gollum and make it work offline. But I see gollum does not bundle MathJax when supporting math equations. So I don't know what the gollum developers' idea are. In fact, I use offline version on my own computer.

dometto:
Thanks for the PR @ProgramFan, and for the input, @victor-torres. I was experimenting myself with switching to ACE. The end goal would be to integrate a live-preview pane into the default gollum editor, to get rid of the two editor modes -- see https://github.com/gollum/gollum/pull/1048.

However, I also think it's absolutely essential to make sure the editor works in offline mode, and thus to bundle ace with gollum somehow. At the moment ace *is* being bundled, for the livepreview editor, but updating it is quite complicated. Ideally there would be a good way of bundling ace and updating it automatically. Last time I checked, there was still not a good way to handle JS dependencies in ruby, but perhaps that has changed in the mean time?

Mathjax is a different issue: in offline mode, the only consequence is that equations are not rendered, but otherwise gollum remains usable without a network connection.

ProgramFan:
Thanks for your comments, @dometto. I can improve the PR by bundling ACE and integrating the drag-and-drop uploading and fix other issues, if this PR can be accepted as an intermediate solution to the end goal.

As a gollum user, I would argue that integrating ACE in the plain mode (no live-preview) solves 80% of the problems. The editing experiences are significcantly enhanced. I do not use the live-preview mode, but one can still use the plain editor + live-preview even when ACE is bundled. Maybe if we have live-preview, the syntax highlightling is less important.

dometto:
It would be great to have this PR first, and worry about the live-preview pane later. We do want to get the live preview pane in in the end, because we just don't want to support the two different editor modes anymore (and getting rid of livepreview alltogether would be a shame).

Could you please work on bundling ACE in a neat manner, and then reopen this PR against the 5.x branch? Your contribution is much appreciated!

ProgramFan:
OK. I will do it.

ProgramFan:
See #1173 for further improvements. Would you please review it? @dometto

Issue #705, opened 2013-06-03
Show only supported formats in dropdown


schmunk42:
I was a bit confused after importing my mediawiki files, because they were not formatted correctly.
After installing the correct parser and restarting gollum everything worked fine, but I'd say it's better only to show only the supported file-types in the dropndown on the edit page.

bartkamphorst:
This has been resolved in the `5.x` branch. Unavailable formatters are grayed out. 👍

Issue #436, opened 2012-07-16
Wiki editor buttons break Ctrl + Z


stereobooster:
If you type text in editor then you can use Ctrl + Z to undo changes (changes separated by arrow keys). But if you use buttons from  wiki editor you will lose current change story.

moved from stereobooster/github-issues#12

dometto:
Resolved in https://github.com/gollum/gollum/pull/1173 (`v5.x`)

PR #1173, opened 2016-11-20
Replace the editor body with Ace editor

ProgramFan:
This PR replaces the editor body with Ace editor, so that editing experiences are enhanced. Full Ace editor functions are available, but some gollum editor functions does not work, including:

1. File upload by drag-n-drop.
2. Insert snippets by clicking help menu items.
3. Change docment format does not change editor's highlighting syntax (However, one can always use Ace's hotkeys to change it).

This is an improvement of #1170.  I am willing to improve it.


polyzen:
Works great. 👍

There are a bunch of other features Ace has. Are these perhaps configurable on a per-installation basis? Such as:
- Syntax highlighting
- Colorschemes
- Resize editor body on-the-fly (like current implementation)

PS: The gap above the editor body is no longer there.

ProgramFan:
@polyzen Currently the colorschemes, key bindings, and syntax highlighting languages can be changed directly via Ace's settings ('cmd+,' to invoke on macOS, 'ctrl+,' on Linux and Windows but does not work on my system). I can add options for colorschemes and key bindings. Syntax highlighting can be handled more gracefully. For the resize, I have not come up with an idea, so help is welcome. 😄

polyzen:
> can be changed directly via Ace's settings

Really cool.

> I can add options for colorschemes and key bindings.

The settings apparently only apply for as long at that editor is open, so this would be very appreciated.

> For the resize,

Found this from the [How-To](https://ace.c9.io/#nav=howto):
> Ace only resizes itself on window events. If you resize the editor div in another manner, and need Ace to resize, use the following:
>
> editor.resize()

polyzen:
A few more notes:
- Doesn't select mode automatically (for rST at least)
- It may be possible to [extend highlighting](https://ace.c9.io/#nav=higlighter) for Gollum syntax
- [Autocompletions can be enabled](https://github.com/ajaxorg/ace/wiki/How-to-enable-Autocomplete-in-the-Ace-editor)

Upstream issues?:
- Difficult to return to normal mode on mobile
- Cannot backspace on mobile

ProgramFan:
Thanks for your feedback, @polyzen. I will consider them when I find time.

dometto:
Thanks for the excellent PR @ProgramFan, and for the comments @polyzen. I haven't had time to respond for a while, but we're still very eager to include this in the `5.0` release of gollum. My $0.02:

* Extended highlighting for gollum syntax is very nice, but not required to merge this (can be added later)
* Same for autocompletion
* Selecting the right syntax mode *is* essential, as is resizing the editor.

With regard to the mobile issues: it would of course be great if that worked nicely, but again I wouldn't consider it a showstopper if it didn't -- we haven't ever consciously designed for mobile, so doing this rightly would probably require more restructuring of the CSS, which we should do in a separate PR IMO.

ProgramFan:
I will fix syntax mode selection and resizing.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -55,7 +55,7 @@
       window.ace_editor = editor;
 
       editor.setTheme("ace/theme/tomorrow");
-      editor.setKeyboardHandler("ace/keyboard/vim");
+      editor.setKeyboardHandler("ace/keyboard/emacs");
```
dometto: What is the reason for this change? I'm not against emacs instead of vim bindings, but if not necessary I would prefer to keep the old settings for continuity.

ProgramFan: I change the keyboard to emacs to make it more comfortable for non-vim users, since it does not involve something like 'insert/command/visual mode'. I have mistakenly hard-coded my preference (vim) in the code and the commit is to revert it.

ProgramFan:
I have also tried the autocompletion. It works greatly. But It might be too noisy for some users if I turn on live completion, and it would be too tedious for the `ctrl+space` trigger if I turn the live completion off. So I decide not to include it at the moment.

One solution would be to customize a completer so that it is only triggered after the user entered 3 initial chars. Even so, some users may prefer no completer at all for distraction-free writing. So this feature may not be appropriate for including in the default. However, enabling it would be easy, just follow the instructions given above.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
bartkamphorst: 👍

bartkamphorst: @ProgramFan Could https://github.com/gollum/gollum/pull/1215 be relevant here, so that people can switch between key bindings?

ProgramFan: Done in latest commit. @bartkamphorst

dometto:
@ProgramFan, PR is looking great! I want to play around with it a bit more before I merge, but I don't think there are any major issues. Any chance you could implement https://github.com/gollum/gollum/issues/705 while you're at it? :)

dometto:
Also, https://github.com/gollum/gollum/pull/1173 still seems to be an issue. I don't want to make you responsible for all this bugs, so please just let me know when you're done working on this. But at the moment you're the person with the most knowledge about the editor, so if these things are relatively easy for you to fix it would be great!

@ProgramFan  :1st_place_medal:

dometto:
Apologies, actually #436 seems to be resolved in this PR!

ProgramFan:
@demetto I am glad to fix #705, but the fix goes to gollum-lib instead of gollum. See https://github.com/gollum/gollum-lib/pull/251.

dometto:
One final issue I just realized is line numbering, which now seems to be turned on by default. If I'm not mistaken ACE also has an option for turning it off, which would stay more true to gollum's current editor. I think the default should be no line numbering, although it would be great if it could be turned on easily (for instance with a keyboard shortcut?) if that's possible. Or do others disagree?

ProgramFan:
Done. To turn it on, use `cmd+,` on macOS and `ctrl+,` on linux and windows to activate Ace's settings panel, where everything can be tuned.

dometto:
Great! Then all that remains is squashing the commits and I'll merge!

ProgramFan:
I just realize that it is best for you to merge using `--squash` instead of me squashing, so I don't have to mess up the commits of my fork. Is is OK for you? @dometto

dometto:
Sure, I'll try to manually merge/squash this PR.

dometto:
Ah, just noticed the 'Squash and merge' option on github. 👍

dometto:
🎉 🎆 

Many thanks @ProgramFan, I'll update `HISTORY.md` to list your contributions! 👍

Issue #1031, opened 2015-07-02
Feature request: Support for local files instead of commits


maxless:
If you're working in an external text editor, you have to commit to see any change even if you're trying to do something small. Not to mention you have to constantly squash commits or have a lot of them around. It would be great if you could use Gollum to preview all changes before committing.

SkyCrawl:
Great idea! I had this issue with external editors myself...

@dometto @bartkamphorst 
How about a couple of URL parameters? E.g.:
*  `/wiki/my-page?cwd` (display page from the working directory)
*  `/wiki/my-page?index`(display page from the git index)
*  `/wiki/my-page` (display page from the latest commit)
*  `/wiki/my-page?commit=...` (display page from the given commit)

dometto:
IMHO this would be too much trouble to implement, and goes a bit against gollum's philosophy. Gollum offers the livepreview option to edit files and preview the changes. I think our efforts should go towards making the livepreview and editor better. See #826.

Of course I see how a markdown preview app would come in handy when working with external editors, but that's not limited to gollum, and I don't think gollum should solve it. Aren't there any other apps (browser-based or otherwise) that take care of this?

@sunny, what's your take on this?

SkyCrawl:
> IMHO this would be too much trouble to implement

I think the only problem is git index :). All the other options are more or less ready/implemented. Integration itself is pretty easy. All of the above could even be squashed into a single parameter, despite personally, I don't think it's a good idea.

> Gollum offers the livepreview option to edit files and preview the changes. I think our efforts should go towards making the livepreview and editor better.

Same here. However, I don't think we should rule out convenient and trivial support for external editors so easily. I had the impression that Gollum didn't impose its live-preview editor on people? Actually, external editors are often much better, honed and more actively developed :). Also, users may be using Gollum only for a fraction of their (e.g.) markdown documents. What's wrong about using one's favourite editor for all (e.g.) markdown documents, if the user desires so? **On top of that, implementing `/wiki/my-page?cwd` (which is what @maxless requests) is, to my understanding, fairly easy :).**

> Of course I see how a markdown preview app would come in handy when working with external editors, but that's not limited to gollum, and I don't think gollum should solve it. Aren't there any other apps (browser-based or otherwise) that take care of this?

[...]

PR #1174, opened 2016-12-08
feature #1031

vitalije:
I have added a new URL /tip/* which will serve pages and files directly from a working folder. That enables preview of not yet committed changes as requested in issue #1031.
Vitalije

PS: ruby is not my language of choice, therefore I didn't install any of dependencies and didn't run tests but the code is working for me and it solves #1031.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -511,6 +511,34 @@ module Precious
       mustache :file_view, { :layout => false }
     end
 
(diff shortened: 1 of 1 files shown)


dometto:
Thanks for the PR! I like the idea, will review it more carefully shortly. However, would you mind reopening this PR against the `5.x` branch? I promise a more swift reply than last time. :)

PR #1178, opened 2016-12-31
Remove logic to deal with "h1_title".

bogado:
This logic was moved into gollum-lib where the Page object now knows how
to create it's title.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -107,7 +107,7 @@ module Precious
     end
 
     get '/' do
-      redirect clean_url(::File.join(@base_url, @page_dir, wiki_new.index_page))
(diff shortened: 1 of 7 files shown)


dometto:
Tests seem to pass except for the latest commit (b676d92). I think that change (don't initialize a new wiki object for every new page) is rather a tricky one, though -- I'm not sure the application is thread-safe without it. Perhaps just leave that commit out?

bogado:
That is true a deeper analysis would be required to check this. 

I'm not used to the ruby infrastructure, is the default server framework threaded? My setup is a ruby background service that can be single threaded, new hosts or processes could be spawned if needed (hardly with my audience of 1 ;-P). An apache that have a cache and proxy the ruby part is threaded and distributed.

bartkamphorst:
Closing due to lack of activity / response.

PR #1180, opened 2017-01-02
Fix page list for files that have regexp special chars.

bogado:
The page list collection logic was using the filename without any
scaping to create a regexp. This not only breaks for some names it might
even be a security problem by introducing bad regular expression as
filenames.


dometto:
:+1: Thanks!

PR #1182, opened 2017-01-10
Add another video to README.

bjoernalbers:
I've made a video to help users without markdown experience get started with gollum.


polyzen:
Nice!

Gollum supports [quite a few](https://github.com/gollum/gollum#markups) other markup languages than Markdown. Perhaps this could be mentioned in the YT description?

My only nitpick is that you use a 2nd level header (##) for the top header in the demo, but another demo even uses a 3rd level header there. :p

bjoernalbers:
I've now mentioned the other markups on YouTube (descriptions).

About the headers: I'd like to leave the page name / title as my only h1 so that my content has to start at h2.

Thanks in advance for merging :-)

dometto:
Thanks 👍

PR #1185, opened 2017-01-15
Test on Ruby 2.4

connorshea:


bartkamphorst:
I think this can be merged without issue. @dometto ?

Issue #1187, opened 2017-01-27
Test suite failures with rugged_adapter 0.4.3


adamniedzielski:
```
cat Gemfile.lock | grep rugged
```
```
    gollum-rugged_adapter (0.4.3)
      rugged (~> 0.24.0, >= 0.21.3)
    rugged (0.24.6.1)
  gollum-rugged_adapter (= 0.4.3)
```

```
bundle exec rake GIT_ADAPTER=rugged
```

```
/Users/adam/.rbenv/versions/2.3.3/bin/ruby -I"lib:lib:test:." -I"/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rake-10.5.0/lib" "/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rake-10.5.0/lib/rake/rake_test_loader.rb" "test/**/test_*.rb" 
Loaded suite /Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rake-10.5.0/lib/rake/rake_test_loader
Started
....................................F
============================================================================================================================================================
Failure: test_urls_transform_unicode(Frontend)
/Users/adam/workspace/gollum/test/test_app.rb:38:in `block (2 levels) in <top (required)>'
<"m-plus-F"> expected but was
<"μ†ℱ">
============================================================================================================================================================
....E
============================================================================================================================================================
Error: test_transliteration(Frontend_Unicode_support): Rugged::InvalidError: Failed to parse signature - Signature cannot have an empty name or email
/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/gollum-rugged_adapter-0.4.3/lib/rugged_adapter/git_layer_rugged.rb:461:in `create'
/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/gollum-rugged_adapter-0.4.3/lib/rugged_adapter/git_layer_rugged.rb:461:in `commit'
/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/gollum-lib-4.2.2/lib/gollum-lib/committer.rb:170:in `commit'
/Users/adam/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/gollum-lib-4.2.2/lib/gollum-lib/wiki.rb:357:in `write_page'
[...]

PR #1188, opened 2017-01-27
Pass non-empty commit author details in transliteration test

adamniedzielski:
Empty name or email are not allowed by libgit2 and cause a test failure when the test suite is run against rugged_adapter.

Please refer to https://github.com/gollum/gollum/issues/1187#issuecomment-275672962 for a more detailed explanation.

### DIFF
--- a/test/test_unicode.rb
+++ b/test/test_unicode.rb
@@ -91,7 +91,7 @@ context "Frontend Unicode support" do
 
   test 'transliteration' do
     # TODO: Remove to_url once write_page changes are merged.
-    @wiki.write_page('ééééé'.to_url, :markdown, '한글 text', { :name => '', :email => '' })
+    @wiki.write_page('ééééé'.to_url, :markdown, '한글 text', commit_details)
     page = @wiki.page('eeeee')
     assert_equal '한글 text', utf8(page.raw_data)
   end


dometto:
I'll merge this, but it would be great to already include a fix for the next test issue you mention in https://github.com/gollum/gollum/issues/1187#issuecomment-279655415 -- the encoding issue. What about just including an ugly test which determines whether to use a transliterated url or not? Something along the lines of:
```ruby
defined?(Gollum::GIT_ADAPTER) && Gollum::GIT_ADAPTER != 'grit' 
```
We could remove this once we switch to rugged as a default adapter.

cc @bartkamphorst

adamniedzielski:
@dometto I'd say that these two things are independent so you can merge this PR and it will still be an improvement over the current situation. I can submit a next PR for the encoding issue.

What do you think about just skipping the tests related to transliteration if the adapter is different than `grit`? We only perform transliteration for `grit` so we could only test it for `grit`.

dometto:
> What do you think about just skipping the tests related to transliteration if the adapter is different than grit? We only perform transliteration for grit so we could only test it for grit.

Yes, that was what I was trying to suggest. :) 👍

PR #1189, opened 2017-01-28
Solve bug when folder contains non-ascii character

nimag42:
When you create a file in a folder with non-ascii character, for exemple "Réseau", after creating the page, it throwed an "URI::InvalidURIError", given the fact that the url returned was "/Réseau/H%C3%A9y", only the part with the name of the file was correctly encoded.

So I propose to encode every part of the url to solve this issue.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -348,7 +348,7 @@ module Precious
         wiki.write_page(name, format, params[:content], commit_message, path)
 
         page_dir = settings.wiki_options[:page_file_dir].to_s
-        redirect to("/#{clean_url(::File.join(page_dir, path, encodeURIComponent(name)))}")
+        redirect to("/#{clean_url(::File.join(encodeURIComponent(page_dir), encodeURIComponent(path), encodeURIComponent(name)))}")
       rescue Gollum::DuplicatePageError => e
         @message = "Duplicate page: #{e.message}"
         mustache :error


lib/gollum/app.rb
```diff
@@ -348,7 +348,7 @@ def wiki_new
         wiki.write_page(name, format, params[:content], commit_message, path)
 
         page_dir = settings.wiki_options[:page_file_dir].to_s
-        redirect to("/#{clean_url(::File.join(page_dir, path, encodeURIComponent(name)))}")
+        redirect to("/#{clean_url(::File.join(encodeURIComponent(page_dir), encodeURIComponent(path), encodeURIComponent(name)))}")
```
bartkamphorst: Untested, but wouldn't it be more efficient to call `encodeURIComponent` once with the joined URL string?

nimag42: Oh yeah, I haven't tested, but if it works, of course It would be better :)

dometto: Any news on that?

bartkamphorst: Not from me.

bartkamphorst: @nimag42 Care to update your PR?

dometto:
Ping @nimag42: do you want to incorporate @bartkamphorst's suggestion?

lib/gollum/app.rb
dometto: Actually, that won't work:
```ruby
2.4.1 :005 > encodeURIComponent("waa/dada")
 => "waa%2Fdada"
```

bartkamphorst: OK.

nimag42: Sorry guys, I didn't answered because I wasn't available at all :(
Glad you finally solved and merged it :)

Thanks for your time

PR #1198, opened 2017-02-24
Template page

brandubh:
sorry for the delay but here is the PR for 5.x branch


dometto:
Restarted the tests and they're passing now! Many thanks @brandubh!

PR #1200, opened 2017-03-09
Use sinatra 2.0. Closes #1193.

### LINKED ISSUE #1193: Rails 5 compatibility
I can't upgrade do Rails 5, getting the following after `bundle update`:

```bash
Resolving dependencies....
Bundler could not find compatible versions for gem "rack":
  In Gemfile:
    rails (>= 5.0.0.1, ~> 5.0.0) was resolved to 5.0.0.1, which depends on
      actionpack (= 5.0.0.1) was resolved to 5.0.0.1, which depends on
        rack (~> 2.0)

    gollum (~> 4.0) was resolved to 4.0.0, which depends on
      sinatra (>= 1.4.4, ~> 1.4) was resolved to 1.4.4, which depends on
        rack (~> 1.4)
```

Is there any work to make it compatible with `rack (~> 2.0)`?

dometto:
PR to test whether there are any troubles upgrading to sinatra `v2.x`.

### DIFF
--- a/gollum.gemspec
+++ b/gollum.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'gollum-lib', '~> 4.0', '>= 4.0.1'
   s.add_dependency 'kramdown', '~> 1.9.0'
-  s.add_dependency 'sinatra', '~> 1.4', '>= 1.4.4'
+  s.add_dependency 'sinatra', '~> 2.0rc1'
   s.add_dependency 'mustache', ['>= 0.99.5', '< 1.0.0']
   s.add_dependency 'useragent', '~> 0.16.2'
   s.add_dependency 'gemojione', '~> 2'


dometto:
Test failures are due to https://github.com/sinatra/mustermann/issues/60. We either wait for this issue to be resolved, or we switch to single-line regexps (which do work in `sinatra v2`).

HarrisonB:
So will this branch work if we are not utilizing Regxp::EXTENDED? Or is that unavoidable? Simply trying to plug gollum into a new Rails 5 site and I'm not sure if I should use this branch or wait for a proper fix.

Thanks for all of your work! 😃

dometto:
I think it *should* work but haven't tried yet. I won't merge this branch eventually since sinatra `v2` is currently only at `rc2`, and we'll wait for the eventual release to use in gollum. If you need something quick and dirty I would recommend using a personal fork for the time being. If you go with it I'd be very interested to hear whether you encounter any issues after removing the `Regexp::EXTENDED`s!

HarrisonB:
Gotcha. So does this mean I would need to fork `gollum`, `sinatra`, and `mustermann`, make the appropriate edits, and then reference those in my Rails 5 project? Sorry for all the questions...I'm new to ruby and not familiar with the way it handles dependency resolution. I ask because it seems odd to me that two gems can't have requirements for (significantly) different versions of the same gem.

dometto:
Nope, you only need to update the `sinatra` dependency as on this branch. Basically you can fork this and proceed by changing the multiline regexps in `app.rb` to single-line ones. Run `bundle exec gollum` to start the app. Don't forget to run `bundle install` when you change a project's dependencies. :)

PR #1201, opened 2017-03-11
Skip tests for transliteration for adapters different than grit

adamniedzielski:
Closes https://github.com/gollum/gollum/issues/1187

/cc @dometto


dometto:
:+1: Thanks!

PR #1215, opened 2017-04-02
Add selection of keybindings in preview page

mkdong:
Add keybinding choices in the preview page, including "vim", "emacs", and the default one.


dometto:
@bartkamphorst I understand that a similar feature will be merged into `5.x` in #1173. 👍 Do we want to merge this into `master` or close it? Either way, thanks for the PR @mkdong!

bartkamphorst:
Does the other PR also concern the preview pane?

dometto:
No, but I would propose we remove support for livepreview mode completely in `5.x` (we might expand the default editor with livepreview support as in a proposal by @rtrvrtg, if I'm not mistaken).

bartkamphorst:
My bad, I thought this concerned the regular preview option. I agree that livepreview in its current state (only for markdown) should be removed in version 5.x.

PR #1217, opened 2017-04-08
Yaml frontmatter

dometto:
* Implement command line switch `--no-display-metadata` to disable metadata rendering globally (enabled by default)
* Implement `rendered_metadata` method in the `Page` view
* Updated `page` layout and CSS to render metadata tables
* Added a test

This will fail until we release `gollum-lib 5.0.a1` and make `gollum` depend on that. However, the tests already pass locally.

Metadata tables render prettily for simple `{:key => :value}` metadata...
screen shot 2017-04-08 at 16 45 55
...but also for more complicated metadata with iterated hashes and arrays:
screen shot 2017-04-08 at 16 45 30
The latter overflows the table and makes it scrollable (so too large tables will not run off the edge of the page).

### DIFF
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -143,7 +143,7 @@ module Precious
       #
       # Returns HTML table.
       def rendered_metadata
-        return '' if !page.display_metadata? || metadata.empty?
+        return '' unless page.display_metadata? && !metadata.empty?
         @rendered_metadata ||= table(metadata)
       end
 
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -196,7 +196,7 @@ module Precious
       def table(data)
         return data.to_s if data.empty?
         result = "<table>\n"
-        keys = data.respond_to?(:keys) ? data.keys : nil
+        keys = data.respond_to?(:keys) && data.respond_to?(:values) ? data.keys : nil
           if keys
             data = data.values
             result << "<tr>\n"
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -130,13 +130,21 @@ module Precious
 
       # Access to embedded metadata.
       #
-      # Examples
-      #
-      #   {{#metadata}}{{name}}{{/metadata}}
-      #
       # Returns Hash.
       def metadata
         @page.metadata
+      end   
+
+      # Access to embedded metadata.
+      #
+      # Examples
+      #
+      #   {{#rendered_metadata}}{{name}}{{/rendered_metadata}}
+      #
+      # Returns HTML table.
+      def rendered_metadata
+        return '' if !page.display_metadata? || metadata.empty?
+        @rendered_metadata ||= table(metadata)
       end
 
       private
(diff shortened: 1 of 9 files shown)


lib/gollum/views/page.rb
```diff
@@ -130,13 +130,21 @@ def use_identicon
...
+
+      # Access to embedded metadata.
+      #
+      # Examples
+      #
+      #   {{#rendered_metadata}}{{name}}{{/rendered_metadata}}
+      #
+      # Returns HTML table.
+      def rendered_metadata
+        return '' if !page.display_metadata? || metadata.empty?
```
bartkamphorst: I would use unless here to better express the intent.
```ruby
return '' unless page.display_metadata? && ! metadata.empty?
```

lib/gollum/views/page.rb
```diff
@@ -184,6 +192,42 @@ def content_without_page_header(content)
         # .inner_html will cause href escaping on UTF-8
         doc.css("div#gollum-root").children.to_xml(@@to_xml)
       end
+
+      def table(data)
+        result = "<table>\n"
+        case data
+          when Hash
```
bartkamphorst: Not a big fan of explicit class checking in Ruby. How desirable is it to allow `Array`s to work?

bartkamphorst:
Very cool stuff. Only thing is that I would expect the first table to have two columns and three rows (1 row for each key-value pair).

dometto:
Thanks! I adopted the horizontal rather than vertical table approach from github. I don't feel very strongly about either, except that I don't know how a vertical table as complex as the second one would look.

lib/gollum/views/page.rb
dometto: The alternative is to just print `Array.to_s`. That's alright in principle, but then I think we also shouldn't iterative over `Hash`es but just print `to_s` for every value, so basically only have a "first order" table. My preference would be to do what github does, but I don't actually know how they handle `Array`s and `Hash`es in front matter. :)

dometto: They do handle `Array`s judging from the screenshot here: https://github.com/blog/1647-viewing-yaml-metadata-in-your-documents

Can we come up with a way to do this without explicitly checking for class?

dometto:
I tried the vertical design with the embedded tables:

screen shot 2017-04-09 at 14 44 27

I must say I prefer the horizontal design to this. See the refactor I've just pushed for a version of that without class detection.

Vertical `table` method for future reference:
```ruby
      def table(data)
        return data.to_s if data.empty?
        result = "<table>\n"
        data.each do |key, value|
          result << "<tr\n>"
          if value
            result << "<th>#{CGI.escapeHTML(key)}</th>\n"
            result << "<td>" << (value.respond_to?(:each) ? table(value) : CGI.escapeHTML(value.to_s)) << "</td>\n"
          else
            result << "<td>" << (key.respond_to?(:each) ? table(key) : CGI.escapeHTML(key.to_s)) << "</td>"
          end
          result << "<tr/>\n"
        end
        result << "</table>\n"
      end
```

lib/gollum/views/page.rb
```diff
@@ -194,38 +194,22 @@ def content_without_page_header(content)
...
-        values.each do |value|
-          result << "<td>"
-          case value
-            when Hash, Array
-              result << table(value)
-            else
-              result << CGI.escapeHTML(value.to_s)
+        keys = data.respond_to?(:keys) ? data.keys : nil
+          if keys
+            data = data.values
```
dometto: @bartkamphorst I've refactored the table construction into a single method, avoiding class detection.

lib/gollum/views/page.rb
dometto: 👍

lib/gollum/views/page.rb
dometto: Perhaps `keys = data.respond_to?(:keys) && data.respond_to?(:values) ? data.keys : nil`?

bartkamphorst: > I've refactored the table construction into a single method, avoiding class detection.

Nicely done. 👍 

> `keys = data.respond_to?(:keys) && data.respond_to?(:values) ? data.keys : nil`

OK.

bartkamphorst:
> I adopted the horizontal rather than vertical table approach from github. 

OK, then that makes sense. I somehow seemed to recall metadata just being a simple hash and therefore being rendered as a two-column table, but I'm okay with more complexity if that better mimics github's behavior.

PR #1218, opened 2017-04-08
Bundle a slim mathjax

ProgramFan:
Without bundled mathjax, math contents are not properly displayed in the absence of an internet connection. Previous concerns with bundling mathjax in gollum may the huge increase in the package size (+25MB approx.). With the help of https://github.com/mathjax/MathJax-grunt-cleaner, mathjax can be tailored to being more slim while retaining almost all functionalities. This PR does just that.


ProgramFan:
@demetto @bartkamphorst I guess it would be nice to include this in the 5.x release so Gollum's offline experience is much better.

dometto:
Good idea!

For `5.x` it would be great if we could come up with an easier way of bundling JS dependencies, perhaps through using git submodules or https://github.com/yarnpkg/yarn. If you have any ideas on that they are very welcome!

ProgramFan:
@dometto For JS like Ace editor and MathJax, I would prefer the current approach, where we bundle explicitly a customized version as part of Gollum's source. This approach has the advantage that we can easily customize the lib to fit our needs. For example, the slimified mathjax is much smaller, which is not possible if we link to upstream repos or releases. It does not make the update harder, either. We can update it whenever we want. The only problem would be dependency management if we use complex enough JS libs, but I don't see Gollum using this type of JS libs (but of course it might have already happened or it would happen in the futhre).

dometto:
Ok, thanks for the feedback! @bartkamphorst, how do you feel about this one?

bartkamphorst:
I'm a little on the fence about JS management, as keeping it all in the project also means that the gollum team is in charge of updating the JS code base in line with security patches etc. In an ideal world, upgrading some javascript lib would just involve updating a version number in a dependency list. The problem is that I don't know how realistic that is...

ProgramFan:
Maybe we can address the problem by split the JS libs out of Gollum into a seperate repo, and use git submodules to link to it. That repo contains only external JS libs used in Gollum, together with scripts or documentation on how to produce customized version from a given upstream version. The repo can still be maintained by Gollum team but it can be managed with yarn or npm to ease the maintaince.

I guess the major problem is caused by  that we want to customize some JS libs.

dometto:
Thanks for the input @ProgramFan. I suspect something like that may be the best we can do. Perhaps we should start listing our actual JS dependencies and the required customizations.

bartkamphorst:
> Perhaps we should start listing our actual JS dependencies and the required customizations.

I agree that that would be a helpful first step. 👍

dometto:
I actually recently noticed how annoying it is that MathJax isn't bundled: when offline, pages will contain unrendered MathJax like `$p \rightarrow q$`. I vote in favor of bundling it, especially since it's not a dependency that's very update-intensive. @bartkamphorst, what do you think?

bartkamphorst:
@dometto I like the idea of being able to run gollum offline, so I'm on board. @ProgramFan Would you be willing to double-check that we would be bundling the latest version of mathjax?

dometto:
Looking at the source, there seem to be these two differences at least:
* Mathjax version currently used in the 5.x branch is 2.7.1, in the PR is 2.7
* Currently we are using these options `?config=TeX-AMS-MML_HTMLorMML'`, where the PR uses `?config=TeX-AMS_HTML-full'` (the difference seems to be that we are currently using MML, too?)

ProgramFan:
I will try to upgrade and fix the differences.

ProgramFan:
Shall we stick to `TeX-AMS-MML_HTMLorMML` for the config? We can get a much smaller mathjax if we use only one config.

ProgramFan:
Upgrade done in the latest commit. A document on how to slimify mathjax is added.

ProgramFan:
Upgrade done and a document on how to upgrade mathjax is added.

bartkamphorst:
> Shall we stick to TeX-AMS-MML_HTMLorMML for the config? We can get a much smaller mathjax if we use only one config.

Seems sensible to me. 👍

ProgramFan:
The latest commit shall have done that. svg output is cleared and only two config are kept.

bartkamphorst:
Great, thanks @ProgramFan !

PR #1220, opened 2017-04-09
Remove ws subs

dometto:
Refactor front end to deal with lack of whitespace substitution and the end of file extension suppression in `gollum-lib`.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -18,20 +18,16 @@ require File.expand_path '../helpers', __FILE__
 Gollum::set_git_timeout(120)
 Gollum::set_git_max_filesize(190 * 10**6)
 
-# Fix to_url
+# Use stringex #to_url only to leverage its #to_ascii method when using grit
 class String
-  alias :upstream_to_url :to_url
-
   if defined?(Gollum::GIT_ADAPTER) && Gollum::GIT_ADAPTER != 'grit'
     def to_ascii
       self # Do not transliterate utf-8 url's unless using Grit
     end
   end
 
-  # _Header => header which causes errors
   def to_url
-    return nil if self.nil?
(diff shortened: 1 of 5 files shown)


lib/gollum/helpers.rb
```diff
@@ -23,7 +27,8 @@ def extract_name(file_path)
 
       # File.basename is too eager to please and will return the last
       # component of the path even if it ends with a directory separator.
-      ::File.basename(file_path)
+      ext = ::File.extname(file_path)
+      return ::File.basename(file_path, ext), ext
```
dometto: I had to first chop the file extension of (via `extract_name`) and then join it back to the name (via `join_page_name`) because currently, if we pass e.g. `name.md` to `gollum-lib`, the page in the repo ends up being called `name.md.md`. This makes the code in `app.rb` even uglier, but I propose we first merge it like this before we can do a more extensive rewrite (that will probably have to include the method signatures of `write_page` and such in `gollum-lib`).

lib/gollum/app.rb
```diff
@@ -18,20 +18,16 @@
...
   if defined?(Gollum::GIT_ADAPTER) && Gollum::GIT_ADAPTER != 'grit'
     def to_ascii
       self # Do not transliterate utf-8 url's unless using Grit
     end
   end
 
-  # _Header => header which causes errors
   def to_url
-    return nil if self.nil?
-    upstream_to_url :exclude => ['_Header', '_Footer', '_Sidebar'], :force_downcase => false
```
dometto: No longer rewrite e.g. `waa.md` to `waa-dot-md`.

lib/gollum/helpers.rb
bartkamphorst: OK.

bartkamphorst:
Exciting times! 😄

PR #1234, opened 2017-04-23
Found more spurious whitespace substitution.

dometto:
The views were still substituting '-' with space, causing problems when trying to edit pages containing '-'.

I now searched the repo for instances of the string `'-'` and `"-"`, which should return any cases of attempted ws substitution in ruby or javascript (unless I'm too unimaginative). The only other hit was the instance in `app.rb`, which I also took care of in this PR.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -305,8 +305,8 @@ module Precious
       if settings.wiki_options[:template_page] then
         temppage = wiki_page("/_Template")
         @template_page = (temppage.page != nil) ? temppage.page.raw_data : "Template page option is set, but no /_Template page is present or committed."
-      end          
-      wikip = wiki_page(params[:splat].first.gsub('+', '-'))
+      end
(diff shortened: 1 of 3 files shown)


lib/gollum/app.rb
```diff
@@ -305,8 +305,8 @@ def wiki_new
       if settings.wiki_options[:template_page] then
         temppage = wiki_page("/_Template")
         @template_page = (temppage.page != nil) ? temppage.page.raw_data : "Template page option is set, but no /_Template page is present or committed."
-      end          
-      wikip = wiki_page(params[:splat].first.gsub('+', '-'))
```
dometto: This `gsub` was added because apparently the `params` would contain strings like 'New+Page', with `+` characters instead of spaces. Tests show that `params` are now already properly sanitized by sinatra, so we get `New Page` and don't want to change that.

bartkamphorst:
Nice. 👍

PR #1247, opened 2017-05-31
CSS to SCSS

dometto:
Ping @bartkamphorst @thomaskuntzz

This is a fresh version of #1097, utilizing the SASS rack plugin. Everything seems to work, but I'm not very experienced with SASS/SCSS and have some questions:

* I don't know whether it's necessary/desirable to have the ability to precompile in production mode, nor how this is normally handled. I don't see a way to do this using the SASS rack plugin (or would this effectively be a matter of setting `:never_update` to `true` -- see [here](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#options)?).
* I had to modify the `.scss` files a bit to deal with new/modified CSS in recent commits to the `5.x` branch. I don't know if I handled that in the correct way. I'll indicate where in the line comments.


lib/gollum/public/gollum/stylesheets/sass/editor.scss
```diff
@@ -0,0 +1,504 @@
...
+  position: absolute;
+  top: -5000px;
+  width: 1px;
+}
+
+.disabled {
+  color: lightgray;
+}
+
+#gollum-editor-body-ace {
```
dometto: I just pasted the CSS for the ace editor here. If there's a better way of incorporating this in the new SCSS I'd be happy to hear about it.

lib/gollum/public/gollum/stylesheets/sass/gollum.scss
```diff
@@ -0,0 +1,712 @@
...
+    margin: 1em 0;
+    padding: 0.25em;
+  }
+}
+
+.clearfloats {
+  clear: both;
+}
+
+.emoji {
```
dometto: Same here

dometto:
Also, currently this is generating `.css` and `.css.map` files in the `lib/gollum/public/gollum/stylesheets` directory. This strikes me as slightly strange behavior for those using gollum as a gem -- it would basically mean modifying the gem directory after installation. Hence my question about preprocessing: should the gem version contain the compiled `css` files? And if so, should it still use the `rack-sass` middleware? 😕

lib/gollum/public/gollum/stylesheets/sass/editor.scss
thomaskuntzz: In SASS, you should not use semicolons nor curly braces :slightly_smiling_face:

thomaskuntzz:
Hum... Do we want to use SCSS or SASS? :question:

dometto:
Honestly, I don't know! SCSS is easiest because someone else already did the work. Are there big advantages to using SASS instead?

thomaskuntzz:
Well, it's a question of preference. As I only knew SASS, I used to think it's better, but after asking which one we are going to use, my interest was piqued, and I found [this](http://thesassway.com/editorial/sass-vs-scss-which-syntax-is-better) which made me think SCSS is better here (especially for [reason #6](http://thesassway.com/editorial/sass-vs-scss-which-syntax-is-better#reason-6-integration-with-an-existing-css-codebase-is-much-easier) and [reason #7](http://thesassway.com/editorial/sass-vs-scss-which-syntax-is-better#reason-7-scss-provides-a-much-lower-barrier-to-entry)).

If all the work is done is done with SCSS, that's one more reason I think we should keep it!

dometto:
> Also, currently this is generating .css and .css.map files in the lib/gollum/public/gollum/stylesheets directory. This strikes me as slightly strange behavior for those using gollum as a gem -- it would basically mean modifying the gem directory after installation. Hence my question about preprocessing: should the gem version contain the compiled css files? And if so, should it still use the rack-sass middleware? 😕

Still puzzled about this. Anyone with a better understanding of SCSS able to help me out? cc @ProgramFan @sunny

lopopolo:
What do you consider to be the gollum document root?

ProgramFan:
Sorry, I am new to the SCSS world and don't really understand SCSS or SASS. Most of the time, I use plain CSS.

thomaskuntzz:
Plain CSS is valid SCSS, so that fine, and SCSS in the that complicated, it only adds a few nice features on too of CSS 😉

dometto:
@lopopolo thanks for chipping in. Yes, this didn't seem like a good idea. But I'm not sure what the proper way to serve assets is then. I was expecting the rack `rack-sass` middleware to just inject the compiled assets into the request chain, but instead it outputs the files to disk. Can you illuminate?

lopopolo:
My proposal with production mode was to allow a `gollum init` (or similar) phase before launching e.g. thin that pre-compiles the assets to a user-specifiable location so they can be served directly by something like nginx.

dometto:
That would be great to have, but it's still essential to have a working rack implementation without nginx or Apache, for one thing so that gollum can be continued to use as part of other applications.

lopopolo:
Have that pre-compile phase and use [Rack::Static](http://www.rubydoc.info/github/rack/rack/master/Rack/Static)?

dometto:
The present implementation uses `sprockets` to dynamically serve assets, compiling everything on the fly by default. We should add an option to precompile assets along these lines: https://gist.github.com/jeffreyiacono/1772989

Also, the file view CSS is not working correctly. For merging into `5.x` I think this is not a showstopper, since we want to redo the file view anyway.

dometto:
Another comment: at present some assets (`ace`, `jquery`) are already uglified in the source tree. Unfortunately, `sprockets` will run them through uglify them again. Seems like [a fix](https://github.com/rails/sprockets/issues/298) for this will have to wait until `sprockets v4`.

dometto:
@lopopolo you might want to check out the current implementations, which comes with a `--static` option to run gollum with precompiled assets, and a `--precompile` switch to generate the assets.

At the moment, almost everything is working, with the exception of some of the JS in the editor view.

lib/gollum/app.rb
```diff
@@ -100,15 +91,35 @@ class App < Sinatra::Base
...
     get '/' do
       redirect clean_url(::File.join(@base_url, @page_dir, wiki_new.index_page))
     end
 
     get '/assets/*' do
-      env['PATH_INFO'].sub!('/assets', '')
-      settings.sprockets.call(env)
+      env['PATH_INFO'].sub!("/#{Precious::Assets::ASSET_URL}", '')
+      if @use_static_assets
+        env['PATH_INFO'].sub!(Sprockets::Helpers.prefix, '') if @base_url
```
dometto: This is a bit of an ugly hack, but I couldn't get it to work any other way. The problem is that when using static assets and `--base-path`, asset paths in the compiled CSS have a double prefix: `/tests/assets/test/assets`, for example. This is because we need to set the prefix to `test/assets` at compile-time, and at run time. Disabling either leads to paths that are malformed in different ways.

As far as I can see, the assets route now works both when using static and dynamic assets, and both with and without a base path.

lib/gollum/views/helpers.rb
```diff
@@ -0,0 +1,35 @@
+require 'yaml'
+
+module Precious
+  module Views
+    module SprocketsHelpers
```
dometto: This is necessary to be able to use the Sprockets Helpers from within a mustache template. The only way you can pass an argument to a View-method from mustache is for the method to return a `lambda`. So we need to wrap the Helper methods into a `Proc` in order to be able to do this inside a template:
```
{{#stylesheet_tag_mustache}}app{{/stylesheet_tag_mustache}}
```
This yields the same result as `<%= stylesheet_tag 'app' %>` inside an `erb` template. You can also pass in an optional options hash using YAML:
```
{{#stylesheet_tag_mustache}}app media: :all{{/stylesheet_tag_mustache}}
```
This yields the same result as `<%= stylesheet_tag 'app', :media => :all %>` inside an `erb` template.

dometto: To clarify further, using the Sprockets helpers is necessary because Sprockets generates assets with an md5-fingerprint. So the correct parth to the application stylesheet will e.g. be `assets/app-580166b40601dad84878e5922b249013547baf5078617b192d9a4eb8d4e32fab.css`. The `stylesheet_tag` helper and its family is necessary to generate a link to the asset that uses the correct fingerprint.

dometto:
Also to clarify further: `--static` is the default option in production and staging, but can be overridden with `--no-static`.

dometto:
Summary:
* everything is now successfully ported to SCSS
* exception: the fileview layout. I've modified that to conform to the `All Pages` view's layout, however, now the mini-icons aren't working (trashcan and expand icon).
  * I would propose to merge this anyway, since we are going to extract the fileview CSS from `gollum-lib` anyway. Once the CSS is in `gollum` it should be much easier to fix it up, for example by making it use fontawesome icons instead of our custom `.png`s.
* `gollum` now has `--precompile` switch. The default asset location is `public/assets`. The `--static` switch will use static assets.
* Static assets are default (no switch required) in staging and production (but this can be overridden with `--no-static`).
* However, the default mode for development and testing is **dynamic** assets, which `sprocket` compiles on the fly the first time the assets are requested. This compilation can take quite a bit of time, in practice making the first request to a new `gollum` instance take about 30 seconds.
  * Furthermore, `gollum` is by default run in development mode! So users who just do a `gem install gollum; gollum /my/wiki` will be confronted with the longish load time.
  * One option is to change `gollum`'s default mode to `production`, thus enabling static assets by default. But then we would have to package the *precompiled* assets in `public/assets` in the gem -- otherwise `gollum` wouldn't work out of the box. I'm not sure what is considered best practice when deploying ruby apps with SASS.

@bartkamphorst when do you think this PR would be ready to merge? Should I somehow include tests for `--precompile`, `--static` and/or `--no-static`?

PR #1252, opened 2017-07-05
Changed gollum script to allow unescaped forward slashes in base_path.

Agilus:
The code previously would escape / characters (replacing them with "%2F"), which caused working with Omnigollum/Omniauth not to work correctly (since path comparisons no longer showed a matching path).  This minor change makes Gollum work nicely with Omnigollum even when setting a base_path that is multiple levels deep.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -76,7 +76,11 @@ opts = OptionParser.new do |opts|
     # make a backup of the option and sanitize it
     base_path_original = base_path.dup
     base_path = CGI.escape(base_path)
-    
+
+    # / characters are perfectly acceptible in paths, and should not be replaced.
+    # This line backs out that change.
+    base_path.gsub!("%2F", "/")
+
     # then let the user know if we changed the URL
     unless base_path_original == base_path
       puts <<MSG
--- a/bin/gollum
+++ b/bin/gollum
@@ -6,6 +6,7 @@ require 'optparse'
 require 'rubygems'
 require 'gollum'
 require 'cgi'
+require 'addressable/uri'
 
 exec         = {}
 options      = {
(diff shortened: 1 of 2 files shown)


dometto:
Thanks for the PR! I see the issue, but don't really like the solution. The fundamental problem, I think, is that we shouldn't be using to CGI escaping to escape URLs. This link mentions some alternatives that are intended for use with URIs:

https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape

Could you update the PR to use one of those methods, so we don't need the gsub? We would prefer the most lightweight implementation, if we need to require a new gem. Thanks!

Agilus:
Thanks for pointing me to that Stack Overflow conversation.  The option I chose was to grab the Addressable gem and use the escape method provided by that library.  I was tempted to go with URI::escape(), since it didn't require an extra library, but that's been marked "obsolete" for a long time now, and there's no good indication on what's going to happen to it.

Let me know if this solution works for you, or if you need anything else from me to get this right.  I'm a little new to contributing to OS projects.

dometto:
Thanks! This PR also suffers from the current Travis problems with jruby. Could you change the target of the branch to `5.x`?

@bartkamphorst do you agree with including the addressable gem?

bartkamphorst:
If all we use the gem for is this one line of code, I would favor a solution without the added dependency.

dometto:
There are a couple of other places (~4) in `gollum-lib` where we improperly use `CGI.escape` for escaping a link (and then manually `gsub` the slashes back in), also one more line in the `gollum` code (`gollum/lib/gollum/views/layout.rb:12`) and two more lines in the `gollum` test. Is that worth the dependency?

Otherwise, the best thing to do may be to define a helper method somewhere in `gollum-lib`: `escape_url`, which handles CGI escaping and re-subbing the slashes. At least then we don't have a bunch of random `gsub` statements in the codebase.

dometto:
Closing in favor of #1329.

PR #1254, opened 2017-08-07
Update sinatra and other dependencies. Closes #1193

### LINKED ISSUE #1193: Rails 5 compatibility

dometto:
Had to replace the multiline regexps in `app.rb` with single line variants to get this to work with `sinatra v.2`. Also had to explicitly disable warnings in the test suite because `rake >= 11` throws *a lot* of warnings.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec
-gem 'rake', '~> 10.4'
+gem 'rake', '~> 12', :group => :development
--- a/Rakefile
+++ b/Rakefile
@@ -71,6 +71,7 @@ Rake::TestTask.new(:test) do |test|
   test.libs << 'lib' << 'test' << '.'
   test.pattern = 'test/**/test_*.rb'
   test.verbose = true
+  test.warning = false
 end
 
 desc "Generate RCov test coverage and open in your browser"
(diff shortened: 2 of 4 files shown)


dometto:
`sinatra v.2` depends on the `mustermann` gem, which requires MRI `>= 2.2.0`. Do we want to drop support for older rubies, or go along with the `sinatra` folks? The latter is important in so far `sinatra v1.4` depends on a fairly outdated `rack`.

dometto:
And `rack` itself apparently also requires MRI `>= 2.2.2`.

bartkamphorst:
Fine to drop older rubies for `5.x`.

dometto:
@bartkamphorst did you already update `sinatra`, or was that just locally?

bartkamphorst:
I did so on the `5.x` branch (here: https://github.com/gollum/gollum/commit/b47db5611b8db1bcf761f5e95f01d371f8740657).

Issue #238, opened 2011-11-25
Using a formatted image as a link only works if no markup precedes


TimJones:
On the GitHub Wiki, I wanted to create an hyperlinked image that floated right. After tinkering with the syntax for quite a while I found this to work `[[[image.png|float|align=right]]](url)` and it generates the following HTML

``` html
<p>
  <a href="url">
    <span class="float-right">
      <span>
        <img src="/path/to/wiki/image.png">
      </span>
    </span>
  </a>
</p>
```

but it breaks as soon as anything (regular text as well as markdown) appears before it and instead creates this HTML

``` html
<p>
  <a class="internal absent" href="/path/to/wiki/float">[image.png</a>
  ](url)
</p>
```

note how the link URL contains the 'float' option instead of the image URL.

PR #1280, opened 2018-01-07
Set template_dir to prepend to search path

townsen:
* Instead of overriding default search path
  This allows you to create a template directory that contains only the
  mustache template you wish to customize, leaving the remainder of them
  in the default Gollum templates directory. I used this to change just
  the layout.

* Uses File::PATH_SEPARATOR to separate path components, if you want to
  specify more than one search directory

* It requires changes in Mustache for which I've submitted PR #238
  Alternatively point your Gemfile to townsen/mustache@multipath
  Note that this does require mustache-sinatra as a separate gem,
  and may limit you to Ruby > 2.0


dometto:
Thanks for working on this. As the new mustache-behavior is essential, we cannot merge this as is. Please resubmit this, selecting the `5.x` branch as a target, when/if your changes have made it into a new release of `mustache`.

PR #1283, opened 2018-01-15
update readme

ranvijayj:
to add 2 lines about adding source rvm path along with installation of 2 system packages required.


bartkamphorst:
@ranvijayj Thanks for the PR but I think this OS-specific installation information should go into the wiki instead.

PR #1294, opened 2018-02-28
Adds upload option 'dirpage'

hannwong:
Combines options 'dir' and 'page'.

Allows the reuse of filenames for uploaded files. Best of both worlds. Allows the consolidation of uploads in folder `uploads`, and also creates subfolder structure that mirrors that provided by option `page`.


hannwong:
Closing in favor of #1299 . Should be adding to the upcoming 5.x codes.

PR #1302, opened 2018-04-13
Handle non-image drag-and-drop uploads correctly.

onewhaleid:
### Description

```
[[/path/to/file.ext]]
```

```
```
### Example

```
**current behaviour**  
[[/Home/document.pdf]]
```
```
**proposed behaviour**  
```

image


bartkamphorst:
Thanks for this. Could you open a PR with this functionality against the `5.x` branch instead? Alternatively I can reuse your code myself, but ideally I want you to have the credit for the patch. 👍

PR #1317, opened 2018-09-17
Handle non-image drag-and-drop uploads correctly. 

onewhaleid:
### Description

```
[[/path/to/file.ext]]
```

```
```
### Example

```
**current behaviour**  
[[/Home/document.pdf]]
```
```
**proposed behaviour**  
```

image


onewhaleid:
This is my first pull request, but I think I was working on master so there are a few conflicts.

The changes specific to this pull request are all in this file:
```lib/gollum/public/gollum/javascript/editor/gollum.editor.js```

bartkamphorst:
Thanks! This way of doing it creates many conflicting files though. The better approach is to checkout the `5.x` branch yourself, reapply your fix there, and open a PR from your local 5.x branch to the project's remote 5.x branch.

onewhaleid:
Thank you for the explanation. I will try to resubmit this later.

PR #1318, opened 2018-09-19
Handle non-image uploads correctly

onewhaleid:
### Description

```
[[/path/to/file.ext]]
```

```
```
### Example

```
**current behaviour**  
[[/Home/document.pdf]]
```
```
**proposed behaviour**  
```

image

### DIFF
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
@@ -217,7 +217,15 @@
             type: 'POST',
             success: function(){
               $editorBody.removeClass('uploading');
-              var text = '[[/' + uploadDest + '/' + file.name + ']]';
+              var ext = file.name.split('.').pop().toLowerCase()
+              var image_ext = ['jpg', 'jpeg', 'tif', 'tiff', 'png', 'gif', 'svg', 'bmp']
+              // Link directly to image files
+              if ((image_ext.indexOf(ext) > -1)) {
+                var text = '[[/' + uploadDest + '/' + file.name + ']]';
+              } else {
+                // Add file name to tag for non-image files, to avoid broken image thumbnail
+                var text = '[[' + file.name + '|/' + uploadDest + '/' + file.name + ']]';
+              }
               window.ace_editor.insert(text);
             },
             error: function(r, textStatus) {


bartkamphorst:
Actually, I'm seeing different behavior than you @onewhaleid (on `5.x` without your fix). 

screen shot 2018-09-26 at 15 06 30

I'm using the RJGit adapter, but in this case I don't think that should matter. I'm using Firefox 62.0.2.

@dometto What behavior are you getting?

dometto:
On 5.x:
* Link to a nonexistent `.png` file renders "Image not found"
* Link to a nonexistent `.pdf` renders as a (red) absent link
* Link to an existing `.pdf` renders a (blue) link to the pdf, as in @bartkamphorst's screenshot

On 4.x (`master`):

All the same, except the link to the existing `.pdf` renders as in @onewhaleid's screenshot.

So conclusion: this appears to already be fixed on `5.x`, probably due to the refactoring of gollum's `[[link]]]` tags.

bartkamphorst:
> So conclusion: this appears to already be fixed on 5.x, probably due to the refactoring of gollum's [[link]]] tags.

That's good. Then the question remains whether @onewhaleid 's solution is better in terms of the link it creates. Thoughts @dometto ?

bartkamphorst:
> Well, one can always override the link's text by using [[Pretty Link|/uploads/some.pdf]]. Or am I missing something?

Correct. I'm just wondering what the best default would be.

dometto:
No objections to that default behavior 👍

dometto:
@onewhaleid asking here for lack of a better place, but I just noticed that if you upload a file using the `Upload` button, and the upload is successful, you're not given notice of this. It may be nice if the dialog (or a new dialog) tells the user the upload was successful, and tells them the location of the uploaded file (so the user can copy/paste the link). Would you be interested in further improving the upload feature in that way?

onewhaleid:
@dometto yes, I will have a look at this. Unfortunately my javascript skills are very limited (and my ruby skills nonexistent), but I'll see how I go.

onewhaleid:
I'm afraid I am a bit out of my depth here. I think the action should be taken after line 207 of `app.rb`:

```ruby
      begin
        committer = Gollum::Committer.new(wiki, options)
        committer.add_to_index(dir, filename, format, contents)
        committer.after_commit do |committer, sha|
          wiki.clear_cache
          committer.update_working_dir(dir, filename, format)
        end
        committer.commit
        # Show dialog here
        redirect to(request.referer)
```

I assume we need to call the jquery function `$.GollumDialog.init`, but I don't know how to do that from ruby.

bartkamphorst:
@dometto Okay to merge?

dometto:
👍

PR #1320, opened 2018-09-23
Remove fileview from gollum and rely on /pages.

bartkamphorst:
@dometto Any comments?


dometto:
👍 Do we want to remove whatever assets are specific to the fileview now or in a later PR?

bartkamphorst:
Are you thinking of the images? Because I was thinking of reusing those for the pages view later on (the trashcan icon, etc.).

dometto:
Good plan. I thought there might also be some specific CSS though, because of the non-standard look of the fileview. But I can't really locate any, except for the stuff to do with the icons.

dometto:
Right! All ready then :)

Issue #1196, opened 2017-02-22
1x1 pixel "delete" button in Files view in `no-edit` mode


christianrondeau:
I'm running Gollum 4.0.1 on Ubuntu 16.04.2 LTS with `--no-edit` mode.

When I select the `Files` view, there are one-pixels buttons that attempt to delete the page (which obviously won't work, since Gollum runs in `--no-edit` mode)

The HTML part that should not be rendered (or optionally hidden with a no-edit css):

    <form method="POST" action="/deleteFile/mypage" onsubmit="return confirm('Do you really want to delete the file mypage?');"><button type="submit" name="delete" value="true"></button></form>

bartkamphorst:
I can confirm this behavior on `5.x`.

bartkamphorst:
Unfortunately the HTML for this is generated in gollum-lib for some reason...

bartkamphorst:
This has been fixed in the `5.x` branch; the delete action is omitted from the HTML when run with `--no-edit`.

PR #1322, opened 2018-10-02
Implement basic file deletion for /pages view.

bartkamphorst:
This PR adds basic deletion functionality to the `/pages` view, using the old fileview's `/deleteFile/*` route. It is responsive to the ``no-edit`` flag and thereby resolves https://github.com/gollum/gollum/issues/1196. There is however no styling in place for the delete button, as styling for `5.x` needs a general overhaul anyway.


dometto:
🎆

PR #1323, opened 2018-10-02
Gollum namespace. Resolves #630

### LINKED ISSUE #630: pages named like internal routes don't display
If you name your page like `fileview.md` or `pages.md`, you can't view it. Following a link to it will get you to the specially-generated view that lists all files or pages. Adding `--base-path` doesn't help (if I use `--base-path wiki` then the internal links become `/wiki/fileview` which IMO should display the page and the special file-view display should perhaps be at `/fileview`)

dometto:
New attempt at this PR due to merge failures in the previous:

* Move gollum internal routes to the `/gollum` namespace.
* Include view helpers so we don't have static urls to assets anymore.
* Use route helpers in JS assets


dometto:
Note important changes:
* The use of asset helpers in `erb` templates for js and through custom helper methods in mustache. If we merge this we shouldn't use any hardcoded asset paths anymore, and we should keep [this](https://github.com/gollum/gollum/pull/1323/files#diff-396b9bf1f9ba99e7327f5d6535a2c193R6) updated.
* NB in `app.rb`: use `redirect_to(path)` within a namespace to redirect to `namespace/path`. `redirect to(path)` within the namespace will redirect to `/`/

gollum.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'gollum-lib', '~> 5.0.a'
   s.add_dependency 'kramdown', '~> 1.9.0'
   s.add_dependency 'sinatra', '~> 2.0'
+  s.add_dependency 'sinatra-contrib', '~> 2.0'
```
bartkamphorst: Not sure how relevant this is, but from the README of sinatra-contrib: 

NOTE: This project has been merged upstream to sinatra/sinatra.

To use this gem from the upstream repository, you can add the following to your Gemfile:
```ruby
github 'sinatra/sinatra' do
  gem 'sinatra-contrib'
end
```

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -209,7 +209,7 @@
           formData.append('file', file);
 
           $.ajax({
-            url: baseUrl + '/uploadFile',
+            url: '<%= upload_file_path %>',
```
bartkamphorst: I really like this. 👍

lib/gollum/templates/layout.mustache
```diff
@@ -5,13 +5,13 @@
   <meta name="MobileOptimized" content="width">
   <meta name="HandheldFriendly" content="true">
   <meta name="viewport" content="width=device-width">
-  {{#stylesheet_tag_mustache}}app media: :all{{/stylesheet_tag_mustache}}
-  {{#stylesheet_tag_mustache}}print media: :print{{/stylesheet_tag_mustache}}
+  {{#sprockets_stylesheet_tag}}app{{/sprockets_stylesheet_tag}}
```
bartkamphorst: I noticed that `media: all` is omitted in the new version of this line. Is that because it's a default?

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
 
 module Precious
   module Views
+    module RouteHelpers
+      ROUTES = {
+        'gollum' => {
+          :last_commit_info => 'last-commit-info',
```
bartkamphorst: Stylistically, it's becoming more common to write `last_commit_info: 'last-commit-info'` (instead of the `=>` notation).

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
...
+          :latest_changes => 'latest_changes',
```
bartkamphorst: I notice that `last-commit-info` is hyphenated, while `latest_changes` uses underscores. Any particular reason for this?

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
...
+          :upload_file => 'uploadFile',
```
bartkamphorst: Similar kind of question here: why `uploadFile` and not `upload_file` or `upload-file`? It seems we should be consistent in our style.

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
...
+          :rename => 'rename',
+          :preview => 'preview',
+          :compare => 'compare',
+          :search => 'search'
+        }
+      }
+
+      def self.parse_routes(routes, prefix = '')
+        routes.each do |name, path|
+          if path.respond_to?(:keys) then
```
bartkamphorst: You can leave off the `then` here.

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
...
+            self.parse_routes(path, "#{prefix}/#{name}")
+          else
+            define_method :"#{name.to_s}_path" do
+              "#{base_url}/#{prefix}/#{path}".gsub(/\/{2,}/, '/')
```
bartkamphorst: Is there anyway to make clearer what the gsub does here?

Also, could `gsub!` be appropriate?

lib/gollum/views/helpers.rb
```diff
@@ -2,31 +2,63 @@
...
-            if args.size > 1 then
-              options = args[1..-1].join(' ')
-              options = YAML.safe_load("---\n#{options}\n", [Symbol])
-            end
-            options = options.respond_to?(:to_h) ? options.to_h : {}
-            options = options.inject({}){|memo,(k,v)| memo[k.to_sym] = v; memo}
-            send(method, args[0], options)
-          end  
+            name = args[0]
+            options = {:media => :all}
```
bartkamphorst: Okay, this answers my earlier remark. 👍

gollum.gemspec
dometto: Cool! One less dependency

lib/gollum/views/helpers.rb
dometto: Yes, I didn't want to turn the stylesheet tag content into quasi-syntax like structures, and it turns out we only ever use `{:media => :all}`. If someone ever designs a mobile frontend we'll adapt. ;)

lib/gollum/views/helpers.rb
bartkamphorst: I think this should actually be resolved in a separate PR.

lib/gollum/views/helpers.rb
dometto: I think `gsub!` might end up modifying `routes`, but I think I can get rid of the substitution entirely by only conditionally including `prefix`.

lib/gollum/views/helpers.rb
dometto: 👍

lib/gollum/views/helpers.rb
dometto: That's not going to work unfortunately, because `base_url` is another source of double slashes (in case it is `'/'`). Do you have a proposal for a better readable `gsub`? Otherwise I could just add a comment.

bartkamphorst: A comment would be sufficient. 👍

bartkamphorst:
Cool. 👍 If you can resolve the merge conflicts then I think this is good to go.

dometto:
Merging was non-trivial so I don't want to push this into `5.x` without viewing the changes in a new PR. Closing in favor of #1328

PR #1324, opened 2018-10-02
Secure custom JS and CSS. Resolves #665

### LINKED ISSUE #665: Secure custom css/js
There should be a secure way to inject CSS and JS. Currently editing `custom.js` is not possible via the web UI because `.` is translated to `dot`

dometto:
Lockdown write access to `custom.css` and `custom.js`, as per https://github.com/gollum/gollum/issues/665.

(The solution I suggest there, to add the custom files to the precompile path, does not seem appealing given recent troubles with asset compilations. Moreover, for users who *want* to have these files committed in the repo, locking down access is still required.)

Are there any routes that I have overlooked? One thing I wonder is whether one couldn't theoretically use `uploadFile` to plug a custom css or js file.

NB: once #1323 is merged the routes in the tests for this PR need to be updated.


test/test_app.rb
```diff
@@ -505,6 +505,31 @@ def nfd utf8
...
+        assert_equal 403, last_response.status, "post /#{route}/custom#{ext} -- #{last_response.inspect}"
+      end
+    end
+
+    ['.css', '.js'].each do |ext|
+      post "/revert/custom#{ext}/02796b1450691f90db5d6dc6a816a4980ce80d07/2f6485c2702c7c8b9b6613672337ffa7d933ddcf"
+      assert_equal 403, last_response.status, "post /revert/custom#{ext} -- #{last_response.inspect}"
+    end
+
+    Precious::App.set(:wiki_options, { :js => nil })
```
bartkamphorst: What is this for?

bartkamphorst: Ah, as a reset for the next set of tests, correct?

bartkamphorst:
Looks solid. 👍

test/test_app.rb
dometto: Indeed!

dometto:
Additional questions:

* Do we want to define the blocking routes always, or just when `--css` or `--js` are enabled?
* The same logic could easily be extended to let the user define other pages which can't be edited, so as to have e.g. 'Home' as an uneditable landing page. Do we want that? (One challenge I just thought of is that the 'Edit' button would still appear on pages locked down in this way.)

lib/gollum/app.rb
```diff
@@ -144,6 +144,14 @@ class App < Sinatra::Base
       end
     end
 
+    get %r{\/(edit|create)\/custom.(js|css)} do
+      forbid
+    end
+
+    post %r{\/(deleteFile|rename|edit|revert|create)\/custom.(js|css)(\/.*)?} do
```
dometto: Because of the `(\/.*)?` at the end, which is needed to match the `revert` route, this regex also blocks e.g. `edit/custom.css/Foo.md`. A bit of an odd usecase, but perhaps still sloppy?

dometto:
>  Do we want to define the blocking routes always, or just when --css or --js are enabled?
    The same logic could easily be extended to let the user define other pages which can't be edited, so as to have e.g. 'Home' as an uneditable landing page. Do we want that? (One challenge I just thought of is that the 'Edit' button would still appear on pages locked down in this way.)

Both of these are hard to accomplish because the wiki options (including `:css` and `:js`) are not available to the `Precious::App` class (only to the instance), and so we cannot use them to conditionally define routes. Forgetting about this. :)

> Note to self: currently the error message thrown is Forbidden. This wiki is set to no-edit mode. That's not accurate :)

Fixed.

PR #1326, opened 2018-10-05
Fix url generation in pages view. 

bartkamphorst:
Also:
* Show full filename with extension for pages. 
* Distinguish in HTML between file and pages, allowing for different icons in the future.

### DIFF
--- a/lib/gollum/views/pages.rb
+++ b/lib/gollum/views/pages.rb
@@ -37,34 +37,34 @@ module Precious
       def files_folders
         if has_results
           folders = {}
-          page_files = {}
+          files = {}
 
           # 1012: Folders and Pages need to be separated
-          @results.each do |page|
-            page_path = page.path
-            page_path = page_path.sub(/^#{Regexp.escape(@path)}\//, '') unless @path.nil?
+          @results.each do |result|
+            result_path = result.path
+            result_path = result_path.sub(/^#{Regexp.escape(@path)}\//, '') unless @path.nil?
             
 
-            if page_path.include?('/')
-              folder      = page_path.split('/').first
+            if result_path.include?('/')
+              # result contains a folder
+              folder      = result_path.split('/').first
               folder_path = @path ? "#{@path}/#{folder}" : folder
               folder_link = %{<li><a href="#{@base_url}/pages/#{folder_path}/" class="folder">#{folder}</a></li>}
 
               folders[folder] = folder_link unless folders.key?(folder)
-            elsif page_path != ".gitkeep"
-              if defined? page.format
-                url = "#{@base_url}/#{page.escaped_url_path}"
-              else
-                url = "#{@base_url}/#{page.escaped_url_path}#{page.name}"
-              end
-              page_link = %{<li><a href="#{url}" class="file">#{page.name}</a>#{delete_file(url) if @allow_editing}</li>}
-              page_files[page.name] = page_link
+            elsif result_path != ".gitkeep"
+              # result is either a valid gollum page or another type of file
+              klass = (defined? result.format) ? "page" : "file"
+              url = "#{@base_url}/#{result.escaped_url_path}"
+              
+              file_link = %{<li><a href="#{url}" class="#{klass}">#{result.filename}</a>#{delete_file(url) if @allow_editing}</li>}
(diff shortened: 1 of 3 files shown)


lib/gollum/views/pages.rb
```diff
@@ -37,34 +37,34 @@ def delete_file(url)
...
           folders = {}
-          page_files = {}
+          files = {}
 
           # 1012: Folders and Pages need to be separated
-          @results.each do |page|
-            page_path = page.path
-            page_path = page_path.sub(/^#{Regexp.escape(@path)}\//, '') unless @path.nil?
+          @results.each do |result|
+            result_path = result.path
```
dometto: is `result.path` escaped, and if not, is there potential for JS-insertion in line 52 below?

dometto: Asking because in `master` we're currently sanitzing the whole HTML block outputted by this method. But it would be preferable to sanitize only the bits that are potentially user-generated.

lib/gollum/views/pages.rb
```diff
@@ -37,34 +37,34 @@ def delete_file(url)
...
 
-            if page_path.include?('/')
-              folder      = page_path.split('/').first
+            if result_path.include?('/')
+              # result contains a folder
+              folder      = result_path.split('/').first
               folder_path = @path ? "#{@path}/#{folder}" : folder
               folder_link = %{<li><a href="#{@base_url}/pages/#{folder_path}/" class="folder">#{folder}</a></li>}
 
               folders[folder] = folder_link unless folders.key?(folder)
```
dometto: Line 39 sets `folders = {}`, so what's the point of the `unless`?

lib/gollum/views/pages.rb
```diff
@@ -37,34 +37,34 @@ def delete_file(url)
...
-            elsif page_path != ".gitkeep"
-              if defined? page.format
-                url = "#{@base_url}/#{page.escaped_url_path}"
-              else
-                url = "#{@base_url}/#{page.escaped_url_path}#{page.name}"
-              end
-              page_link = %{<li><a href="#{url}" class="file">#{page.name}</a>#{delete_file(url) if @allow_editing}</li>}
-              page_files[page.name] = page_link
+            elsif result_path != ".gitkeep"
```
dometto: Nice catch of the `.gitkeep` edge case 👍

lib/gollum/views/pages.rb
```diff
@@ -37,34 +37,34 @@ def delete_file(url)
...
+              # result is either a valid gollum page or another type of file
+              klass = (defined? result.format) ? "page" : "file"
+              url = "#{@base_url}/#{result.escaped_url_path}"
+              
+              file_link = %{<li><a href="#{url}" class="#{klass}">#{result.filename}</a>#{delete_file(url) if @allow_editing}</li>}
```
dometto: Same question about sanitization for `result.filename`. `result.escaped_url_path` above already sanitizes fortunately:
```ruby
ERB::Util.url_encode('http://test.com"<script>alert("hello");</script>/test')
 => "http%3A%2F%2Ftest.com%22%3Cscript%3Ealert%28%22hello%22%29%3B%3C%2Fscript%3E%2Ftest"
```

lib/gollum/views/pages.rb
bartkamphorst: I don't think `result.path` is, but the outputted html uses `result.escaped_url_path`.

lib/gollum/views/pages.rb
bartkamphorst: Not mine, but yeah. :)

lib/gollum/views/pages.rb
bartkamphorst: I don't know how `result.filename` could be exploited in this scenario. Is it a showstopper for the PR though?

dometto: Definitely not a showstopper but we need to remember to think of whether any escaping is necessary, as in `master` we now have [this](https://github.com/gollum/gollum/blob/master/lib/gollum/views/pages.rb#L63) and [this](https://github.com/gollum/gollum/blob/master/lib/gollum/views/pages.rb#L26).

PR #1328, opened 2018-10-14
Namespace

dometto:
New attempt at this PR due to merge failures in the [previous](https://github.com/gollum/gollum/pull/1323)
  * Move gollum internal routes to the /gollum namespace.
  * Include view helpers so we don't have static urls to assets anymore.
  * Use route helpers in JS assets

Note important changes:
  * The use of asset helpers in erb templates for js and through custom helper methods in mustache. If we merge this we shouldn't use any hardcoded asset paths anymore, and we should keep this updated.
 * NB in app.rb: use redirect_to(path) within a namespace to redirect to namespace/path. By contrast, redirect to(path) (!) within the namespace will redirect to /path. The difference is just the underscore, which is annoying -- but that's imported from sinatra-contrib.


dometto:
Looks like we still require the `sinatra-contrib` gem. It's still a separate gem, only the repository has been merged into `sinatra`.

bartkamphorst:
Great stuff! 🎉

PR #1330, opened 2018-10-14
Implement header counting using YAML frontmatter. Resolves #908

### LINKED ISSUE #908: Header numbering
@SkyCrawl in #906:

> Gollum should, in my opinion, offer to number headers automatically. As an example, this is what I have in mind. The best way to achieve it would probably be an "annotation" like [[_TOC_]], perhaps [[_HN_]]?

@sunny on achieving this with Macro's or syntax helpers:

> Even if it's not on the body you would still need some DOM that wraps the whole text, at the least. Or get some help from JavaScript.

Dumping javascript into the page's rendered data doesn't seem very tidy. But now that I think of it, if we would implement a syntax helper `[[_HN_]]`, we could make it replace those tags and wrap the rest of the data in a `<div class="counter">` or whatever. Would that work?

dometto:
Use YAML frontmatter to enable header counting via CSS:
```
---
countheads: true
---
```
Result:
screen shot 2018-10-15 at 01 19 42

But it is also possible to specify another counter style (also e.g. roman numerals, see `VALID_COUNTER_STYLES` in `page.rb`):

```
---
countheads: 'tibetan'
---
```
Result:
screen shot 2018-10-15 at 01 19 28

### DIFF
--- a/lib/gollum/public/gollum/javascript/gollum.js.erb
+++ b/lib/gollum/public/gollum/javascript/gollum.js.erb
@@ -151,8 +151,6 @@ $(document).ready(function() {
     }
   }
 
-  $('.countheads').get(0).style.setProperty('--countheads-style', $('.countheads').data('countheads-style'));
-
   if ($('#minibutton-upload-page').length) {
     $('#minibutton-upload-page').parent().removeClass('jaws');
     $('#minibutton-upload-page').click(function(e) {
--- a/lib/gollum/public/gollum/javascript/gollum.js.erb
+++ b/lib/gollum/public/gollum/javascript/gollum.js.erb
@@ -151,6 +151,8 @@ $(document).ready(function() {
     }
   }
 
+  $('.countheads').get(0).style.setProperty('--countheads-style', $('.countheads').data('countheads-style'));
+
   if ($('#minibutton-upload-page').length) {
     $('#minibutton-upload-page').parent().removeClass('jaws');
     $('#minibutton-upload-page').click(function(e) {
--- a/lib/gollum/public/gollum/stylesheets/template.scss
+++ b/lib/gollum/public/gollum/stylesheets/template.scss
@@ -92,6 +92,7 @@ body {
   font: 13.34px $font-default;
   font-size: small;
   line-height: 1.4;
+  counter-reset: h1;
 }
 
 img {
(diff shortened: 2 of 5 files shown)


dometto:
Two questions: 
* Do we want "1.2." as it is now, or rather "1.2"? (I take it we do want "1." and not "1")
* Call the option `countheads` or something else?

bartkamphorst:
> Do we want "1.2." as it is now, or rather "1.2"? (I take it we do want "1." and not "1")

I like "1.2." as it is now.

> Call the option `countheads` or something else?

What about `header_enumeration`?

dometto:
> I like "1.2." as it is now.
👍 

> What about header_enumeration?
Maybe a bit long. What about `count_headers` or simply `counter`?

Here's an idea I just had: why don't we add a `--metadata` (?) switch to gollum where you can specify a `.yaml` file that contains global metadata, then merge that with the frontmatter the user sets in each page (so that what's in the global file can be overridden on a per-page basis)? That way it will be easy to globally turn on header enumeration, but also `h1_titles`, `display_metadata` (we can lose or keep the current switches for those), and whatever the user wants. Does that sound good?

bartkamphorst:
> Maybe a bit long. What about `count_headers` or simply `counter`?

I think `count` (and `counter` as well, though to a lesser extent) generally suggests addition (e.g., total number of headers). Enumeration better expresses what the key is meant for. Maybe `header_enum` as a compromise? 

> why don't we add a --metadata (?) switch to gollum where you can specify a .yaml file that contains global metadata, then merge that with the frontmatter the user sets in each page (so that what's in the global file can be overridden on a per-page basis)?

I like the functionality, though I'm a bit hesitant to add this as (yet another) cli flag. I prefer a solution (or even a snippet in the wiki) using `config.rb` to achieve this.

bartkamphorst:
BTW, the functionality is pretty sweet; nicely done. 👍

dometto:
> I think count (and counter as well, though to a lesser extent) generally suggests addition (e.g., total number of headers). Enumeration better expresses what the key is meant for. Maybe header_enum as a compromise?

Good point, `header_enum` it is 👍 

> I like the functionality, though I'm a bit hesitant to add this as (yet another) cli flag. I prefer a solution (or even a snippet in the wiki) using config.rb to achieve this.

Yeah, we could add the option, but leave out the switch. 👍 

> BTW, the functionality is pretty sweet; nicely done. 👍

Thanks! 😄

PR #1331, opened 2018-10-16
Allow searching multiple keywords

### LINKED ISSUE #615: Searching multiple keywords
It seems the current version of Gollum (2.4.10) does not support searching multiple keywords simultaneously across the wiki site. For example, if a page contains the following content:

> Problems with the software includes...

Searching `problems software` will not return this page; although using a regular expression like `problems.*software` works - but that will require the order between "problems" and "software" to be the same as it appears in the page. 

It would be great if Gollum could support searching multiple keywords as this is an important feature for a wiki system.

m3nd3s:
Git grep supports regex, so in this PR the term (q) is splited by any space caracter and joined using pipe (or in regex).

Now it is possible to search by multiple keywords.

This PR resolve #615.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -477,7 +477,8 @@ module Precious
       @query   = params[:q] || ''
       wiki     = wiki_new
       # Sort wiki search results by count (desc) and then by name (asc)
-      @results = wiki.search(@query).sort { |a, b| (a[:count] <=> b[:count]).nonzero? || b[:name] <=> a[:name] }.reverse
+      query_regex = @query.split(/\s/).join('|')
+      @results = wiki.search(query_regex).sort { |a, b| (a[:count] <=> b[:count]).nonzero? || b[:name] <=> a[:name] }.reverse
       @name    = @query
       mustache :search
     end


dometto:
@m3nd3s great, thanks for working on this! The fix itself is nice and clean, but I did just think of a complications: what if I want to search for `|`, or some other regex character? In other words, if the query is going to be passed to `gollum-lib` as a regex, we should perform a `Regexp.escape` first, no?

It would also be great if you could add some tests for the new behavior, including a test to guard against the above mentioned edge case. You should probably add them to `test_app.rb`.

dometto:
Just another note: it looks as if this strategy will currently not work with the `rugged_adapter`, because it is [escaping the search query](https://github.com/gollum/rugged_adapter/blob/master/lib/rugged_adapter/git_layer_rugged.rb#L170).

That may be the same for `grit` -- I seem to recall the escaping was added for security reasons, so we won't be able to get rid of that. In `5.x` we're planning to get rid of `grit` though, if the `rugged_adapter` is fully functional in time.

The `rjgit_adapter` does not appear to escape the query. We'll see how much of it works when you add the tests!

We could also add a feature to optionally enable full-regex based searching, in which case we do no escaping and do not apply the logic from this PR, but just pass the query on directly. Perhaps the search box could have a checkbox or something to enable that.

dometto:
> In 5.x we're planning to get rid of grit though, if the rugged_adapter is fully functional in time.

Ah, I just noticed you opened this PR against the `master` branch. Could you reopen it against `5.x`?

m3nd3s:
Hi @dometto , you're right!

So, I'll put a checkbox to enable "regex mode", with this option enabled the user will be able to perform search using regex. In normal mode the _term_ will [be escaped](http://ruby-doc.org/core-2.5.1/Regexp.html#method-c-escape).

What do you think?

I'll close this PR and open another to `5.x`, ok?

dometto:
Sounds great!

dometto:
I just noticed there's little space next to the search bar in the 'normal' page view. Perhaps we should only have the checkbox in the [search view](https://github.com/gollum/gollum/blob/5.x/lib/gollum/templates/search.mustache)?

dometto:
@m3nd3s, still interested in working on this?

m3nd3s:
Hi @dometto , yes I still interested.

I was not able to work on this last weeks, but I'm still interested.

Can you give me some days? Perhaps next week I can solve this. OK?

dometto:
Sure! Let us know if we can help. Thanks for helping out.

Issue #1333, opened 2018-10-17
Print view


dometto:
Currently printing a gollum page will give you all kinds of clutter from the header and footer. We should have a 'Print' button that allows one to view only the `wiki-content` div (and optionally fires the browser's 'print' dialog with JS). This could be achieved either by copying the div to a new window's DOM, or by hiding the other elements on the page. (A prima facie problem with the latter is that the user will have no way to return to normal view, as we'll have hidden all the buttons  😄). Perhaps this should also change the CSS media option to 'print', so users can optionally use self-defined print CSS.

bartkamphorst:
There already is a `print.scss` that sets `display: none` for two links at the bottom of the page. Perhaps we can extend that?

dometto:
Actually, you're right. All that's needed is to add the `div#footer{ display: none;}`.

dometto:
A bit more explanation: I was wrong about the buttons being shown, and although I personally prefer to lose the entire header div when printing, many users will want the page title to be printed. At any rate, they can already make their own choices by adding custom CSS. To close this issue, all that's necessary is to hide the entire `div#footer`.

bartkamphorst:
Sweet! 👍

dometto:
Implemented in 5.x

PR #1335, opened 2018-10-17
updated print.scss to remove footer when printing

lpbearden:
This should resolve issue #1333. Added in display:none for the footer, and tested locally.


bartkamphorst:
Nice, thanks! 👍

PR #1336, opened 2018-10-17
Unify deletion routes. Closes #1332.

### LINKED ISSUE #1332: 5.x: unify /delete and /deleteFile routes
Currently, there are two deletion routes that seem to be doing pretty much the same thing, except one is geared specifically to pages and the other can handle both pages and other files. I suggest to identify the relevant differences and unify the two routes.

bartkamphorst:


lib/gollum/app.rb
```diff
@@ -153,7 +153,7 @@ class App < Sinatra::Base
         forbid('Changing this resource is not allowed.')
       end
 
-      post %r{/(deleteFile|rename|edit|create)/custom\.(js|css)} do
+      post %r{/(delete|rename|edit|create)/custom\.(js|css)} do
```
dometto: Nice catch, basically I had forgotten about the `delete` route in securing the special files... thank! 👍

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -47,8 +47,12 @@ $(document).ready(function() {
   $('#delete-link').click( function(e) {
     var ok = confirm($(this).data('confirm'));
     if ( ok ) {
-      var loc = '<%= delete_path %>/' + pageFullPath;
-      window.location = loc;
```
dometto: Argh, that was ugly. Much better your way.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -47,8 +47,12 @@ $(document).ready(function() {
...
+      $.post('<%=delete_path %>/' + pageFullPath,
+        {},
+        function (result) {
+          // page successfully deleted, return to landing page
+          window.location = '/';
```
dometto: The route itself is redirecting to `/pages`. I'm fine with either behavior really, but I'm not sure what will end up happening here, as the `delete` route is being triggered, but the ajax call is redirecting somewhere else?

bartkamphorst: This works, but I agree that it isn't very pretty. Going forward, I think the `/pages` view should be refactored anyway so that the delete there also uses ajax (so that the overview is updated without reloading the entire page). Until then, I suggest we keep it like this.

bartkamphorst: @dometto Is this okay with you?

dometto: :+1: sounds right

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -47,13 +47,33 @@ $(document).ready(function() {
...
```
bartkamphorst: @dometto I implemented JS deletion for `/pages` so that the redirect statement in the `/delete` route can be removed with this PR. Cleaner this way. Thoughts?

dometto:
Very nice on the new ajax call 👍

bartkamphorst:
Cool. Will fix in a moment.

dometto:
🎆 🎉

PR #1337, opened 2018-10-19
Use snake_case uniformly for all routes.

bartkamphorst:


lib/gollum/public/gollum/javascript/gollum.dialog.js
```diff
@@ -101,7 +101,7 @@
 
       var id = fieldAttributes.id || 'upload';
       var name = fieldAttributes.name || 'file';
-      var action = fieldAttributes.action || '/uploadFile';
+      var action = fieldAttributes.action || '/upload_file';
```
dometto: You could consider moving this file to `gollum.dialog.js` and using the path helper here, too

dometto: Correction: move it to `gollum.dialog.js.erb`

dometto:
Code zen! I love how little needs to be changed now that we have the path helpers :)

bartkamphorst:
Absolutely, I was pleasantly surprised by that! Kudos!

lib/gollum/public/gollum/javascript/gollum.dialog.js
bartkamphorst: Agreed. I'm planning to restructure the uploading code base, though, so I'll do that in another PR if you don't mind.

PR #1338, opened 2018-10-19
Use path helpers consistently

dometto:
Found a couple of instances where the javascript was not using path helpers. I did some extensive grepping, including for `'/`, `baseUrl`, and all the currently existing route names, and moved to using the path helpers.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -191,7 +191,7 @@ task :precompile do
       config.environment = env
       config.prefix      = Precious::Assets::ASSET_URL
       config.digest      = true
-      config.public_path = './public/assets'
+      config.public_path = ENV.fetch('assets_path', './public/assets')
       config.manifest    = manifest
     end
   puts "Precompiling assets to #{::File.expand_path('./public/assets')}..."
--- a/Rakefile
+++ b/Rakefile
@@ -191,7 +191,7 @@ task :precompile do
       config.environment = env
       config.prefix      = Precious::Assets::ASSET_URL
       config.digest      = true
-      config.public_path = ENV.fetch('assets_path', './public/assets')
+      config.public_path = ENV.fetch('GOLLUM_ASSETS_PATH', './public/assets')
       config.manifest    = manifest
     end
   puts "Precompiling assets to #{::File.expand_path('./public/assets')}..."
--- a/Rakefile
+++ b/Rakefile
@@ -177,3 +177,23 @@ task :validate do
     exit!
   end
 end
+
+desc 'Precompile assets'
+task :precompile do
+ require './lib/gollum/views/helpers.rb'
+ require './lib/gollum/assets.rb'
+ require 'sprockets'
+ require 'sprockets-helpers'
+ require 'sass'
+  env = Precious::Assets.sprockets
+  manifest = Sprockets::Manifest.new(env, './public/assets')
+    Sprockets::Helpers.configure do |config|
+      config.environment = env
+      config.prefix      = Precious::Assets::ASSET_URL
+      config.digest      = true
+      config.public_path = './public/assets'
+      config.manifest    = manifest
+    end
+  puts "Precompiling assets to #{::File.expand_path('./public/assets')}..."
+  manifest.compile(Precious::Assets::MANIFEST)
+end
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
@@ -151,7 +151,7 @@
           // get form fields
           var oldAction = $('#gollum-editor form').attr('action');
           var $form = $($('#gollum-editor form').get(0));
-          $form.attr('action', this.href || '<%= preview_path >');
+          $form.attr('action', this.href || '<%= preview_path %>');
           $form.attr('target', '_blank');
           var paths = window.location.pathname.split('/');
           $form.attr('page', paths[ paths.length - 1 ] || '')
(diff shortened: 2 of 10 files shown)


dometto:
One of the things I found was `gollum.editor.js.erb:12225`:
`          $.post('/wiki/help?_method=delete');`
I honestly have no idea what that's supposed to do. Do you think we can remove it?

dometto:
Also, remind me again while we still have a `fileview.scss`?

dometto:
Just realized this: the current setup would make it impossible to use `base_path` with static assets, because the paths to routers in the assets are generated at asset precompilation time. So we should, after all, prefix `baseUrl` in the JS. We can do this cleaner than previously, though, by having a JS function that prefixes `baseUrl` and consistently using that wherever we're now using a path helper.

bartkamphorst:
> So we should, after all, prefix baseUrl in the JS.

Good observation. I take it you no longer need my review for this PR then, right?

dometto:
I'll make the required changes and then ask for a review!

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -405,43 +398,6 @@
       return null;
     },
 
-
```
dometto: Turns out the `loadFor` method wasn't doing anything anymore, since we're requiring all the language defintions in `app.js`, so was able to remove this block.

dometto:
@bartkamphorst I've now added a `routePath` helper method to the JS which borrows the route definitions from the route helpers. Using `routePath` in the JS thus allows us to easily generate urls, while prefixing `baseUrl` on the go (if necessary).

I've also:
* addressed #1355 
* caught a bug with the file deletion button in the pages view when using basepath (please check)
* moved the precompilation task to the rakefile

I've tested with static and dynamic assets, with and without basefile, and wasn't able to catch any bugs. But please inspect. :)

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
bartkamphorst: 👍

bartkamphorst:
> $.post('/wiki/help?_method=delete');  I honestly have no idea what that's supposed to do. Do you think we can remove it?

No, I believe this is used for showing or removing the "help" functionality in the editor.

bartkamphorst:
> Also, remind me again while we still have a fileview.scss?

Only as a reminder / example for how to improve some aspects of the Overview page.

bartkamphorst:
> I've now added a routePath helper method to the JS which borrows the route definitions from the ruby route helpers.

Elegantly done. 👍

bartkamphorst:
> caught a bug with the file deletion button in the pages view when using basepath

Your change seems sensible. I'll test everything out tomorrow.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -1,4 +1,27 @@
 // Helpers
+
+// Get path for named route, prefixing baseUrl if necessary
+// Uses the route definitions in /lib/gollum/views/helpers.rb.
+// For example, routePath('delete') is equivalent to 'delete_path' in the mustache templates.
+function routePath(name){
+  var routes = $.parseJSON('<%= routes_to_json %>')
+  path = routes[name]
+  if (baseUrl == undefined ) {
+    console.log('Gollum error: baseUrl undefined')
```
bartkamphorst: When does this happen (baseUrl == undefined)? And if it happens, does it break functionality?

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -405,43 +398,6 @@
...
-    /**
-     *  loadFor
```
bartkamphorst: Are we sure that this is deprecated and no longer used?

Rakefile
```diff
@@ -177,3 +177,23 @@ task :validate do
...
+
+desc 'Precompile assets'
+task :precompile do
+ require './lib/gollum/views/helpers.rb'
+ require './lib/gollum/assets.rb'
+ require 'sprockets'
+ require 'sprockets-helpers'
+ require 'sass'
+  env = Precious::Assets.sprockets
+  manifest = Sprockets::Manifest.new(env, './public/assets')
```
bartkamphorst: It would be nice to add an option here to compile the assets to a custom location.

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: It would happen if we somehow forget to set the `baseUrl` variable in the page template. I thought it would be useful to add the debug message in case that ever happens.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
dometto: Define "sure" ;)

Rakefile
dometto: 👍

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: 👍

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
bartkamphorst: OK.

bartkamphorst:
Actually, did you resolve https://github.com/gollum/gollum/issues/1356 in this PR?

Issue #1327, opened 2018-10-10
5.x: uploading with Ace editor does not show drag/drop/uploading CSS


bartkamphorst:
While test-driving 5.x, I noticed that the editor is no longer visually responsive to [dragging and dropping and uploading](https://github.com/gollum/gollum/blob/master/lib/gollum/public/gollum/css/editor.css#L343-L362). I suspect this has to do with the recent change to the ACE editor. Specifically, I think there must have been a change in the HTML that makes it that [this Javascript](https://github.com/gollum/gollum/blob/master/lib/gollum/public/gollum/javascript/editor/gollum.editor.js#L102-L116) no longer has any effect.

```
Gollum 5.0.1b
Running on: java with Ruby version 2.3.3
Using:
rjgit 5.1.1.0
gollum-rjgit_adapter 0.3.4
gollum-lib 5.0.a.4
With the following renderers:
kramdown 1.17.0
```

bartkamphorst:
@ProgramFan, would you mind investigating? Both the JS and the CSS seem to be loaded, so I suspect it's an issue with `div` ids having changed.

bartkamphorst:
Ping @ProgramFan. 😄

ProgramFan:
I will check it.

bartkamphorst:
Terrific, thank you!

ProgramFan:
It turns out an ajax error when uploading. The error message is 'error' and gives no information. What can be the cause? Any ideas?

ProgramFan:
Gollum prints these errors when uploading:
```
2018-10-28 18:33:38 - NoMethodError - undefined method `tree' for nil:NilClass:
	/usr/local/lib/ruby/gems/2.5.0/gems/gitlab-grit-2.8.2/lib/grit/lazy.rb:31:in `block (2 levels) in lazy_reader'
	/usr/local/lib/ruby/gems/2.5.0/gems/gollum-grit_adapter-1.0.1/lib/grit_adapter/git_layer_grit.rb:115:in `tree'
	/usr/local/lib/ruby/gems/2.5.0/gems/gollum-lib-5.0.a.4/lib/gollum-lib/committer.rb:45:in `index'
	/usr/local/lib/ruby/gems/2.5.0/gems/gollum-lib-5.0.a.4/lib/gollum-lib/committer.rb:94:in `add_to_index'
[...]

PR #1341, opened 2018-10-28
Fix absense of visual feedback when uploading with drag-and-drop

ProgramFan:
Address #1327


ProgramFan:
I would prefer a separate entry as they serve different purposes like that of `#gollum-editor-body`.

No spinner at the moment. ACE editor takes over the place of the original editor body and where shall we display the spinner? I failed to find the spinner in the CSS. 😢

ProgramFan:
All the concerns are addressed in the latest commit. The spinner is now shown when uploading and the CSS is cleaned up.

dometto:
Great!

bartkamphorst:
Works like a charm. Thanks for resolving this, @ProgramFan !

Issue #1016, opened 2015-05-26
CriticMarkup


SkyCrawl:
Just found a potentially interesting "markup" to implement:
https://github.com/CriticMarkup/CriticMarkup-toolkit/#the-basic-syntax

Gollum can already provide some of the functionality and the proposed HTML-diff would only enhance it, even if not completely. On the other hand, having all possible versions of something displayed in the page right away like [CriticMarkup suggests](http://criticmarkup.com/) might be beneficial as well. 
Implementation should be simple - macros  :). Before the source content reaches the renderer, that is...

**EDIT:** I think it could be a worthwhile convenience, especially after Gollum allows to redefine the way it generates HTML snippets (on user-level). That's actually another of my changes that I have implemented in my forks, despite not all the code has been rewritten to respect them yet.

What do you think?

bartkamphorst:
I recently had a use case for CriticMarkup and I quite liked it. So +1 for adding it to gollum at some point down the line.

bartkamphorst:
Implemented in `5.x`. 🙌

PR #1345, opened 2018-11-08
Implements frontend support for CriticMarkup. Related to #1016.

bartkamphorst:
This PR implements styling for CriticMarkup (backend support is already in gollum-lib: https://github.com/gollum/gollum-lib/pull/308). 

screen shot 2018-11-08 at 16 18 21

In addition, this PR also adds two buttons to the editor for accepting or rejecting proposed changes in selected text.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -135,6 +135,9 @@ MSG
     "By default, uses the 'TeX-AMS-MML_HTMLorMML' config with the 'autoload-all' extension.") do
     wiki_options[:mathjax] = true
   end
+  opts.on("--critic-markup", "Enable CriticMarkup support.") do
+    wiki_options[:critic_markup] = true
+  end
   opts.on("--irb", "Launch Gollum in 'console mode', with a predefined API.") do
     options[:irb] = true
   end
--- a/bin/gollum
+++ b/bin/gollum
@@ -135,7 +135,7 @@ MSG
     "By default, uses the 'TeX-AMS-MML_HTMLorMML' config with the 'autoload-all' extension.") do
     wiki_options[:mathjax] = true
   end
-  opts.on("--critic-markup", "Enable CriticMarkup support.") do
+  opts.on("--critic-markup", "Enable support for annotations using CriticMarkup.") do
     wiki_options[:critic_markup] = true
   end
   opts.on("--irb", "Launch Gollum in 'console mode', with a predefined API.") do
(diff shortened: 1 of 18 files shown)


bin/gollum
```diff
@@ -135,6 +135,9 @@ MSG
     "By default, uses the 'TeX-AMS-MML_HTMLorMML' config with the 'autoload-all' extension.") do
     wiki_options[:mathjax] = true
   end
+  opts.on("--critic-markup", "Enable CriticMarkup support.") do
```
dometto: Perhaps something more descriptive for people who don't know CM yet:
`Enable support for annotations using CriticMarkup`
?

lib/gollum/public/gollum/javascript/editor/langs/asciidoc.js
```diff
@@ -113,7 +113,34 @@ var AsciiDoc = {
                                   }
                                 });
                               }
-                            }
+                            },
+                            
+  'function-critic-accept' : {
```
dometto: Having to repeat these functions is obviously not very DRY (by the looks of it, there are no differences for the various language definitions, right?). Do you see a way of making it more so? If not, then never mind -- this PR shouldn't be held up because of the way language definitions are currently working.

lib/gollum/public/gollum/stylesheets/editor.scss
```diff
@@ -367,8 +367,8 @@ a#function-h2 { @include editor-button(9); }
 a#function-h3 { @include editor-button(10); }
 a#function-link { @include editor-button(11); }
 a#function-image { @include editor-button(12); }
-a#function-plus { @include editor-button(13); }
-a#function-minus { @include editor-button(14); }
```
dometto: Why are the new critic functions coming the place of `plus` and `minus`?

lib/gollum/views/edit.rb
```diff
@@ -60,6 +60,10 @@ def is_edit_page
         true
       end
 
+      def critic_markup
```
dometto: Isn't this superfluous given that `critic_markup` is defined in `layout.rb`? Or is the `Edit` view not a subclass of `Layout`?

bartkamphorst: You are right. Will fix.

bartkamphorst:
@dometto OK like this?

dometto:
🎆

PR #1347, opened 2018-11-10
Force path variable in pages route to end with /. Resolves #1241

### LINKED ISSUE #1241: Breadcrumbs (All pages) navigation doesn't work when Gollum runs inside a Rails app
In the All Pages view, clicking on a directory that has a child directory should update the breadcrumbs trail control and take the user to the next location within a page's hierarchical structure, however, this doesn't work when Gollum is integrated into a Rails application. One of the incentives to integrating Gollum into a Rails application is that you can leverage a Rails' authentication OAuth2/Device/* and any additional custom authentication policies that an application might already have in place to also authenticate Gollum Wiki users (eliminates the need for omnigollum and so forth). Authentication and most of the Gollum's common functionally seems to be working fine except the breadcrumbs. Wandering if anyone hit the same issues before.

dometto:
The current logic for the /pages routes assumes that the path the user requests ends with a slash. Because of the way the `extract_path` helper works, a request for `/gollum/pages/a/b` will result in the path `a`, and only `/gollum/pages/a/b/` (with trailing slash) will result int the path `a/b` being displayed.

According to #1241 embedding gollum in a rails application has the consequence that the trailing slash removed from the request path that's handed over to gollum. I don't know if that's still an issue (don't have a quick way to reproduce), but I can see how using gollum with rack middleware could result in such situations. The current PR just forces the presence of a trailing slash. I don't think that has any ill consequences, and it may help to solve/prevent such issues in the future. Without a testing environment I'm shooting blind here, however.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -454,7 +454,9 @@ module Precious
           /(.+) # capture any path after the "/pages" excluding the leading slash
         )?      # end the optional non-capturing group
       }x do |path|
-        @path        = extract_path(path) if path
+        if path
+          @path = path[-1] == '/' ? extract_path(path) : extract_path("#{path}/")
+        end
         wiki_options = settings.wiki_options.merge({ :page_file_dir => @path })
         wiki         = Gollum::Wiki.new(settings.gollum_path, wiki_options)
         @results     = wiki.pages


bartkamphorst:
> Without a testing environment I'm shooting blind here, however.

I also don't have a testing environment available. I seem to recall that I did make some changes to the way breadcrumbs were generated in `pages` but I don't remember if those changes would affect this issue. 😞

dometto:
Closing in favor of #1369

PR #1348, opened 2018-11-10
[WIP] Improve editor buttons: insert markup at the correct position. Resolves #504

### LINKED ISSUE #504: Fix markdown headings in editor

dometto:
Adds two keys to the language definitions for the editor function buttons:
* `break_line`: when true, then if...
  * there is no selection, then a newline is prefixed to the inserted text
  * there is a selection, then a newline is prefixed to AND added to the end of the inserted text
  * e.g. unordered lists and block quotes should break the line, whereas e.g. bold and italic should not
  * Edge case: if the markup is inserted at the beginning of the line, then no newline is prefixed
* `whole_line`: when true, then if there is no selection, then the markup is applied to the whole line
  * e.g. h1-h3 should be applied to the whole line, whereas e.g. unordered lists and bold should not

I've set `break_line` and `whole_line` to true where applicable, but only for the markdown language definition for now. I can add it to the other languages definitions if this general approach is approved (help appreciated in determining which buttons in which languages should be `break_line` and `whole_line`, respectively).

### DIFF
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
@@ -667,7 +667,35 @@
         // get the selected text from the textarea
         var editor = window.ace_editor;
         var txt = editor.getValue();
+        var rtPrefix = '';
+        var rtSuffix = '';
+        var breakBefore = false;
+        var breakAfter = false;
+        var selRange = editor.getSelectionRange();
         var selText = editor.getSelectedText();
+        if (selText === '' ) {
+          if (definitionObject.whole_line && definitionObject.whole_line == true) {
+            var line = selRange.start.row;
+            selText = editor.session.getLine(line);
+            var Range = ace.require('ace/range').Range;
+            selRange = new Range(line, 0, line, selText.length);
+          } else if (definitionObject.break_line && definitionObject.break_line == true) {
+            breakBefore = true;
+          }
+        } else if (definitionObject.break_line && definitionObject.break_line == true) {
+          breakBefore = true;
+          breakAfter  = true;
+        }
+
+        // Only prefix a newline to the replacement text if we're not already at the start of the line
+        if (breakBefore == true && selRange.start.column > 0) {
+          rtPrefix = '\n';
+        }
+
+        if (breakAfter == true) {
+          rtSuffix = '\n';
+        }
+
         var repText = selText;
         var reselect = true;
         var cursor = null;
@@ -697,7 +725,8 @@
         if ( definitionObject.replace &&
              typeof definitionObject.replace == 'string' ) {
           debug('Running replacement - using ' + definitionObject.replace);
-          var rt = definitionObject.replace;
+          var rt = definitionObject.replace
+          rt = rtPrefix + rt + rtSuffix;
 
           repText = escape( repText );
           repText = repText.replace( searchExp, rt );
@@ -732,8 +761,13 @@
         }
 
         if ( repText ) {
-          editor.session.replace(editor.selection.getRange(), repText);
+          editor.session.replace(selRange, repText);
           editor.focus();
+          if (breakAfter == true) {
+            var row = editor.getCursorPosition().row-1;
+            var column = editor.session.getLine(row).length;
+            editor.selection.moveTo(row, column)
+          }
         }
 
       },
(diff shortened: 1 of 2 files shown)


bartkamphorst:
Interesting PR. I'll pull it down sometime next week to inspect it in a live environment. 👍

bartkamphorst:
I think this is a great effort and I think the buttons behave much more intuitively now. I have noticed some strange behavior as well though, mostly with the lists, that do not behave in the same way:

* With no selection, cursor at the end of the line, the ordered list does not insert anything at all. The unordered list however inserts an asterisk on a new line and moves the cursor there.
* With a one-line selection, the ordered list inserts a number and a newline. The unordered list however inserts an asterisk and no newline (keeping the cursor on the current line).
* With a multi-line selection (say, for example, 3 lines), the ordered list inserts three numbers, one for each selected line. The cursor is placed on the fourth (empty) line. The unordered however only inserts one asterisk on the first line and keeps the cursor at the end of line 3.

Also, I noticed that the header buttons behave differently depending on whether there is any content after the current header content. That is, when I select the second line out of three, it inserts a header and adds a newline (effectively moving the third line down to line four). However, when I use the same button for the last line, in inserts the header but without the newline. Any idea why this is?

dometto:
Thanks for the feedback. The discrepancies between unordered and ordered lists are because of the very different ways those buttons have been implemented. Unordered lists use the generic substitution algorithm implemented in `gollum.editor.js`, whereas ordered lists have their own custom function, precisely in order to prefix the correct number (`1..n`) to each line, defined (for markdown) [here](https://github.com/gollum/gollum/pull/1348/files#diff-adbb9c024392145a437a0c912e6ae2d2R50). What do you think -- should we aim for more similar behavior, by writing a custom replacement function for the unordered lists, too?

> That is, when I select the second line out of three, it inserts a header and adds a newline (effectively moving the third line down to line four). However, when I use the same button for the last line, in inserts the header but without the newline. Any idea why this is?

Not directly, but I'm sure I can change it either way. Which behavior do you think makes more sense?

bartkamphorst:
> What do you think -- should we aim for more similar behavior, by writing a custom replacement function for the unordered lists, too?

Github has the same behavior for both ordered and unordered list, so I think we should strive to do the same.

> Not directly, but I'm sure I can change it either way. Which behavior do you think makes more sense?

I'm not sure! Having my cursor moved down feels a little obtrusive, but then again, most people will likely want to have whitespace between their header and their content. So I'm fine with the whitespace behavior for the last line as well.

dometto:
> Github has the same behavior for both ordered and unordered list, so I think we should strive to do the same.

Done 👍

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -367,6 +344,46 @@
...
+        };
+      }
+
+      window.ace_editor.commands.addCommand({
+        name: "header-1",
+        bindKey: {win: "Ctrl-1", mac: "Command-1"},
+        exec: hotkey('function-h1')
+      });
+      window.ace_editor.commands.addCommand({
+        name: "header-1",
```
bartkamphorst: Duplicate command.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -666,10 +683,29 @@
         // get the selected text from the textarea
         var editor = window.ace_editor;
         var txt = editor.getValue();
+        var breakBefore = false;
+        var breakAfter = false;
+        var selRange = editor.getSelectionRange();
         var selText = editor.getSelectedText();
+        var selTextLength = selText.length;
+        var wholeLine = false;
+        if (selText === '' ) {
```
bartkamphorst: Why is this `===` and not `==`?

lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb
```diff
@@ -696,133 +732,53 @@
         if ( definitionObject.replace &&
              typeof definitionObject.replace == 'string' ) {
           debug('Running replacement - using ' + definitionObject.replace);
-          var rt = definitionObject.replace;
+          var rt = definitionObject.replace
```
bartkamphorst: Missing semi-colon.

Issue #527, opened 2012-09-30
Auto Saving & Offline Support


bootstraponline:
As mentioned on #525.

It'd be easy to use HTML5 local storage to automatically save posts and provide offline support. Then if the browser crashed or the internet disconnected, no work would be lost.

bartkamphorst:
Here's a straightforward tutorial on [how to use localStorage](https://toddmotto.com/storing-data-in-the-browser-with-the-html5-local-storage-api/). @dometto I think this shouldn't be too difficult to implement. I'd like to discuss what the correct behavior should be though. Here's a first proposal:

* Creating or editing a page means storing page content data to localStorage. For keys, we could store page data with `'gollum-' + page.sha` as the key while editing an existing page, and `gollum-new-page` whilst creating a new page. 
* When saving the page, we should delete the page content from localStorage. 
* When hitting 'Cancel' the localStorage should also be emptied.
* When hitting the 'Back' button, the localStorage should persist.
* When creating or editing a page gollum should look if the browser has any stored data, and if so, overwrite the loaded data from the server with the stored data from the browser. This is the scenario where one is trying to recover from a crash or accidental closing of a tab.
* When a version conflict is detected, gollum should empty the localStorage, since a reload of the page will give a different `sha` and we don't want to keep increasing storage space.

Questions:
* Should gollum alert the user about recovering stored data or should it "just work"?
* How often should autosave run? Every second?

dometto:
On second thought, maybe it would be nice if there was an unobtrusive text, say just above or below the Save button (or wherever), 'Gollum just recovered your changes (undo)' (where clicking undo reloads the page's content)?

pedrocr:
A probably simple sub-feature would be to just have a way for hitting save in the edit view and have that just save but not go back to the normal view. That way I can hit save just like I do in my normal text-editor, to make sure the changes made it to disk, while continuing to edit.

mivok:
If anyone is interested in having something like this now, I have made the following in my custom.js file which implements autosaving using local storage. It probably needs some cleanup/style changes to be usable in gollum (e.g. this uses some newer JS features, and doesn't make use of jquery, which I believe gollum does, and I haven't made any attempt to make it work across multiple browsers), but it works well for me:

```js
//
// Autorecover/autosave functionality when editing
//
function autoRecover() {
    // Don't do anything if there isn't an editor
    if (!window.ace_editor) { return }

    const storageKey = `autorecover_${window.location}`;
[...]

PR #1349, opened 2018-11-16
Implement infrastructure for detecting and handling simultaneous edits

bartkamphorst:
This PR implements the infrastructure for detecting and handling simultaneous edits. It does this on the basis of a hash of the contents of the page. The hash is generated in the `get 'edit/*'` action and passed into the `gollum-editor-form` form as a hidden field named `etag`. When, after editing the page, the form is submitted via an ajax call to `post 'edit/*'` the original hash is compared to a new hash of the contents of the page before the new changes are applied. If the hashes match, this means that the page has not been modified since the user started editing. If the hashes do not match, the page has been changed in the meantime, and action has to be taken to resolve the version conflict. In the current implementation, a conflict results in a notification warning the user about the conflict. In a future implementation, this could be expanded to offer the user a user-friendly way of merging the two versions.

screen shot 2018-11-16 at 14 41 49

Caveats:
* This is not thread-safe. It is technically possible that while the two hashes are being compared for  request A, the page is updated by B and subsequently overwritten by A's changes.
* Even if this code was thread-safe, it cannot be guaranteed that the page is not updated via the CLI simultaneously.

Still, having this mechanism in place seems better than not having anything at all (the current situation). Comments welcome!

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -287,7 +287,8 @@ module Precious
         
         return if page.nil?
         if etag != page.sha
-          halt 412, 'For future use: some helpful data for resolving the conflict here.'
+          # Signal edit collision and return the page's most recent version
+          halt 412, {etag: page.sha, text_data: page.text_data}.to_json
         end
         
         committer = Gollum::Committer.new(wiki, commit_message)
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -176,6 +176,7 @@ module Precious
               @page         = page
               @page.version = wiki.repo.log(wiki.ref, @page.path).first
               @content      = page.text_data
+              @etag       = Digest::SHA1.hexdigest(@content)
               mustache :edit
           else
             redirect_to("/create/#{encodeURIComponent(@name)}")
@@ -278,11 +279,17 @@ module Precious
       end
 
       post '/edit/*' do
+        etag = params[:etag]        
         path      = "/#{clean_url(sanitize_empty_params(params[:path]))}"
         page_name = CGI.unescape(params[:page])
         wiki      = wiki_new
         page      = wiki.paged(page_name, path, exact = true)
+        
         return if page.nil?
+        if etag != Digest::SHA1.hexdigest(page.text_data) 
+          halt 412, 'For future use: some helpful data for resolving the conflict here.'
+        end
+        
         committer = Gollum::Committer.new(wiki, commit_message)
         commit    = { :committer => committer }
 
@@ -292,7 +299,6 @@ module Precious
         update_wiki_page(wiki, page.sidebar, params[:sidebar], commit) if params[:sidebar]
         committer.commit
 
-        redirect to("/#{page.escaped_url_path}") unless page.nil?
       end
 
 
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -176,7 +176,7 @@ module Precious
               @page         = page
               @page.version = wiki.repo.log(wiki.ref, @page.path).first
               @content      = page.text_data
-              @etag       = Digest::SHA1.hexdigest(@content)
+              @etag         = page.sha
               mustache :edit
           else
             redirect_to("/create/#{encodeURIComponent(@name)}")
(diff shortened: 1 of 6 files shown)


bartkamphorst:
Tests are still failing; will investigate why at a later time.

lib/gollum/app.rb
```diff
@@ -176,6 +176,7 @@ class App < Sinatra::Base
               @page         = page
               @page.version = wiki.repo.log(wiki.ref, @page.path).first
               @content      = page.text_data
+              @etag       = Digest::SHA1.hexdigest(@content)
```
dometto: Here you can also do `page.sha`

dometto:
Looks nice and clean to me! Awesome work.

One suggestion in the reviews, but maybe the same should be implemented for the `post create` route? Also, a test case would be nice!

> This is not thread-safe. It is technically possible that while the two hashes are being compared for request A, the page is updated by B and subsequently overwritten by A's changes.

:-( Not worth pursuing that though, given that git CLI access is always possible. Perhaps implementing [this](https://github.com/gollum/gollum/issues/527) could provide some additional insurance against data being lost.

bartkamphorst:
Thanks! 😄 

> maybe the same should be implemented for the post create route? 

Non-existing pages don't have any content to hash yet, right? So I don't see that happening in the same way.

> Also, a test case would be nice!

Agreed, I should fix the existing tests and add a test for conflicting versions.

dometto:
> Non-existing pages don't have any content to hash yet, right? So I don't see that happening in the same way.

Right, so the collision-detection shouldn't use hashes. 👍 I see we have some kind of collision detection for new pages already: https://github.com/gollum/gollum/blob/5.x/lib/gollum/app.rb#L358

Are we happy with throwing the mustache error there, or do we prefer a `halt` (412?)?

bartkamphorst:
> I see we have some kind of collision detection for new pages already

Ah, yes, that seems sufficient for now. 

> Are we happy with throwing the mustache error there, or do we prefer a halt (412?)?

There are no callbacks waiting to do something with the error, so I prefer the mustache error at the moment.

lib/gollum/app.rb
```diff
@@ -278,11 +279,17 @@ class App < Sinatra::Base
...
 
       post '/edit/*' do
+        etag = params[:etag]        
         path      = "/#{clean_url(sanitize_empty_params(params[:path]))}"
         page_name = CGI.unescape(params[:page])
         wiki      = wiki_new
         page      = wiki.paged(page_name, path, exact = true)
+        
         return if page.nil?
+        if etag != Digest::SHA1.hexdigest(page.text_data)
```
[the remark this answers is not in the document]
bartkamphorst: Will fix.

lib/gollum/app.rb
bartkamphorst: Will fix.

bartkamphorst:
In the latest commits, I've fixed the tests and have added a test for edit collision. In addition, the `post 'edit/*'` action now returns the latest version of the page's content together with the latest `sha` as a json string when a collision is detected (so with the `halt 412` call). This will make it easy in the future to take the latest version and present the user with the option of manually merging the two colliding versions.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -367,6 +367,34 @@ $(document).ready(function() {
...
+        type: 'POST',
+        data: formData,
+        processData: false,
+        contentType: false,
+        success: function(data) {
+          window.location = window.location.href.replace(/gollum\/edit\//, '')
+        },
+        error: function(data, textStatus, errorThrown) {
+          if (data.status == 412) {
+            $('#gollum-editor-submit').attr('disabled', false)
```
dometto: What's the case where a 412 is returend but the submissions button is disabled?

dometto:
Good stuff! Looks ready to merge, just one question of understanding in the comments.

> Ideally, we could use ace-diff (demo) to let the user sort out the conflict.

All for it!

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: When the page is originally submitted, the submit button is set to disabled so it cannot be clicked twice (which would cause another, parallel, ajax post request).
```javascript
    $("#gollum-editor-submit").click( function(e) {
      e.preventDefault();
      // Prevent button from being clicked again
      $(this).attr('disabled', true);
```
 This is necessary because with the ajax submit we are not necessarily redirected upon submit as was the case before. When the edit is unsuccessful due to an edit conflict however, we want the submit button to become responsive again, hence line 389. Does that make sense?

dometto: 👍

PR #1350, opened 2018-11-18
Improve RTL support with updated ACE. Resolves #1314

### LINKED ISSUE #1314: Finishing RTL support
Hello,

could you, please, finish RTL support, so it works perfectly, not just very good?...
For now Gollum (on @GitLab) seems to be the best solution for supporting RTL languages.

Gollum automatically recognizes RTL texts and renders wiki pages properly, aligning them to the right!
See example here:

https://gitlab.com/pninim.org/pninim/wikis/test.org

Update: It looks like Markdown renders correctly (right aligned), but RDoc and AsciiDoc - do not.

However when you start editing an RTL text (after pressing the "Edit" button) it get's aligned to the left. This can be solved in Firefox by pressing the "Ctrl-Shift-x", thus forcing RTL manually. Gollum is the only app that doesn't block it - thank you!

So in order to make the RTL support perfect, could you, please enable it also in the editor, by:
1. Automatically recognizing an RTL language, the way you do during rendering.
2. If an RTL language is detected, imitate the Firefox's "Ctrl-Shift-x" behavior.

Thank you!

dometto:
ace has improved its support for RTL in https://github.com/ajaxorg/ace/pull/3658/

Updating ace doesn't seem to break anything. The user can switch modes on a per-line or per-selection basis with `Cmd-alt-shift-R` or `Cmd-alt-shift-L` (`Ctrl` instead of `Cmd` on Windows). This PR enables support for this.

I've also set `dir="auto"` for the content divs (including the sidebars). This HTML5 attribute determines the text's direction (for the whole div) based on the language (or rather, script) of the first character. To support different scripts in one page, I've added some JS that sets the `dir="auto"` attribute for block level elements (`p` and `span`) in the page's content. Again, the browser will determine the direction of the text in a `p` or `span` based on the first character.

Example result:
screen shot 2018-11-18 at 20 25 40

Slight complication: ace sets the direction on a per-line basis using [invisible unicode characters](https://en.wikipedia.org/wiki/Right-to-left_mark), but even though those are saved in the page, the browser doesn't appear to respect them (instead, it only follows the `dir` attributes). So when editing with ace, you can set a line that uses latin characters to RTL, but on the rendered page the latin characters will be displayed as LTR. I have no idea how to get the browser to respect the unicode characters, even though it appears that ought to be supported in HTML5. But maybe this functionality is good enough.

- [x] Update release notes
- [x] Add a button to the editor to switch direction, or clearly document the keyboard shortcut for this
- [x] Check with an actual RTL language user whether everything works as expected 
- [x] File bug with ace: inserting a newline in an RTL line mistakenly starts a new LTR line.


dometto:
@AvtechScientific care to weigh in on this?

dometto:
Another approach could be to let the user set a single direction for the page in the `yaml` frontmatter (but then Latin lines in e.g. a Hebrew document would also be displayed RTL).

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -398,6 +398,8 @@ $(document).ready(function() {
     $('span.critic.comment').filter(function() {return $(this).text()!="";}).before('&#8225;');
   }
 
+  $('.markdown-body p, .markdown-body span').attr('dir','auto');
```
bartkamphorst: What's the reason for doing this with JS and not in de template?

bartkamphorst: I don't mind it, just wondering.

dometto: These `p` and `span` tags are the one generated by the rendering gem, so they're not there yet in the template.

dometto:
> Files changed: 565. 🤣

I'm waiting for a comprehensive review on all of them!

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: 👍

dometto: I think `p` and `span` covers it, or do you know of other block-level elements generated by the markup?

dometto: Right, `pre` should probably be on the list, as should `table`. 👍

If I understand correctly, `dir` is a global attribute that *can* be added to any element, including `ul`, `h`, etc. But adding it to every tag seems to be [not recommended](https://www.w3.org/International/questions/qa-html-dir#markkupcrazy). The question is what's reasonable. Is it it enough to give the user the option of *changing* direction on the level of `p`, `span`, etc., or do we want to add the tags you meant?

dometto: I should add that I think it is ultimately the responsibility of markup specifications to support RTL in a sane way. There's some talk of adding RTL indication to markdown, but I don't think it exists yet: https://talk.commonmark.org/t/explicit-rtl-indication-in-pure-markdown/286
Again, the question is what's reasonable for us to do when it comes to switching direction (if you use only an RTL script on the page, that should work fine with the current implementation)

bartkamphorst: >  adding it to every tag seems to be not recommended.

That makes sense! My intuition says that users or RTL would want to have their entire page, including the headers, to be RTL. Changing direction mid-page and only for certain elements seems niche-behavior to me. Just adding `pre` and `table` should be fine. 👍

AvtechScientific:
The "Example result" screenshot from the first comment looks fine. Thank you!

dometto:
Thanks @AvtechScientific. Does the idea of checking text direction at the level of each `p`, `pre`, `span`, and `table` seem sufficient to you? Also, if you'd like to take the new functionality for a testdrive, drop me an e-mail and I can send you the link to a public test instance.

AvtechScientific:
@dometto , I don't have enough knowledge of html/css to understand exactly what you mean, but I can try to check the test instance. Mail: contact [at] <myGitHubusername> . com

bartkamphorst:
> The remaining bug in ACE has been solved

Nice! 👍

bartkamphorst:
Sorry, hit the wrong button. 😨

bartkamphorst:
>  I say we wait until the next bugfix release of ACE before we merge this, to limit the huge file change commits.

@dometto Update: Ace [v.1.4.3](https://github.com/ajaxorg/ace/releases/tag/v1.4.3) is available.

dometto:
Thanks!

dometto:
@bartkamphorst what do you think, should we just commit the newest ACE release, or rather use a submodule?

dometto:
@bartkamphorst I have introduced a mini button that reverses the direction of the selected text. I've placed it to the right of the help button. Right now, it's styled as the "Bold" minibutton. I was thinking we might give it a more appropriate styling in your Primer PR, or after that's merged. Does that make sense?

By the way, would merging this PR create merge difficulties for your Primer work?

bartkamphorst:
> Right now, it's styled as the "Bold" minibutton. I was thinking we might give it a more appropriate styling in your Primer PR

That seems fine. 👍 

> By the way, would merging this PR create merge difficulties for your Primer work?

Can't really tell, probably. 🤷‍♂

dometto:
Slight change of plans after updating to ace-1.4.4: this release supports changing the direction of the editor *globally*. I thought that might be a better fit for the minibutton. So now you can:

* Change the text direction of the whole editor with the button.
* Change the direction of a single line with the keyboard combination (`Cmd-alt-shift-L/R`).

That sound good?

dometto:
Added bonus: on page load, the editor will check if its content is in an RTL language, and if so, switch to RTL mode by default.

bartkamphorst:
Yup, sounds good!

dometto:
Fun final bug that needs to be solved: the minibuttons aren't inserting text at the right places when a line is RTL. However, only the *first* and *last* words on a line seem to be effected, while text replacement for the words in the middle is working correctly.

bartkamphorst:
> the minibuttons aren't inserting text at the right places when a line is RTL.

Ah... Well, at least you caught it now... Happy hunting. 😉

dometto:
May have stumbled upon an ACE bug:
https://github.com/ajaxorg/ace/issues/3957

dometto:
Turns out this was not really a bug, but was a result of the combination of English characters on an RTL line. Text insertion on RTL lines works correctly for characters in an RTL language. @bartkamphorst shall I merge this then?

bartkamphorst:
Good news! Yes, go ahead and merge! 🎆

Issue #1303, opened 2018-04-15
open page when exact match in search 


deadc0de6:
Would it be possible that if a search term matches exactly an existing page in the wiki, gollum would open that page? Let's say for example there's a page named "test", a search with `test` should open the page `[[test]]`.

dometto:
@deadc0de6 happy to merge this into gollum. Would you like to try your hand at a PR?

acaporrini:
@dometto I made a PR for this:
#1352

deadc0de6:
@acaporrini nice, thanks! Sorry but I'm not good at ruby :-s

dometto:
We've decided not to include this feature, because the behavior is not always as intuitive as it might seem. See #1352.

PR #1352, opened 2018-11-21
On search redirect to result if exact match

acaporrini:


bartkamphorst:
Thanks for your PR @acaporrini . I understand the approach you've taken here, but computationally it seems ineffective to do the full search, sort the results, reverse the results, and then dig around in the results for an exact match (which may not even be there). It would make more sense for `wiki.search` (defined in `gollum-lib`) to stop searching when an exact match is found and to return that directly, wouldn't you agree? Any additional thoughts on this, @dometto ?

bartkamphorst:
Ping @dometto .

dometto:
I think the reasoning is that we should only open the page if there are no other results, so that the user never "misses" any results. That seems ok to me. PR would have to be re-submitted against the `5.x` branch, though, and there are some issues I'll mark in the line comments.

bartkamphorst:
> I think gollum should only open the page when there are no other results (say the user just wants to search for foo, but it so happens someone has created a page called foo.md). 

If you think this, then I suspect that you do not actually agree with my point about stopping the search on an exact match. 😉 

After some consideration I now agree with you that, contra https://github.com/gollum/gollum/issues/1303, the automatic redirect should only happen if there is only one result. Because searching for `foo` while having both `foo.md` and `foo.txt` should show both results and not redirect to the first one found. One question that I have about this way is whether it is worth the trouble. Because in order to only get one result back, you have to be quite specific in your search (in the test repo for example, searching for `Hobbit` will yield 4 results while `Hobbit.md` will yield 1).

dometto:
Haha, I guess my agreement was conditional on the premise that we want to open the first hit. :)

But I actually think your new point is valid, and not just conditionally. It may not be worth the trouble, and it could be quite surprising if, in the rare case that there is only one match, the user suddenly gets redirected to the page.

I guess in a really ideal world, we could have a sort of launcher/searchbar hybrid, like Spotlight on OS X. But that would be nontrivial, and not very necessary to have.

bartkamphorst:
> But I actually think your new point is valid, and not just conditionally. It may not be worth the trouble, and it could be quite surprising if, in the rare case that there is only one match, the user suddenly gets redirected to the page.

So... Closing this then @dometto ?

dometto:
Yes indeed. My apologies, @acaporrini, I hope you understand our reasoning. If you'd like to work on gollum's search function, there is a ticket on [showing more verbose results](https://github.com/gollum/gollum-lib/issues?q=is%3Aissue+is%3Aopen+label%3Aenhancement) that we're definitely interested in getting into `5.x`.

dometto:
Also https://github.com/gollum/gollum/issues/615

PR #1353, opened 2018-11-27
Do not attempt to normalize uploaded files unless they have valid page extensions.

bartkamphorst:
Fairly straightforward fix since the relevant code will live in `gollum-lib`. Basically, the proposal is to add to class methods to `Gollum::File`, namely `image?(filename)` and `binary?(filename)`. Thoughts?

NB. This PR should not be merged until `gollum-lib-5.x` implements these methods.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -227,9 +227,11 @@ module Precious
           options.merge! author
         end
 
+        normalize = !(Gollum::File.image?(fullname) || Gollum::File.binary?(fullname))
+
         begin
           committer = Gollum::Committer.new(wiki, options)
-          committer.add_to_index(dir, filename, format, contents)
+          committer.add_to_index(dir, filename, format, contents, {normalize: normalize})
           committer.after_commit do |committer, sha|
             wiki.clear_cache
             committer.update_working_dir(dir, filename, format)
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -227,7 +227,7 @@ module Precious
           options.merge! author
         end
 
-        normalize = !(Gollum::File.image?(fullname) || Gollum::File.binary?(fullname))
+        normalize = Gollum::Page.valid_extension?(fullname)
 
         begin
           committer = Gollum::Committer.new(wiki, options)


dometto:
Sounds good. Is the idea that `image?` will just be a file extension check? If so, we should use the same logic for that as in the tag filter [here](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/filter/tags.rb#L79) and [here](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/filter/tags.rb#L87).

lib/gollum/app.rb
```diff
@@ -227,9 +227,11 @@ class App < Sinatra::Base
           options.merge! author
         end
 
+        normalize = !(Gollum::File.image?(fullname) || Gollum::File.binary?(fullname))
```
dometto: You could dispense with the new `normalize` variable and just add the `!(...)` to the hash below, but if you think this is more readable, 👍

bartkamphorst:
> Sounds good. Is the idea that image? will just be a file extension check?

That was the plan, but I actually like the use of mime types better. 😄 Regardless, you are right that the class method could then be reused throughout the code base. 👍

lib/gollum/app.rb
bartkamphorst: Okay, in that case I'd like to keep it as is for readability. 👍

bartkamphorst:
@dometto What would be the desired behavior when the mime-type cannot be determined based on file extension? For example, a file without an extension (e.g., `file`) or a file with an unknown extension (e.g., `file.foo` but also `file.wiki` or `file.rest`)? Options:

1. Always normalize with the risk of breaking the file;
1. Never normalize with the risk of causing strange behavior further down the pipeline;
1. Attempt to determine if the file is binary by reading a chunk of it, and skipping normalization if it seems to be a binary file.

I'm leaning towards option 3, but curious to hear where you stand on this.

dometto:
> Attempt to determine if the file is binary by reading a chunk of it, and skipping normalization if it seems to be a binary file.

I think that would be entirely acceptable, assuming that there is an actual risk involved in *not* normalizing (?) -- the app legitimately uses mime and/or file extension to determine behavior, it's the user's responsibility to set those correctly.

bartkamphorst:
> I think that would be entirely acceptable [...] the app legitimately uses mime and/or file extension to determine behavior, it's the user's responsibility to set those correctly.

Thanks. I've been playing around with the `ptools` heuristic for binary content and I'm not very confident about it. For example, some people on the forums report that it will erroneously judge that utf-8 content is binary, and in one of my own tests the heuristic judged a PDF document without extension **not** to be a binary file. Given this behavior, I would rather just rely on mime type (which relies on file extensions). 

> assuming that there is an actual risk involved in not normalizing (?) 

I don't really know what the risk is to be honest. I assume that normalization is there in `add_to_index` because `\r` could break page rendering, but I'm not sure about regular files. We could also reverse the policy and do the following:

    Exempt uploaded files from normalization, unless the mime type of the file is text. 

This still has some caveats as not all mime-types are known. For example, `.rest` is not recognized as text (but `.rst` is) nor is `.pod`. To deal with this we could add a second check to see if the file extension is one of the wiki formats that gollum supports. @dometto What are you thoughts about this approach?

dometto:
In a hurry, but is it an option to just do the normalization in the filter chain?

bartkamphorst:
> we can normalize only if the filename is recognized as a page? 

I had the same thought and I think this approach should suffice. 👍 See https://github.com/gollum/gollum/pull/1353/commits/ddfaf8c0915371ccd01292948e37d4277999a9c0.

> Downside: if a file doesn't have a valid page extension and is then renamed to have a valid page extension, it won't be normalized.

True, but it will be normalized when the file is edited in gollum, which should be good enough, right?

dometto:
> True, but it will be normalized when the file is edited in gollum, which should be good enough, right?

Ah, yes! 👍

bartkamphorst:
Okay with this latest approach then?

dometto:
Go for it!

PR #1358, opened 2018-12-28
Make mathjax.config.js the standard filename. Fixes #1340.

### LINKED ISSUE #1340: 5.x: mathjax config not among protected files
Just realized the mathjax js config file is not among the newly protected routes introduced by [this PR](https://github.com/gollum/gollum/pull/1324). However, at the time the protecting routes are being defined, we don't have access to the wiki options yet, so we can't fetch the location of the mathjax config file. My proposal is to do the same thing we're doing with JS and CSS: don't let the user determine the filename, but let it always be e.g. `mathjax.config.js` (?), and block that from being edited.

> However, at the time the protecting routes are being defined, we don't have access to the wiki options yet

For the same reason, we should clearly document that these routes are *always* blocked -- whether you're using `--css`, `--js`, or -`--mathjax-config`. Is that a reason to prefix the filenames, so that we get e.g. `gollum_custom.js`, etc.?

bartkamphorst:


dometto:
Maybe add to the test in `test_app` around line 508?

bartkamphorst:
> Maybe add to the test in test_app around line 508?

Added tests; subsequently fixed the issue they demonstrated. 😉 @dometto PR approved?

dometto:
👍

Issue #920, opened 2014-12-18
Default header layout is weird


DannyBen:
With any page that has a title longer than "Home", the header looks very wrong. As if someone forgot a CSS.

Example:
![gollumn](https://cloud.githubusercontent.com/assets/2405099/5488516/b37a5f34-86c7-11e4-8874-7d8df98fbc5c.png)

Caption aligned left, right below it with no visual separation or "air", a long, right aligned button bar, and below it starts the actual document.

I was expecting to either see the button bar first and the title below it.
Better yet, I think all these buttons could be collapsed into a single 3-horizontal-lines button, and be placed on the right together with the search box.

Is there any setting I can change to make it less messy?

aldanor:
So this is paused until the big CSS change is merged in?

dometto:
Pretty much, yeah :)

SkyCrawl:
> I was expecting to either see the button bar first and the title below it.
> Perhaps this could be incorporated as a part of #917?
@DannyBen @techwiz24 @aldanor 
With this particular problem, you don't need to wait until the CSS changes are done :). If you follow issue references from #917 , you will find [this issue](https://github.com/gollum/gollum/issues/906), where you can open my css file addressing the header issue. If you use with the `--css` wiki option, it will push the header below the button bar and add some space. I might also center the header but if it does not, applying `text-align: center;` should do the trick.

aldanor:
Thanks @SkyCrawl! These sort of fixes should really be merged in the master :)

/* btw: another little css bug I spotted: previewing a page with sidebar makes the sidebar float in a weird location; I don't know if this one is known */

dometto:
> These sort of fixes should really be merged in the master :)

I won't have much time to work on gollum myself again until april, but PR's are very welcome! :)

PR #1362, opened 2018-12-30
Move page title (h1) down. Move Home button to upper left corner on all views.

bartkamphorst:
This PR fixes issue https://github.com/gollum/gollum/issues/920. In addition, it moves the Home button to the upper left corner for consistency between views.

screen shot 2018-12-30 at 19 09 03


dometto:
Looks nice! Curious how this will play out when a customer _Header is present.

bartkamphorst:
(BTW, ignore the fact that the Home button is not in the top left corner on this screenshot.)

PR #1363, opened 2019-01-03
Implements a breadcrumb for normal wiki pages. Resolves #629.

### LINKED ISSUE #629: Breadcrumb trail
Add a new feature to show a breadcrumb trail for all the links clicked

bartkamphorst:
Notice that this breadcrumb trail will currently include `page-file-dir` as an element when it is enabled. However, since `page-fir-dir` [is set to become internal](https://github.com/gollum/gollum/issues/1052), this should resolve itself once https://github.com/gollum/gollum/issues/1052 is implemented.

### DIFF
--- a/lib/gollum/public/gollum/stylesheets/gollum.scss
+++ b/lib/gollum/public/gollum/stylesheets/gollum.scss
@@ -34,6 +34,7 @@ a {
   margin: 1em 0 0;
   padding: 0;
   overflow: hidden;
+  width: 100%;
 
   @include desktop-breakpoint {
     border-bottom: $border-standard;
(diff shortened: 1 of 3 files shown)


lib/gollum/views/page.rb
```diff
@@ -18,12 +18,23 @@ class Page < Layout
...
         h1 = @h1_title ? page_header_from_content(@content) : false
-        h1 || @page.url_path_title
+        h1 || @page.title
       end
 
       def page_header
         title
       end
+      
+      def breadcrumb
```
dometto: Nice and clean implementation!

lib/gollum/public/gollum/stylesheets/gollum.scss
```diff
@@ -46,14 +47,24 @@ a {
     float: left;
     line-height: normal;
     margin: 0;
-    margin-top: 2em;
+    margin-top: 3em;
```
dometto: This affects the margins for all h1's, right? Have you tested how that looks, and would it be an alternative to add the extra spacing only to the first h1?

bartkamphorst: > This affects the margins for all h1's, right?

No, only the h1 within the head div.

lib/gollum/views/page.rb
bartkamphorst: Thanks. :)

PR #1368, opened 2019-02-12
Correct matching of '/pages' and '/compare'.

CodeFloh:
So far, these keywords were matched anywhere in the path. This means that folders with these names are not possible anywhere, and if they exist, they lead to unexpected behavior.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -435,7 +435,7 @@ module Precious
     end
 
     get %r{
-      /compare/ # match any URL beginning with /compare/
+      ^/compare/ # match any URL beginning with /compare/
       (.+)      # extract the full path (including any directories)
       /         # match the final slash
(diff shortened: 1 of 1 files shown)


dometto:
Thanks @CodeFloh! However, in the 5.x branch we've introduced a /gollum namespace for all of the app's internal routes. Would you mind checking if the same problem is still present there, and resubmit a fix against the 5.x branch if necessary?

CodeFloh:
I tried out the 5.x branch. In the use cases that I tried the problem is no longer there. However, there will always be filenames that can match keywords. I would actually suggest using two separate paths, e.g.: `http://localhost:4567/wiki/Home` etc. for content viewing and `http://...:4567/gollum/...` for other actions. When `-b /wiki` is used, the prefix sadly also switches to `/wiki/gollum`. I'm not sure, yet, if I will make such a change in my fork.

PR #1372, opened 2019-03-15
Fix h1-title at org-mode

hirocaster:
h1-title cannot fetched H1 tag at org-mode.
It fixed.

## before

`H1 Title1` should be the page title.

before

## after

`H1 Title1` is the page title.

after

##  content

This is page content.

content


bartkamphorst:
@hirocaster Looks sensible, thanks for the PR. Would you mind reopening this against the `5.x` branch though? Thanks!

hirocaster:
OK #1373

PR #1384, opened 2019-05-16
Don't lookup page's last version on edit

dometto:
Slow and unused (?).

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -177,7 +177,6 @@ module Precious
         @allow_uploads = wiki.allow_uploads
           if page = wikip.page
               @page         = page
-              @page.version = wiki.repo.log(wiki.ref, @page.path).first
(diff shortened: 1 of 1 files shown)


bartkamphorst:
IIRC this was implemented in an effort to avoid a call to `#versions`. Are you sure that this call is the culprit for the slowdown?

dometto:
> Are you sure that this call is the culprit for the slowdown?

Yes, removing the line resolves the crash when using the rugged adapter.

> IIRC this was implemented in an effort to avoid a call to #versions.

According to the Blame information, the line was added 7 years ago: https://github.com/gollum/gollum/commit/e1517c29074329fcd473990708cdae73f14f9473

Or is that implementing what you think?

bartkamphorst:
No I must have been thinking of something else. I'm okay with this PR then!

PR #1387, opened 2019-05-17
Fix breadcrumbs. Resolves #1383

### LINKED ISSUE #1383: 5.x: no breadcrumbs on page
Somewhere during the recent updates the newly implemented breadcrumbs on pages (Page, not `/overview`) stopped working. Could possibly be related to changes in how paths are set / treated.

dometto:
Also added a test

### DIFF
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -26,14 +26,14 @@ module Precious
       end
       
       def breadcrumb
-        path = Pathname.new(@page.url_path_title)
+        path = Pathname.new(@page.url_path).parent
+        return '' if path.to_s == '.'
         breadcrumb = []
         path.descend do |crumb|
           element = "#{crumb.basename}"
-          next if element == @page.title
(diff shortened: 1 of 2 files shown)


lib/gollum/views/page.rb
```diff
@@ -26,14 +26,14 @@ def page_header
       end
       
       def breadcrumb
-        path = Pathname.new(@page.url_path_title)
+        path = Pathname.new(@page.url_path).parent
```
dometto: This is the fix proper (using `url_path` rather than `url_path_title`), the rest is some mild refactoring, taking advantage of `Pathname#parent` to make sure the page's title doesn't end up in the breadcrumb.

dometto:
Just to make sure I've reintroduced the correct behavior, this is what it looks like now (with the page title below the crumb):

Screenshot 2019-05-17 at 13 13 44

bartkamphorst:
> Just to make sure I've reintroduced the correct behavior, this is what it looks like now (with the page title below the crumb):

Yes, that's what's expected. Thanks for the quick PR!

PR #1389, opened 2019-07-01
seamless omnigollum support

jamincollins:
Add login indication div that is only displayed if omnigollum is
enabled.

Signed-off-by: Jamin W. Collins <jamin.collins@gmail.com>

### DIFF
--- a/lib/gollum/views/layout.rb
+++ b/lib/gollum/views/layout.rb
@@ -40,6 +40,22 @@ module Precious
         @js
       end
 
+      # Passthrough additional omniauth parameters for status bar
+      def user_authed
+        @user_authed
+      end
+
+      def user_provider
+        @user.provider
+      end
+
+      def user_name
+        @user.name
+      end
+
+      def omnigollum_enabled
+        @omnigollum_enabled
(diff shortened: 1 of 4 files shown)


lib/gollum/views/layout.rb
```diff
@@ -40,6 +40,22 @@ def js # custom js
...
+      def user_provider
+        @user.provider
+      end
+
+      def user_name
+        @user.name
+      end
+
+      def omnigollum_enabled
+        @omnigollum_enabled
```
dometto: Where is this variable, as well as the ones above, being set? In omnigollum?

jamincollins: Yes, they are set in omingollum.

dometto:
Sorry for the late reply @jamincollins. I think it's a bit strange to add references to variables and functionality that belong to a different project that's not even under the same org. In 5.x, you can easily register new/different templates and routes. So I think the best way to go would be to include these changes in omnigollum itself. I'll close this but will be happy to provide help if you'd like to implement something along those lines.

PR #1393, opened 2019-08-13
Major CSS/styling overhaul with Primer.css.

bartkamphorst:
Sometimes images say more than words.

Screen Shot 2019-08-13 at 22 05 11

and

Screen Shot 2019-08-13 at 22 15 07

and

Screen Shot 2019-08-13 at 22 17 08

and

Screen Shot 2019-08-13 at 22 16 48


bartkamphorst:
There appears to be an issue with running the tests on travis, but the test suite runs without failures locally. 👍 

Screen Shot 2019-08-14 at 13 59 21

dometto:
I committed the Travis fix to master, and not 5.x:
https://github.com/gollum/gollum/pull/1392
You could push the same changes to `.travis.yml` to fix the tests here :)

lib/gollum/templates/navbar.mustache
```diff
@@ -0,0 +1,43 @@
+<nav class="actions pt-4">
+	
+	<div class="TableObject">
+		<div class="TableObject-item">
+	    <a class="btn" href="{{base_url}}/">Home</a>
+	  </div>
+
+
+  	<div class="TableObject-item TableObject-item--primary px-2" {{^search}}style="visibility:hidden"{{/search}}>
```
dometto: Perhaps just `hidden` would be enough, to avoid inline CSS?

lib/gollum/templates/overview.mustache
```diff
@@ -0,0 +1,42 @@
+<div id="wiki-wrapper" class="results">
+<div id="head" style="border:none;">
```
dometto: Prefer a non-inline CSS option to keep the templates clean, but no dealbreaker if it's too complicated!

lib/gollum/templates/page.mustache
```diff
@@ -6,93 +6,15 @@ Mousetrap.bind(['e'], function( e ) {
...
-  {{/has_sidebar}}
-  <div id="wiki-body" class="gollum-{{format}}-content">
-    {{#has_header}}
-    <div id="wiki-header" class="gollum-{{header_format}}-content">
-      <div id="header-content" class="markdown-body" dir="auto">
-        {{{header_content}}}
-      </div>
-    </div>
-    {{/has_header}}
-    <div class="markdown-body{{#header_enum?}} header-enum{{/header_enum?}}" {{#header_enum?}}style="--header-enum-style:{{header_enum_style}};"{{/header_enum?}} dir="auto">
```
dometto: Same comment about inline CSS

lib/gollum/templates/wiki_content.mustache
```diff
@@ -0,0 +1,52 @@
...
+	  {{/has_sidebar}}
+	  <div id="wiki-body" class="gollum-{{format}}-content">
+	    {{#has_header}}
+	    <div id="wiki-header" class="gollum-{{header_format}}-content">
+	      <div id="header-content" class="markdown-body">
+	        {{{header_content}}}
+	      </div>
+	    </div>
+	    {{/has_header}}
+	    <div class="markdown-body {{#header_enum?}}header-enum{{/header_enum?}}" {{#header_enum?}}style="--header-enum-style:{{header_enum_style}};"{{/header_enum?}}>
```
dometto: Inline CSS

lib/gollum/templates/page.mustache
```diff
@@ -6,93 +6,15 @@ Mousetrap.bind(['e'], function( e ) {
...
-  <div id="wiki-footer" class="gollum-{{footer_format}}-content">
-    <div id="footer-content" class="markdown-body" dir="auto">
-      {{{footer_content}}}
-    </div>
-  </div>
-  {{/has_footer}}
-  </div>
 
-</div>
-<div id="footer">
```
dometto: We could
1. wrap the footer in `{{^preview}}{{/preview}}`
2. add a `preview` method to the Preview view (`preview.rb`) that always returns true.
to avoid rendering the footer on preview pages.
Could also implement that solution in a new PR

dometto:
Excellent work! Great improvements to the app 🔥

lib/gollum/templates/page.mustache
bartkamphorst: There is no `preview.rb` view, only `page`. So we would have to differentiate within the `page.rb` view, but I suppose we could just check if the Page object is a PreviewPage, correct?

bartkamphorst:
Thanks. I'll see what I can do about the inline CSS.

lib/gollum/templates/page.mustache
dometto: Ah, good point. Or just add an `attr_reader :preview` to the Page view, and only set it to `true` in the `/preview` route?

lib/gollum/templates/navbar.mustache
bartkamphorst: Those do not appear to be equivalents, or at least it gives different results. I think someone with more know-how could tackle this one.

lib/gollum/templates/page.mustache
bartkamphorst: This was already in place.

lib/gollum/templates/page.mustache
bartkamphorst: I'll see what I can do!

lib/gollum/templates/overview.mustache
bartkamphorst: Fixed.

lib/gollum/templates/wiki_content.mustache
bartkamphorst: This was also already in place. Not sure how to deal with it in a better way quickly, so I think we should leave this for now.

lib/gollum/templates/page.mustache
bartkamphorst: > just add an attr_reader :preview to the Page view, and only set it to true in the /preview route

Works like a charm.

lib/gollum/templates/wiki_content.mustache
```diff
@@ -40,6 +40,7 @@
 
 	</div>
 
+	{{^preview}}
```
dometto: Cool! I should have asked before: will this result in a lack of whitespace at the bottom of the preview page? If so we should make the *contents* of the footer div, not the div itself, conditional on `@preview`. Sorry for the lateness

bartkamphorst: Good point. Fixed it in latest commit.

dometto:
🎆 🎉 🎈

dometto:
@bartkamphorst could you have a look which CSS-related issues have been resolved, and which still need to be open?
https://github.com/gollum/gollum/issues?q=is%3Aissue+is%3Aopen+label%3A%22CSS+styling%22

bartkamphorst:
> could you have a look which CSS-related issues have been resolved, and which still need to be open?

Done, @dometto .

PR #1395, opened 2019-08-18
Extract page dir from latest changes. Fix #1388

### LINKED ISSUE #1388: 5.x: Latest changes still shows page-file-dir
In `5.x` the page-file-dir should remain hidden. However, the `/latest_changes` view still shows it in its links.

dometto:
As per the title. I also got rid of the method that was extracting renamed file paths of the following form:

'/path1 -> /path2'

...as this only works on grit. On rugged and rjgit, renamed old_path and new_path are just added separately to the array containing paths that the commit in question touches. This means there is no visible clue that some paths are the old_path of a rename, and if you click the link, it takes you to a non-existent page. Do we want to open an issue for this? I'm afraid detecting renamed file paths for each commit will be very resource intensive.


lib/gollum/views/latest_changes.rb
```diff
@@ -10,6 +10,7 @@ def title
 
       def versions
         i = @versions.size + 1
+        url_start_pos = @page_dir.to_s.empty? ? 0 : Pathname.new(@page_dir).cleanpath.to_s.length+1
```
bartkamphorst: Could you try to make this a little more legible? Difficult to understand what is going on here.

lib/gollum/views/helpers.rb
```diff
@@ -2,6 +2,15 @@
 
 module Precious
   module Views
+    
+    module AppHelpers
+      def extract_page_dir(path)
+        return path if @page_dir.to_s.empty?
+        @extract_path ||= "#{Pathname.new(@page_dir).cleanpath.to_s}/"
```
bartkamphorst: The `to_s` call here is superfluous since you are using string interpolation. So `"#{Pathname.new(@page_dir).cleanpath}/"` will already end up being a String.

lib/gollum/views/helpers.rb
```diff
@@ -2,6 +2,15 @@
 
 module Precious
   module Views
+    
+    module AppHelpers
+      def extract_page_dir(path)
+        return path if @page_dir.to_s.empty?
```
bartkamphorst: Why convert `@page_dir` to a String here? If `@page_dir` is non-empty, it will evaluate as true, and if it's `nil`, it will evaluate to false. So this should also work:
```ruby
return path unless @page_dir
```

lib/gollum/views/helpers.rb
```diff
@@ -2,6 +2,15 @@
...
+        @extract_path ||= "#{Pathname.new(@page_dir).cleanpath.to_s}/"
+        path.start_with?(@extract_path) ? path[@extract_path.length..-1] : path
```
bartkamphorst: I wonder if `@path_to_extract` would be a more expressive variable name than `@extract_path`.

Also, I think `path[@extract_path.length..-1]` could also be implemented using `slice` or `slice!`, which would look a little cleaner.

lib/gollum/views/helpers.rb
```diff
@@ -43,6 +52,10 @@ def self.included(base)
           @@route_methods.to_json
         end
       end
+
+      def page_route(page)
+        "#{base_url}/#{page}".gsub(/\/{2,}/, '/')
```
bartkamphorst: Could you add a comment for future reference to explain what substitution is being performed here?

dometto:
Thanks for the suggestions!

dometto:
I noticed the following: the file links in the latest changes view are all to the historical version of the file at the given commit. That also goes for the topmost change, even though that commit is by definition `HEAD`. Should we change the link for the topmost change by doing e.g. the following in `latest_changes.rb:27`:
`:link => i == @versions.size ? "#{page_route(page_path)}" : "#{page_route(page_path)}/#{v.id}"`
...orr just leave it as is and opt for consistency?

bartkamphorst:
I would leave it as is.

dometto:
Changed the API for `Gollum::Commit#stats` to be more readable *and* to include rename detection: if `stats` includes an `:old_path` key, the view will know that the change is a rename from `:old_path` to `:new_path`.

bartkamphorst:
> Changed the API for Gollum::Commit#stats to be more readable and to include rename detection

Nice! 👍

dometto:
Because I didn't want to update the `grit_adapter` to conform to the new requirements for `Commit#stats` I've changed the tests to using `rugged` and `rjgit` respectively. I've added an item to the Release Todo to remind us to remove this from the test helper file when we're ready to release (the git adapter should be inferred by gollum-lib at that point).

dometto:
Also to clarify: the RJGit adapter doesn't yet support this new behavior, so (only) the jruby tests currently fail. See https://github.com/repotag/gollum-lib_rjgit_adapter/issues/8

dometto:
The only remaining test failure is the (unrelated) file upload detection one. The last_changes tests pass. Also tested locally, so merging this.

PR #1396, opened 2019-08-18
Pagination for history and latest changes.

dometto:
Add pagination for the latest_changes and history views.

Results:
Screenshot 2019-08-18 at 17 03 28
Screenshot 2019-08-18 at 17 03 40

(There is only a 'Next' link on the first screenshot because it's the first page)

I also noticed that when you'd view an old version of a file (e.g. by getting `Bilbo-Baggins.md/59721c7d7ca4fa6a07a50d76b3b0c92adfb8c3db`) the 'Delete Page' and 'When was this page last edited?' links were still visible in the footer. On historical pages, I've replaced this with the following:

Screenshot 2019-08-18 at 17 24 16

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -388,7 +388,7 @@ module Precious
         unless @page.nil?
           @versions = @page.versions(
             per_page: @max_count,
-            page: @page_num,
+            page_num: @page_num,
             follow: settings.wiki_options.fetch(:follow_renames,
               ::Gollum::GIT_ADAPTER == 'rjgit' ? false : true)
           )
@@ -402,7 +402,7 @@ module Precious
         @wiki = wiki_new
         @page_num = [params[:page_num].to_i, 1].max
         @max_count = settings.wiki_options.fetch(:pagination_count, 10)
-        @versions = @wiki.latest_changes(::Gollum::Page.log_pagination_options(per_page: @max_count, page: @page_num))
+        @versions = @wiki.latest_changes(::Gollum::Page.log_pagination_options(per_page: @max_count, page_num: @page_num))
         mustache :latest_changes
       end
 
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -385,9 +385,9 @@ module Precious
         @name     = wikip.fullname
         @page     = wikip.page
         @page_num = [params[:page_num].to_i, 1].max
-        @max_count = settings.wiki_options.fetch(:latest_changes_count, 10)
+        @max_count = settings.wiki_options.fetch(:pagination_count, 10)
         unless @page.nil?
-          @versions = @page.versions(:max_count => @max_count, :skip => (@page_num-1) * @max_count, :follow => settings.wiki_options.fetch(:follow_renames, ::Gollum::GIT_ADAPTER == 'rjgit' ? false : true))
+          @versions = @page.versions(:per_page => @max_count, :page => @page_num, :follow => settings.wiki_options.fetch(:follow_renames, ::Gollum::GIT_ADAPTER == 'rjgit' ? false : true))
           mustache :history
         else
           redirect to("/")
@@ -397,7 +397,7 @@ module Precious
       get '/latest_changes' do
         @wiki = wiki_new
         @page_num = [params[:page_num].to_i, 1].max
-        @max_count = settings.wiki_options.fetch(:latest_changes_count, 10)
+        @max_count = settings.wiki_options.fetch(:pagination_count, 10)
         @versions = @wiki.latest_changes({:max_count => @max_count, :skip => (@page_num-1) * @max_count})
         mustache :latest_changes
       end
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -389,7 +389,8 @@ module Precious
           @versions = @page.versions(
             per_page: @max_count,
             page: @page_num,
-            follow: settings.wiki_options.fetch(:follow_renames, ::Gollum::GIT_ADAPTER == 'rjgit' ? false : true)
+            follow: settings.wiki_options.fetch(:follow_renames,
+              ::Gollum::GIT_ADAPTER == 'rjgit' ? false : true)
           )
           mustache :history
         else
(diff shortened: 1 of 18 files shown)


lib/gollum/app.rb
```diff
@@ -382,9 +384,10 @@ class App < Sinatra::Base
         wikip     = wiki_page(params[:splat].first)
         @name     = wikip.fullname
         @page     = wikip.page
-        @page_num = [params[:page].to_i, 1].max
+        @page_num = [params[:page_num].to_i, 1].max
+        @max_count = settings.wiki_options.fetch(:latest_changes_count, 10)
```
bartkamphorst: Why `:latest_changes_count` in `/history` and not something more generic like `pagination_take`?

lib/gollum/public/gollum/stylesheets/template.scss
```diff
@@ -727,3 +727,14 @@ a {
     }
   } 
 }
+
```
bartkamphorst: Instead of adding CSS to `template.scss`, it would be preferable I think to use Primer.css classes to the pagination mustache template.

bartkamphorst: (For example, using the class `float-right` on the span "next".)

lib/gollum/app.rb
dometto: Thanks, I like `pagination_take`!

dometto: Or `pagination_take_count` or `pagination_count`?

bartkamphorst: `pagination_take_size`?

lib/gollum/public/gollum/stylesheets/template.scss
dometto: Done! Thanks for the tip, much cleaner like this.

dometto:
Discovered a little problem: when scrolling to the last page of commits on the History view, I saw this;
Screenshot 2019-08-19 at 12 07 26
Obviously it's not possible to compare this revision to another. :) In general, you can't currently compare a commit on one page to one on another. Any ideas about how we could fix this? Perhaps some ajax magic again?

bartkamphorst:
> In general, you can't currently compare a commit on one page to one on another. Any ideas about how we could fix this? Perhaps some ajax magic again?

Ah. Yes, that would be an issue. One possible solution would be to persist the selected checkboxes using jquery. So each time Next/Prev is hit, you keep track of the selection and update it visually if needed (e.g., if you return to the first page where you already selected one or more revisions).

dometto:
> Ah. Yes, that would be an issue. One possible solution would be to persist the selected checkboxes using jquery. So each time Next/Prev is hit, you keep track of the selection and update it visually if needed (e.g., if you return to the first page where you already selected one or more revisions).

Good idea! We could also indicate what commit is currently selected next to the Compare Revisions button (when the selection is on a different page)?

On the buttons: do you think it's necessary to have two of them? If not, which should go?

bartkamphorst:
> We could also indicate what commit is currently selected next to the Compare Revisions button (when the selection is on a different page)?

That would certainly be nice. 👍 

> On the buttons: do you think it's necessary to have two of them? If not, which should go?

No, the need for two buttons is gone now that we have pagination. I would opt to keep only the button on the bottom end of the page. That's where I would expect it in any case.

dometto:
Completed work on the history view:
* selection from a previous/next page is now remembered
  * accomplished this by storing selection in a hidden form, which is preserved when clicking the 'Next' or 'Previous' links, because these now use ajax.
* checkboxes are disabled when the number of selected commits is two
* the submit button is disabled unless exactly two commits are selected

Question: it appears that the current functionality (already before this PR) is to make the comparison order (when clicking 'Compare') depend *on which commit was selected first*. That is: if you first select A and then B, then click compare, you will get the reverse diff from when you first select B, and then A. Do you think is desirable?

I don't think it's very harmful, but theoretically, we could strictly determine comparison order by submitting the dates of the commits along to the compare route, and letting it select the oldest.

dometto:
> No, the need for two buttons is gone now that we have pagination. I would opt to keep only the button on the bottom end of the page. That's where I would expect it in any case.

👍  Also removed the 'Back to top' link.

lib/gollum/app.rb
```diff
@@ -397,7 +401,10 @@ class App < Sinatra::Base
         @wiki = wiki_new
         @page_num = [params[:page_num].to_i, 1].max
         @max_count = settings.wiki_options.fetch(:pagination_count, 10)
-        @versions = @wiki.latest_changes({:max_count => @max_count, :skip => (@page_num-1) * @max_count})
+        @versions = @wiki.latest_changes(::Gollum::Page.log_pagination_options(
+          per_page: @max_count,
+          page: @page_num
+        ))
```
dometto: This an improvement?

bartkamphorst: > This an improvement?

See my comments.

dometto:
One more question: what about having some more buttons in the nav bar on the history overview? For example:

Screenshot 2019-08-21 at 21 16 26

(I added the Current Page button for experimentation)

I also like the current, cleaner look, but one can imagine people wanting to search, go to the overview, etc., based on what they see in a page's history.

bartkamphorst:
> Question: it appears that the current functionality (already before this PR) is to make the comparison order (when clicking 'Compare') depend on which commit was selected first.  [...] Do you think is desirable?

I'm not actually seeing this behavior. However, I would like this behavior to exist. 😸 If it's possible to create diffs and reverse diffs of the same two commits, we could make the selected items visually distinct also in the colors of the diff: one selected row green (A) and the other red (B).

dometto:
> The per_page option I understand, but the page option is confusing. I think we should reserve page variables strictly for Gollum::Pages and use page_num here. Alternatively we could have pagination_count and pagination_page.

I agree, but this would require changes to gollum-lib. Happy to make them if you think it's best.

dometto:
> I'm not actually seeing this behavior. However, I would like this behavior to exist. 😸 If it's possible to create diffs and reverse diffs of the same two commits, we could make the selected items visually distinct also in the colors of the diff: one selected row green (A) and the other red (B).

Ah, so then it's a result of my changes after all! Accidental but neat. I really like the idea about coloring, will get to it :)

bartkamphorst:
> One more question: what about having some more buttons in the nav bar on the history overview?

If you think it's functional then I'm not principally opposed to it. The risk I guess with the search bar there is that it suggests being able to search through the table, using for example words from a commit message that you are looking for. It might be an unwelcome surprise to be brought to the general search results.

bartkamphorst:
> I agree, but this would require changes to gollum-lib. Happy to make them if you think it's best.

Yes, let's do this now that we have some momentum going! 😉

dometto:
> If it's possible to create diffs and reverse diffs of the same two commits, we could make the selected items visually distinct also in the colors of the diff: one selected row green (A) and the other red (B).

Screenshot 2019-08-21 at 22 12 49
Screenshot 2019-08-21 at 22 13 01

PR #1398, opened 2019-08-18
Show subpages in preview

dometto:
Show subpages on preview page. They are rendered dynamically based on the content of the subpage textareas in the editor -- so it's an actual preview, rather than rendering the committed sidebar/header/footer!

Requires merge of https://github.com/gollum/gollum-lib/pull/333

Result:
Screenshot 2019-08-18 at 19 41 20

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -366,6 +366,9 @@ module Precious
         wiki           = wiki_new
         @name          = params[:page] || "Preview"
         @page          = wiki.preview_page(@name, params[:content], params[:format])
+        ['sidebar', 'header', 'footer'].each do |subpage|
+          @page.send("set_#{subpage}".to_sym, params[subpage]) if params[subpage]
+        end
         @content       = @page.formatted_data
         @toc_content   = wiki.universal_toc ? @page.toc_data : nil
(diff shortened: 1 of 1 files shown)


lib/gollum/app.rb
```diff
@@ -366,6 +366,9 @@ class App < Sinatra::Base
         wiki           = wiki_new
         @name          = params[:page] || "Preview"
         @page          = wiki.preview_page(@name, params[:content], params[:format])
+        ['sidebar', 'header', 'footer'].each do |subpage|
+          @page.send("set_#{subpage}".to_sym, params[subpage]) if params[subpage]
```
bartkamphorst: What sets `params[subpage]`? Also, what does `params[subpage]` contain? A single string or an array of strings?

dometto: It's being set by the ajax call [here](https://github.com/gollum/gollum/blob/5.x/lib/gollum/public/gollum/javascript/gollum.js.erb#L390). Each field contains a string.

bartkamphorst: Got it!

bartkamphorst:
All good, merge when ready.

PR #1405, opened 2019-08-27
Remove FontAwesome

dometto:
Noticed that the little link symbols that appear (in 4.x) when you hover above a header anchor in a page, left of the header, are not working anymore since the Primer merge. This seems to be because FontAwesome isn't working anymore. When looking into this, I discovered there were still traces of FontAwesome in the CSS, though. This PR:

* Removes references to FontAwesome

Todo:
- [x] Implement alternative anchor hover symbol
- [x] Remove `<i class="fa fa-link">` tags from gollum-lib
  * See https://github.com/gollum/gollum-lib/pull/337
- [x] Replace FA spinner at
  * `gollum.js.erb:465`
  * `editor.mustache:103`, `editor.scss:278-284`, `gollum.js.erb:114-164`


dometto:
@bartkamphorst do you know of anything else that we might still be using FA for?

bartkamphorst:
> do you know of anything else that we might still be using FA for?

Nope.

dometto:
I first thought we could implement anchors using [AnchorJS](https://github.com/bryanbraun/anchorjs), but that would require us to set the `id` tags on our headers. That conflicts with some markup rendering gems, which set their own `id`. So I styled my own anchor links. I've added:
* An octicon `link` icon to the left of each header (on hover)
  * These anchors are added by `gollum-lib`
* An octicon `pencil` icon to the right of each header with class `editable` (on hover)
  * This link takes you to `/gollum/edit/PageName#anchor_name`, for later use in creating editable sections
  * These anchors are added by `gollum.js` (to limit html generation in `gollum-lib`)

Result (on hover screenshot):

Screenshot 2019-08-28 at 23 57 29

dometto:
@bartkamphorst feel free to comment on/tweak the styling!

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -482,7 +482,19 @@ $(document).ready(function() {
     $('span.critic.comment').filter(function() {return $(this).text()!="";}).before('&#8225;');
   }
 
-  $('.markdown-body p, .markdown-body span, .markdown-body pre, .markdown-body table').attr('dir','auto');
+  if($('#wiki-content').length ){
+    // Set text direction auto
+    $('.markdown-body p, .markdown-body span, .markdown-body pre, .markdown-body table').attr('dir','auto');
+
+    // Copy anchors for each editable header and give it the new anchor the 'edit' class, to display a link to edit the section
```
bartkamphorst: Check grammar of this comment.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -482,7 +482,19 @@ $(document).ready(function() {
...
+    $('a.anchor').each(function (index, anchor) {
+      header = $(anchor).closest(':header');
+      if (header.hasClass('editable')){
+        var newUrl = routePath('edit') + '/' + pageName() + $(anchor).attr('href');
+        $(anchor).clone().addClass('edit').attr('href', newUrl).appendTo(header);
```
bartkamphorst: Why clone the anchor here?

lib/gollum/public/gollum/stylesheets/template.scss.erb
```diff
@@ -2,72 +2,6 @@
...
-  line-height: 1;
-  text-decoration: none;
-  @include font-smoothing(on);
-}
-
-.fa-link:before {
-  content: "\f0c1";
-}
-
-.fa-spinner:before {
```
bartkamphorst: Actually, don't we use the spinner for uploading?

lib/gollum/views/helpers.rb
```diff
@@ -51,6 +51,13 @@ def self.included(base)
         def rocticon(symbol, parameters = {})
           Octicons::Octicon.new(symbol, parameters).to_svg
         end
+
+        def rocticon_css(symbol, parameters = {})
```
bartkamphorst: I dont' know if it's faster per se, but here's a more Ruby-ish way of achieving what you want:
```ruby
def rocticon_css(symbol, parameters = {})
  octicon = Octicons::Octicon.new(symbol, parameters.merge({xmlns: 'http://www.w3.org/2000/svg'}))
  [:width, :height].each {|option| octicon.options.delete(option)}
  octicon.to_svg
end
```
The outputted string from this snippet is already utf-8 encoded.

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: 😄

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: Do you mean: why clone it at all, or why clone it here? Since I wanted to insert a second (edit) anchor, I thought cloning the first one might require less jQuery calls.

lib/gollum/public/gollum/stylesheets/template.scss.erb
dometto: Good point! That probably broke after I removed the fonts anyway. Maybe there is a Primer replacement?

I'll search the source for ` fa ` and `fa-` to see if we forgot about more features.

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: I was wondering why to clone it at all; didn't realize you were adding anchors, I just thought you were modifying the existing ones.

lib/gollum/views/helpers.rb
dometto: Did you look at the resulting HTML for that snippet? Looking at the source for `Octicon` it seemed to me that deleting the `:width` and `:height` options would result in the calculation of a default width and height. (We need the SVG not to set any height/width at all, for use in backgrounds.)

bartkamphorst:
```irb
jruby-9.1.13.0 :001 > require 'octicons'
 => true 
jruby-9.1.13.0 :002 > octicon = Octicons::Octicon.new('x', {}.merge({xmlns: 'http://www.w3.org/2000/svg'}))
 => #<Octicons::Octicon:0x341814d3 @symbol="x", @keywords=["remove", "close", "delete"], @height=16, @width=12, @options={:xmlns=>"http://www.w3.org/2000/svg", :class=>"octicon octicon-x", :viewBox=>"0 0 12 16", :version=>"1.1", :width=>12, :height=>16, :"aria-hidden"=>"true"}, @path="<path fill-rule=\"evenodd\" d=\"M7.48 8l3.75 3.75-1.48 1.48L6 9.48l-3.75 3.75-1.48-1.48L4.52 8 .77 4.25l1.48-1.48L6 6.52l3.75-3.75 1.48 1.48L7.48 8z\"/>"> 
jruby-9.1.13.0 :003 > [:width, :height].each {|option| octicon.options.delete(option)}
 => [:width, :height] 
jruby-9.1.13.0 :004 > octicon.to_svg
 => "<svg xmlns=\"http://www.w3.org/2000/svg\" class=\"octicon octicon-x\" viewBox=\"0 0 12 16\" version=\"1.1\" aria-hidden=\"true\"><path fill-rule=\"evenodd\" d=\"M7.48 8l3.75 3.75-1.48 1.48L6 9.48l-3.75 3.75-1.48-1.48L4.52 8 .77 4.25l1.48-1.48L6 6.52l3.75-3.75 1.48 1.48L7.48 8z\"/></svg>" 
jruby-9.1.13.0 :005 > 
```

bartkamphorst: No height or width set; `to_svg` just processes the keys present in the options hash.

dometto: Sweeet! Thanks

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: Needed the second anchor to display the pencil icon/edit link. Still unsure whether it's best to do that in the JS or in `gollum-lib`. As I explained in the conversation I would have preferred to do it all with AnchorJS so there is less html-generating in gollum-lib, and I thought this might be second best. If you think it may be clearer to have it all in one place (gollum-lib) let me know.

bartkamphorst: This is fine.

dometto:
The spinner was the only remaining FontAwesome implement I could find (searched by grepping ` fa ` and `fa-`). We could replace it with a pure CSS spinner from [here](https://loading.io/css/)

bartkamphorst:
Okay, sounds good. I would opt for the lds-default or lds-spinner.

dometto:
Turns out these are much easier to implement:
http://franzheidl.github.io/spinners/

dometto:
@bartkamphorst had to manually merge/push this branch. Please have a look at the changes to the spinner to see if I messed up somehow.

dometto:
Also, I noticed that the JS for uploading via the Upload button (and then the dialog) is setting and removing the `uploading` class to `#wiki-content` (see gollum.js.erb:114-164). However, this is not toggling a spinner. If we want to show a spinner, that is a separate issue. I'm also alright with there not being a spinner when uploading in that way.

PR #1408, opened 2019-08-29
Fix usericons

dometto:
Discovered the usericons (gravatar/identicon) where not working due to a minor mistake in `bin/gollum`. After fixing that, discovered the styling and markup was off -- gravatar was completely broken, identicon didn't work on the history page.

Also discovered a pretty good opportunity for refactoring: the latest_changes and history views had duplicate code related to generating identicon and gravatar codes. It also turns out it was unnecessary to generate a different hash for gravatar and identicon (the latter is abritrary, so might as well use the gravatar hash).

New look for gravatars (history page):

Screenshot 2019-08-29 at 15 30 13

And on the latest changes page:

Screenshot 2019-08-29 at 15 29 33

After getting identicons working, I decided I didn't particularly like their styling:

Screenshot 2019-08-29 at 15 47 16

...so I replaced `identicon_canvas.js` with [identicon.js](https://github.com/stewartlord/identicon.js/tree/master). Nicer look, and also allows some simpler JS. Result is much more github-like:

Screenshot 2019-08-29 at 16 32 21

Finally, I have removed the `man24.png` image, which was used as a default image when gravatar/identicon rendering failed. Now using the `person` octicon for this (had to paste in the `rocticon_css` view helper from the other PR to get this to work):

Screenshot 2019-08-29 at 17 02 02


dometto:
Tests are failing because of an update to gollum-lib. Restart after merging https://github.com/gollum/gollum/pull/1405

bin/gollum
```diff
@@ -151,7 +151,7 @@ MSG
   end
   opts.on("--user-icons [MODE]", [:gravatar, :identicon, :none], "Use specific user-icons for history view.",
     "Can be set to 'gravatar', 'identicon' or 'none'. Default: 'none'.") do |mode|
-    wiki_options[:user_icons] = mode
+    wiki_options[:user_icons] = mode.to_s
```
bartkamphorst: This shouldn't be necessary, should it? I would assume that arguments are parsed as strings already?

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -1,5 +1,14 @@
 // Helpers
 
+// Replace broken user icons with 'person' octicon
+function brokenAvatarImage(image){
+  image.onerror = '';
+  image.src = 'data:image/svg+xml;utf8,<%= rocticon_css(:person) %>';
+  image.height = '20';
```
bartkamphorst: Is this the best way of setting the dimensions?

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -484,20 +493,18 @@ $(document).ready(function() {
 
   $('.markdown-body p, .markdown-body span, .markdown-body pre, .markdown-body table').attr('dir','auto');
 
-  if( $('#wiki-history').length ){
-    var lookup = {};
+  if( $('#wiki-history').length || $('#page-history').length){
+    var options = {
+      format: 'svg',
+      size: 20,
```
bartkamphorst: Same question. Not great to have these size dimensions be defined in so many different places.

bin/gollum
dometto: Annoyingly, `mode` is a symbol here. :(

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: I see your point. I could set the size in the CSS for `img.gravatar` and `img.identicon`, but I think I would still need to enter a size option here. (Or if I leave it out, maybe it would default to a much larger size, which would be scaled back by the CSS). Or did you have another suggestion?

dometto:
@bartkamphorst I removed the size parameters from the JS. It's now controlled via the CSS, much cleaner. Thanks for the feedback.

Final question: one idea could be to add the `person` octicon as a default avatar when `--user-icons` is set to `none`. Looks like this (history and latest changes view respectively):

Screenshot 2019-08-31 at 21 42 17
Screenshot 2019-08-31 at 21 42 28

Versus the current look:

Screenshot 2019-08-31 at 21 43 45
Screenshot 2019-08-31 at 21 43 53

I'm on the fence about this. What would you prefer?

dometto:
Tests failures are due to problems with file upload, ignoring.

PR #1409, opened 2019-08-31
Improve page-info-toggle JS

dometto:
Toggling the 'When was this page last edited?' link caused text in the footer to move upward. Fix.


dometto:
Test failures preexisting and unrelated, merging.

PR #1412, opened 2019-09-03
Add uploading tests for mode page, with and without base-path enabled.

bartkamphorst:


bartkamphorst:
> Any idea whence the travis errors?

The 2 failures on JRuby are the ones that are expected since 1) conflict detection is currently broken and 2) because base-path is not respected in the upload_file route. However, the errors for ruby 2.3 and 2.4 (closed IO stream) I have to look into.

bartkamphorst:
> the errors for ruby 2.3 and 2.4 (closed IO stream) I have to look into.

Fixed these issues in https://github.com/gollum/gollum/pull/1412/commits/1c98a787ab20e082ea68f7c9024840b0e4cea7da by reopening the tempfile before passing it to `Rack::Test::UploadedFile.new()`.

PR #1415, opened 2019-09-03
Editable sections

dometto:
When clicking the little edit anchor on a section header on a page, the user will be taken to `/gollum/edit/PageName#section`. In the edit view, we use ace to search through the document for section definitions. When we find one, we generate an anchor for it, and compare it to the anchor searched for (`#section`). When there's a match, the editor scrolls to the relevant line, and the cursor is place there.

Complications:
  * Not all markups have their own ace modes. If they don't, we can't apply the above procedure. Hence we test whether the markup supports editable sections before applying it, and also when determining whether to insert the clickable pencil/edit anchor next to the header in the first place (in `gollum.js`).
  * Some markups (markdown, rest) have syntax that puts the section definition and its title on different lines. Ace only finds the definition (e.g. in markdown `====`), and doesn't know about the title on the other line. I added a `'gollum-helpers'` field to the relevant language definitions, which provides a function for getting the title from the other line. (In both markdown and rest, this is the previous line, but since we can imagine a markup language which has first `====` and then the title on the next line, I've let this up for determination by the helper function.)

Todo: - [ ] extend the ace modes for the supported formats to recognize gollum-specific block-level. syntax, such as UML and backtick-style code blocks.
 * This is so ace correctly ignores what will look like section definitions within such blocks.

### DIFF
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
@@ -219,6 +219,15 @@
         }
       } // EditorHas.functionBar
 
+      if (ActiveOptions['section'] && $.markupSupportsEditableSections ( ActiveOptions.MarkupType )) {
+        var sectionLine = $.findSection(ActiveOptions['section'], LanguageDefinition.getDefinitionFor('gollum-helpers'));
+        if (sectionLine) {
+          window.ace_editor.gotoLine(sectionLine+1, 0, false);
+          window.ace_editor.scrollToLine(sectionLine, false, false);
+          window.ace_editor.focus();
+        }
+      }
+
       if ( EditorHas.dragDropUpload() ) {
         var $editorBody = $('#gollum-editor-body-ace');
         var editorBody = $('#gollum-editor-body-ace')[0];
--- a/lib/gollum/public/gollum/javascript/editor/langs/markdown.js
+++ b/lib/gollum/public/gollum/javascript/editor/langs/markdown.js
@@ -6,8 +6,35 @@
 
 (function($) {
 
-var MarkDown = $.extend(true, {}, $.DefaultLang); // The default language definition is markdown, so copy it into a new variable.
-$.GollumEditor.defineLanguage('markdown', MarkDown);
+// The default language definition (in default.js) is markdown, so we only need to set the special helper functions.
+var MarkDown = {
+
+  'gollum-helpers': {
+
+    'find-header-line': function (line, token) {
+      // For use in sections.js
+      // Checks if token matches alternative-syntax markdown headers. These are of the form '===' or '---'.
+      // If so, returns an array containing the text of the header and the previous line number.
+      // If not, returns null.
+      if ( token.match(/(^[=]+$)/) || token.match(/(^[-]+$)/) ){
+        var foundLine = line-1
+        var text = window.ace_editor.getSession().getLine(foundLine);
+        if (foundLine >= 0 && !/^\s+$/.test(text)) { // Sanity check
+          return [text, foundLine];
+        }
+        else {
+          return null;
+        }
+      } else {
+        return null;
+      }
+    }
+
+  }
+
+};
+
+$.GollumEditor.defineLanguage('markdown', $.constructLanguageDefinition(MarkDown));
 
 var MarkDownHelp = [
 
--- a/lib/gollum/public/gollum/javascript/editor/langs/rest.js
+++ b/lib/gollum/public/gollum/javascript/editor/langs/rest.js
@@ -83,6 +83,22 @@
                                   }
                                   return rep;
                                 }
+                              },
+
+    'gollum-helpers'      :   {
+
+                                'find-header-line': function (line, token) {
+                                  // For use in sections.js
+                                  // Return an array containing the text of the header and the previous line number.
+                                  var foundLine = line-1;
+                                  var text = window.ace_editor.getSession().getLine(foundLine);
+                                  if (foundLine >= 0 && !/^\s+$/.test(text)) { // Sanity check
+                                    return [text, foundLine];
+                                  }
+                                  else {
+                                    return null;
+                                  }
+                              }
                               }
 
   };
--- a/lib/gollum/public/gollum/javascript/editor/sections.js
+++ b/lib/gollum/public/gollum/javascript/editor/sections.js
@@ -1,0 +1,93 @@
+(function($) {
+
+  $.markupSupportsEditableSections = function ( markup ) {
+    var supportedFormats = ['markdown', 'textile', 'asciidoc', 'rst'];
+    return supportedFormats.includes(markup);
+  }
+
+  $.findSection = function ( anchor, languageHelpers ) {
+    var titles = {};
+
+    for (i = 0; i < window.ace_editor.getSession().getLength(); i++) {
+      var foundLine = null;
+
+      var results = getSectionTokens(i);
+      var section = results[0];
+      var header  = results[1];
+
+      if (!section) {
+        continue; // No section on this line, go to the next
+      }
+
+      if (header) {
+        header = header['value'];
+        foundLine = i; // Header text was found on this line
+      } else
+      {
+        // No header text found on this line. Perhaps the markup language puts section-defining syntax and the title of the section on different lines?
+        var header_and_line = getHeader(i, section['value'], languageHelpers);
+        if ( header_and_line.length ) {
+          header = header_and_line[0];
+          foundLine = header_and_line[1];        
+        } else {
+          continue; // No header text was found again. Skip this line.
+        }
+      }
+
+      // Now that we have the header text, compare it to the anchor we're looking for.
+      header = generateAnchor(header);
+
+      // If we've encountered this header before, add an index to it (e.g. 'bilbo-baggins-1'.
+      if (index = titles[header]) {
+        header = header + '-' + index.toString();
+        titles[header] = index+1;
+      } else {
+        titles[header] = 1;
+      }
+
+      // Compare the final header to the anchor we're looking for
+      if (header == anchor) {
+        return foundLine;
+      }
+    }
+    return null;  
+  }
+
+  function getSectionTokens ( line ) {
+    var session = window.ace_editor.getSession();
+    tokens = session.getTokens(line);
+
+    // Assume we can't have more than one section heading per line, so take the first section and the first header text
+    section = tokens.filter(function (el) {
+      return /markup.heading/.test(el.type);
+    })[0];
+
+    header = tokens.filter(function (el) {
+      return /^(heading)|(text)$/.test(el.type); 
+    })[0];
+
+    return [section, header];
+  }
+
+  function getHeader ( section_line, section_token, languageHelpers ) {
+    if (!languageHelpers || !languageHelpers['find-header-line']) {
+      return [];
+    }
+    var results  = languageHelpers['find-header-line'](section_line, section_token);
+    if ( results == null ) {
+      return [];
+    } else {
+      return [results[0], results[1]];
+    }
+  }
+
+  function generateAnchor ( header ) {
+    header = header.trim();
+    header = header.replace(/[^a-z0-9_\s]/gi, ' ');
+    header = header.replace(/\s\s+/g, ' ');
+    header = header.split(/\s/).join('-');
+    header = header.replace(/^-/,'');
+    header = header.toLowerCase();
+    return header;
+  }
+})(jQuery);
--- a/lib/gollum/public/gollum/javascript/gollum.js.erb
+++ b/lib/gollum/public/gollum/javascript/gollum.js.erb
@@ -340,7 +340,9 @@ $(document).ready(function() {
     $("#gollum-editor-body").one('change', function(){
       window.onbeforeunload = function(){ return "Leaving will discard all edits!" };
     });
-    $.GollumEditor();
+    $.GollumEditor({
+      section: window.location.hash.substr(1)
+    });
     $("#gollum-editor-submit").click( function(e) {
       e.preventDefault();
       // Prevent button from being clicked again
@@ -499,14 +501,16 @@ $(document).ready(function() {
     // Set text direction auto
     $('.markdown-body p, .markdown-body span, .markdown-body pre, .markdown-body table').attr('dir','auto');
 
-    // Copy anchors for each editable header and give the new anchor the 'edit' class, to display an "edit section" link
-    $('a.anchor').each(function (index, anchor) {
-      header = $(anchor).closest(':header');
-      if (header.hasClass('editable')){
-        var newUrl = routePath('edit') + '/' + pageName() + $(anchor).attr('href');
-        $(anchor).clone().addClass('edit').attr('href', newUrl).appendTo(header);
-      }
-    });
+    if ($.markupSupportsEditableSections(pageFormat)){
+      // Copy anchors for each editable header and give the new anchor the 'edit' class, to display an "edit section" link
+      $('a.anchor').each(function (index, anchor) {
+        header = $(anchor).closest(':header');
+        if (header.hasClass('editable')){
+          var newUrl = routePath('edit') + '/' + pageName() + $(anchor).attr('href');
+          $(anchor).clone().addClass('edit').attr('href', newUrl).appendTo(header);
+        }
+      });
+    }
   }
 
   if( $('#wiki-history').length || $('#page-history').length){
--- a/lib/gollum/templates/layout.mustache
+++ b/lib/gollum/templates/layout.mustache
@@ -22,6 +22,7 @@
 	}
 	{{#page}}
 	  var pageFullPath = '{{escaped_url_path}}';
+    var pageFormat   = '{{format}}';
 	{{/page}}
 
   </script>
--- a/lib/gollum/public/gollum/javascript/app.js
+++ b/lib/gollum/public/gollum/javascript/app.js
@@ -6,6 +6,7 @@
 //= require gollum.dialog
 //= require gollum.placeholder
 //= require editor/gollum.editor
+//= require editor/sections
 //= require editor/langs/default
 //= require_tree ./editor/langs
 //= require ace/ace


dometto:
Also a caveat: when the document is very small, this implementation won't have any recognizable effect for the user, as the editor will not be large enough to scroll at all.

A more advanced implementation would hide/cut off text outside the requested section, but I believe this is already a nice feature.

bartkamphorst:
> Also a caveat: when the document is very small, this implementation won't have any recognizable effect for the user, as the editor will not be large enough to scroll at all.

Makes sense, I don't think this even counts as a caveat. 😉 

> A more advanced implementation would hide/cut off text outside the requested section [...]

I tend to disagree about this. If you're in a large document, it's super helpful to launch straight into the relevant section, but strange to lose all context of the surrounding text. It would also mean having to save, go back to the page, and select another section to continue editing if you want to edit another section, instead of staying in the editor and scrolling.

bartkamphorst:
Question: will the preview also be responsive to the section anchor, so that the user will see the edited section in the preview without having to scroll?

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -219,6 +219,15 @@
         }
       } // EditorHas.functionBar
 
+      if (ActiveOptions['section'] && $.markupSupportsEditableSections ( ActiveOptions.MarkupType )) {
+        var sectionLine = $.findSection(ActiveOptions['section'], LanguageDefinition.getDefinitionFor('gollum-helpers'));
+        if (sectionLine) {
+          window.ace_editor.gotoLine(sectionLine+1, 0, false);
```
bartkamphorst: Is there any way to make this more expressive? Why `sectionLine+1`? What do the second and third argument mean? Perhaps just something like:
```javascript
window.ace_editor.gotoLine(sectionLine+1, offset=0, slap=false);
```

lib/gollum/public/gollum/javascript/editor/sections.js
```diff
@@ -0,0 +1,93 @@
...
+    return null;  
+  }
+
+  function getSectionTokens ( line ) {
+    var session = window.ace_editor.getSession();
+    tokens = session.getTokens(line);
+
+    // Assume we can't have more than one section heading per line, so take the first section and the first header text
+    section = tokens.filter(function (el) {
+      return /markup.heading/.test(el.type);
```
bartkamphorst: What does this do?

lib/gollum/public/gollum/javascript/editor/sections.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    var results  = languageHelpers['find-header-line'](section_line, section_token);
+    if ( results == null ) {
+      return [];
+    } else {
+      return [results[0], results[1]];
+    }
+  }
+
+  function generateAnchor ( header ) {
```
bartkamphorst: Why not chain these method calls? Should be quicker.

lib/gollum/templates/layout.mustache
```diff
@@ -22,6 +22,7 @@
 	}
 	{{#page}}
 	  var pageFullPath = '{{escaped_url_path}}';
+    var pageFormat   = '{{format}}';
```
bartkamphorst: Indentation is strange.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
dometto: Nice, didn't realize you could use variable assignments for this purpose!

dometto:
Thanks for the feedback!

> Makes sense, I don't think this even counts as a caveat. 😉

I guess I agree, also because there is, after all, *some* effect for the user: the cursor will still be positioned at the start of the desired header.

> Question: will the preview also be responsive to the section anchor, so that the user will see the edited section in the preview without having to scroll?

Not yet, but that's a great idea!

lib/gollum/public/gollum/javascript/editor/sections.js
dometto: it filters `tokens` whose `type` matches `/markup.heading/`. An alternative would be:
`return el.type.startsWith('markup.heading');`
Regex or `startsWith` is required because for some markup languages, type is defined e.g. as `markup.heading.1`, whereas for others it's generic `markup.heading`

bartkamphorst: `el.type.startsWith('markup.heading');` is more expressive. What is `el` short for?

dometto: `element`. I guess `token` would be preferable?

bartkamphorst: Yes.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -304,12 +304,14 @@ $(document).ready(function() {
   }
 
   if( $('.tabnav-tabs').length ){
-    $(".tabnav-tab").click( function() {
+    $(".tabnav-tab").click( function(e) {
+      e.preventDefault();
```
dometto: Without this `preventDefault`, clicking the Preview tab respects the edit anchor, but then replaces the current anchor with simply `#`. So if you go back to the Edit tab, the section heading is lost. With `preventDefault`, the anchor stays in place, so no matter how often the user switches between the Edit and Preview tab, gollum will automatically move to the selected section. I am not sure which behavior is more desirable, though: if I want to go back to the Edit tab after having edited section `Foo`, do I  thereafter want the editor to jump to section `Foo` continuously? What if I want to edit a different section after `Foo`?

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -495,18 +504,20 @@ $(document).ready(function() {
     $('span.critic.comment').filter(function() {return $(this).text()!="";}).before('&#8225;');
   }
 
-  if($('#wiki-content').length ){
+  if($('.markdown-body').length ){
```
dometto: Noticed that before, this `if` statement returned true on the `create` page.

lib/gollum/public/gollum/javascript/editor/modes.js.erb
```diff
@@ -0,0 +1,151 @@
...
+
+<%
+codeblock_rule_name = 'GitHubBlock'
+codeblock_rule_id = 'githubblock'
+%>
+
+// The rule defining the github-codeblock state for the ACE Tokenizer
+var <%= codeblock_rule_name %> = [{
+  token: "support.function",
+  regex: /^```$/,
```
dometto: gollum's current behavior is to only recognize three backticks without any whitespace before/after as a github-style codeblock, so I've copied that behavior here: the regex is strict about whitespace.

However, [here](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/filter/code.rb#L41) it looks like gollum is trying to allow for whitespace (judging by the `([ \t]*)` clauses).

According to [rubular](http:/rubular.com), `\t` is not valid ruby regex. Perhaps this has changed over the years. Question is now: do we want to be lenient about whitespace, or not? If not, should we modify the current regex in gollum-lib?

dometto:
I've extended ACE's highlighting modes to recognize gollum-specific block-level syntax:

* UML blocks
* Three-backtick codeblocks

If I haven't overlooked anything, those are the only two gollum-defined block-level entities (which therefore work universally, regardless of the page format).

This is nice anyway, but also it prevents editable sections from breaking. Previously, this code would cause the algorithm to return the incorrect, earlier line number when searching for section `## Legolas`:
```markdown
# LOTR Characters

@startuml
Some text.
## Legolas
@enduml

## Legolas
```
(and similarly for "headers" defined within 3-backtick-codeblocks on markups which previously didn't recognize these -- markdown being an exception)

This bug is now solved.

@bartkamphorst if you could give this branch a spin and see if things work intuitively, that would be great!

dometto:
Oh yes: using `erb` in the new `modes.js` was the only way I could see of preventing extensive code duplication, but if you think it's too confusing and have another strategy, let me know. :)

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: > if I want to go back to the Edit tab after having edited section Foo, do I thereafter want the editor to jump to section Foo continuously? What if I want to edit a different section after Foo?

I think one should land roughly where one left it when one switched to the preview pane. Just make sure that the behavior also makes sense for regular, full page editing.

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: Ok.

lib/gollum/public/gollum/javascript/editor/modes.js.erb
bartkamphorst: What does github do? I think that should be leading in this case. For me personally I think being strict about whitespace is fine.

bartkamphorst:
> I've extended ACE's highlighting modes to recognize gollum-specific block-level syntax:

Very cool! I'll have a look at the implementation and will also try it out. Tbc.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -219,6 +205,15 @@
         }
       } // EditorHas.functionBar
 
+      if (ActiveOptions['section'] && $.markupSupportsEditableSections ( ActiveOptions.MarkupType )) {
+        var sectionLine = $.findSection(ActiveOptions['section'], LanguageDefinition.getDefinitionFor('gollum-helpers'));
+        if (sectionLine) {
+          window.ace_editor.gotoLine(sectionLine+1, offset=0, slap=false); //gotoLine starts counting at 1 instead of 0, so +1
```
bartkamphorst: Please replace these dummy variable names with meaningful ones! 😂 
```javascript
window.ace_editor.gotoLine(sectionLine+1, column=0, animate=false);
```

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -219,6 +205,15 @@
...
+          window.ace_editor.scrollToLine(sectionLine, false, false);
```
bartkamphorst:
```javascript
window.ace_editor.scrollToLine(sectionLine, center=false, animate=false);
```
Why scrollToLine here instead of goToLine?

lib/gollum/public/gollum/javascript/editor/modes.js.erb
dometto: Github is lenient about the whitespace. That means a bit of tweaking for both the regexes in gollum-lib, and in the JS here.

dometto: Then again, gollum only supports markdown, and markdown is lenient about whitespace in general. So there is another question about whether it makes sense to be whitespace lenient in all markups.

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
```diff
@@ -6,8 +6,35 @@
...
+
+    'find-header-line': function (line, token) {
+      // For use in sections.js
+      // Checks if token matches alternative-syntax markdown headers. These are of the form '===' or '---'.
+      // If so, returns an array containing the text of the header and the previous line number.
+      // If not, returns null.
+      if ( token.match(/(^[=]+$)/) || token.match(/(^[-]+$)/) ){
+        var foundLine = line-1
+        var text = window.ace_editor.getSession().getLine(foundLine);
+        if (foundLine >= 0 && !/^\s+$/.test(text)) { // Sanity check
```
bartkamphorst: Why the sanity check? Why assume that the previous regexp could somehow fail?

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
```diff
@@ -6,8 +6,35 @@
...
+var MarkDown = {
+
+  'gollum-helpers': {
+
+    'find-header-line': function (line, token) {
+      // For use in sections.js
+      // Checks if token matches alternative-syntax markdown headers. These are of the form '===' or '---'.
+      // If so, returns an array containing the text of the header and the previous line number.
+      // If not, returns null.
+      if ( token.match(/(^[=]+$)/) || token.match(/(^[-]+$)/) ){
```
bartkamphorst: I would prefer to use just one regex. Something like this might do the trick:
```regex
^([=]{3,}|[-]{3,})$
```
(Three or more equal signs or three or more dashes, going from the beginning of the line to the end of the line.)

lib/gollum/public/gollum/javascript/editor/langs/rest.js
```diff
@@ -7,9 +7,15 @@
 
 
   var headerFunc = function (selText, selectedRange, symbol) {
-    adornment = symbol.repeat(selText.length);
-    repText = selText + "\n" + adornment + "\n";
-    $.GollumEditor.replaceSelection(repText, false, false, selectedRange);
+      var num = 10;
```
bartkamphorst: Why is this value hardcoded? What does it signify exactly?

lib/gollum/public/gollum/javascript/editor/modes.js.erb
bartkamphorst: > According to rubular, \t is not valid ruby regex.

This is hard to believe!

lib/gollum/public/gollum/javascript/editor/modes.js.erb
```diff
@@ -0,0 +1,151 @@
...
+
+<%
+uml_rule_name = 'UMLBlock'
+uml_rule_id = 'umlblock'
+%>
+
+// The rule defining the UML state for the ACE Tokenizer
+var <%= uml_rule_name %>  = [{
+  token: "support.function",
+  regex: /^\@enduml/,
```
bartkamphorst: Why only `@enduml` here?

lib/gollum/public/gollum/javascript/editor/modes.js.erb
```diff
@@ -0,0 +1,151 @@
...
+  "use strict";
+
+  var oop = require("ace/lib/oop");
+  var BaseHighlightRules = require('<%= highlighter[:path] %>').<%= highlighter[:rule_name] %>;
+
+  var GollumHighlightRules = function() {
+    var mergeRules = new BaseHighlightRules();
+    this.$rules = mergeRules.getRules();
+    <% highlighter[:start_rules].each do |rule| %>
+    this.$rules.start.unshift(<%= rule %>);
```
bartkamphorst: Indentation missing.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
dometto: Hahaha more coffee.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
dometto: we need both:
* goto moves the cursor
* scroll to changes the editor focus

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
dometto: `token` is a header found on `line`. We're looking for the title corresponding to that header, which is a different token type. It's not on the same line as `line`, that's why we're in this helper method. *If* there are situations where ACE finds a header token without there being a corresponding header title token, I thought some sanity checking might be in order: making sure at least the line exists (i.e. `line` is not 0, so `foundLine` is `>= 0`), and contains some text. But we could also leave it out: consequences would not be dire anyway.

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
dometto: 👍

lib/gollum/public/gollum/javascript/editor/langs/rest.js
dometto: I noticed if you clicked a rst header button without having any text on a line, nothing would happen. In other markups, this would just give you the header syntax without any text. This is because the rst header function produces a number of header-characters (e.g. `=`) based on the number of characters in the title (rst requires at least as many header-characters as the title length). So if you didn't have a title yet, the buttons generated 0 adornment characters. But that also means you couldn't use the header buttons to 'find out' about proper syntax in rst. `10` is just an arbitrary default that. We could also stick with the original behavior of generating no header syntax when the title is empty.

lib/gollum/public/gollum/javascript/editor/modes.js.erb
dometto: You're right, `\t` is valid. But for some reason it's not being matched in gollum-lib's regex:
`/^([ \t]*)``` ?([^\r\n]+)?\r?\n(.+?)\r?\n\1```[ \t]*\r?$/m`
Strangely, neither are the spaces -- or any other characters I change the `[ \t]` group to! Any idea why this may be?
(Even more weirdly, removing the `([ \t]*)` group at the beginning, even though it doesn't match anything, will stop a code block from being matched)

lib/gollum/public/gollum/javascript/editor/modes.js.erb
dometto: Once the tokenizer is in the `umlblock` state, we only need to look for the line that brings us back to the `start` state.

Actually, I just realized this means we're not marking the contents of UML/code blocks as code (although they will be prevented from being marked as different tokens). Maybe it would be nicer if the entire block, not just the start/stop syntax, was marked as code (and so also highlighted blue in the editor).

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
bartkamphorst: OK, keep as is then. 👍

lib/gollum/public/gollum/javascript/editor/langs/rest.js
bartkamphorst: I see. What about generating a dummy title if there is no text yet?

`Title here`
`==========`

bartkamphorst: The crucial part of the syntax is of course that the adornments have to be at least as long as the title. This comes out a little better with a dummy title than with an arbitrary number of 10 I think.

lib/gollum/public/gollum/javascript/editor/modes.js.erb
bartkamphorst: > Maybe it would be nicer if the entire block, not just the start/stop syntax, was marked as code

Agreed.

lib/gollum/public/gollum/javascript/editor/langs/markdown.js
dometto: https://xkcd.com/2200/

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
dometto: Do you think it would make more sense to place the cursor at the end of the header line?

lib/gollum/public/gollum/javascript/editor/langs/rest.js
dometto: :+1:

lib/gollum/public/gollum/javascript/editor/modes.js.erb
dometto: Think I prefer to stick to the identation level for the JS here

lib/gollum/public/gollum/javascript/editor/modes.js.erb
dometto: I've made the regexes in the JS lenient about whitespace, but now noticed the regex in gollum-lib is actually more sophisticated. It allowes whitespace only if the amount of whitespace before the opening backticks is the same as before the final backticks (see the `\1` characters in the regex). I don't see an easy way of implementing this in the JS, as the starting and closing backticks are being detected by different regexes.
Options:
* go full lenient in both gollum and the JS
* be strict in both gollum and the JS
Both options depart from github's behavior. I'll see if there is really no way of doing better in the JS.

dometto: This would be the new regex for gollum-lib if we go full lenient:
```ruby
/^[ \t]*``` ?([^\r\n]+)?\r?\n(.+?)\r?\n[ \t]*```[ \t]*\r?$/m
```

dometto: Managed to steal the github code block rule from ace's default markdown highlighter, which seems to cohere with github!

dometto: GitHub's rules seem simpler than what's implemented by ace: max 3 spaces indent for both the opening and the closing fence. Updated the JS, and the regex in gollum-lib here:
https://github.com/gollum/gollum-lib/pull/338

dometto: Changed the regexes to support github style fenced block, which means max 3 spaces indent for both the start and end fence. Ace's current implementation is more lenient about whitespace, so to make things consistent I also overrode the default highlighter for markdown (while adding support for `~~~` under markdown only).

https://github.com/gollum/gollum-lib/pull/338 will make gollum-lib consistent with github and the current JS

dometto:
@bartkamphorst let me know what you think of the current version. Implementing the code block rules turned out to be more complicated than I hoped, because markdown pages must also support tilde fenced codeblocks. ACE's rules correctly expected the closing tag of a fenced block to be longer than the opening tag, but incorrectly allowed more than three spaces. So I had to copy ACE's own approach to these highlighters, but correct for the latter problem.

See also https://github.com/gollum/gollum-lib/pull/339 . If you think we should support more than three backtick-based codeblocks on all pages, I'll still have to tweak a regex in this PR.

dometto:
Saw a new ACE was released 3 days ago so thought we'd better package/test it.

dometto:
FYI I've come up with a symlink solution so we can more easily switch/test ace versions (`javascript/ace` points to `javascript/ace-1.4.6`).

All seems fine with the new version, but waiting for @bartkamphorst to give this branch a spin.

lib/gollum/public/gollum/javascript/HOWTO_UPDATE_ACE.md
```diff
@@ -1,8 +1,8 @@
-Current ACE version: 1.4.4
-
 # How to update the ACE editor component
 
 1. Download a new ACE release from (ajax.org/ace-builds)[https://github.com/ajaxorg/ace-builds/releases].
 2. Extract the `src-min-noconflict` folder from the archive and rename it to `ace-x.y.z`, where `x.y.z` matches exactly the release version.
-3. Replace the version in `app.js` to `x.y.z`.
-4. Test if the editor works.
+3. Place the version in the javascript directory
+4. Update the symlink 'ace' to point to the new version
```
bartkamphorst: Will symlinks work on Windows?

dometto: Also note: this is only an issue for developing/running from source. When running with precompiled assets the symlink is not an issue at all.

bartkamphorst:
> waiting for @bartkamphorst to give this branch a spin.

Running this version now. There's three things I'm noticing:
1) The "edit section pencil" shows up with a delay. It's not a long wait, but it's also not instantaneous. I actually have to hover over each h1-h6 element to see if it is editable, which is a bit of a pain.
2) If I click the pencil, go to the edit page, and then hit my browser's back button, I return to the page but the pencil remains visible for the header that I previously clicked. Only the pencil, not the corresponding anchor image.
3) If I click the pencil, my cursor is being dropped at seemingly random positions in the section title. Sometimes it's at the beginning, sometimes at the end, sometimes right in the middle of the title.

The gollum code blocks and UML blocks show up in blue nicely. 👍 

The new ACE version seems fine.

I very much like that the preview button now brings me to the section I was editing, but I now see the need for a keyboard shortcut to bring me back to the edit pane, since I currently have scroll way up to click the edit tab.

That's all I have for now!

bartkamphorst:
> Do you think we should get rid of the fadein effect entirely, or tweak it?

I would be perfectly happy if it were removed to be honest.

> the pencil remains visible for the header that I previously clicked [...] this may depend on the browser

I'm seeing this on Firefox 69.0 but not on Safari 12.1. Actually, the anchor image also shows up, but disappears after half a second. Weird, no? Not a showstopper for me though.

> my cursor is being dropped at seemingly random positions in the section title

Yes, this is in the editor.

One header I have is `# A page needs headers`. The cursor ends up between the two 'e's of 'needs' consistently, both in Firefox and Safari. There's also a header `# Another section` where the cursor is dropped at the beginning of the line (before #). Finally, for `# Good stuff` the cursor is placed at the end, just after the final 'f'.

> would you prefer it always at the end, or always at the beginning of the line?

I would prefer either at the beginning of the line, or at the beginning of the paragraph/text following the header.

PR #1422, opened 2019-09-12
Split editor and app JS. Resolves #1402

### LINKED ISSUE #1402: 5.0: Split app and editor CSS
The editor JS (including the whole of ACE) is currently loaded on every page. It would be cleaner to require that JS only on the edit page. However, if we do that, we should make sure the editor JS is also precompiled.

dometto:
Also fix asset precompilation (include octicons in `assets.rb`).


bartkamphorst:
I was testing this setup and it seems to work fine, except now I have to wait for on-the-fly compilation twice, once when I load Home and once when I hit edit.

dometto:
I know. 😂 Not quite what I had intended. But for public instances in production it may make sense not to serve all the assets all the time. Happy to roll it back, too, I honestly don't quite know what's best.

PR #1430, opened 2019-09-30
Fix redirect location for editor submit

dometto:
Noticed two issues:
* editor was taking me to the wrong page when editing/saving a file in a subdirectory.
  * this was because `newPath` was missing a leading slash when `pagePath` was not empty
* In the Overview, when clicking 'New' in a subdirectory, the dialog box would say 'This page will be created under //Mordor'

Luckily, both could be addressed by introducing a new JS helper method that ensures a leading slash is present.


bartkamphorst:
Looks fine to me. 👍

PR #1433, opened 2019-10-05
Add redirect support (e.g., after renames). Fixes #1023.

### LINKED ISSUE #1023: Feature request: add redirect support
Currently, if you go to a URL which corresponds to a page which does not exist, Gollum pops up a "Create New Page" page. It would be nice to be able to override this behavior, and have selected pages redirect to other pages.

How I envision it is that there would be something like an optional "_Redirect" file which contains pairing of "source" and "target" names. This facility would only be triggered if the standard page resolution machinery failed. (That is, insert the _Redirect check right before the "Create New Page" launch.) If the requested source path is listed in the _Redirect file, the corresponding target path would be used instead. (For example, with a 301 or 307 http redirect.) If the requested path isn't listed, then the standard "Create New Page" page is shown for the original path.

This is somewhat related to issues  #516 and #565 but is different in that it is not rename-specific, instead proposing a mechanism which can handle arbitrary redirects. (The rename feature could be somewhat compatible with this feature by adding a line to the _Rename file for page moves done through gollum. Page moves done externally would require manual fixing, though.)

bartkamphorst:
This PR adds basic redirect support (#1023). Any time a rename action is successful, a redirect entry is added to the protected file `.redirects.gollum` via `wiki#add_redirect(old_path, new_path)`.

The redirects file is only queried as a last resort, when no page or file could be found.

When a redirect is found and is followed, the user is notified of this through a dismissable flash notification.

This PR also adds a helper `flashNotice(type, notice)` for adding flash notices to pages.

Screen Shot 2019-10-05 at 09 22 04

Screen Shot 2019-10-05 at 09 22 12

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -89,6 +89,7 @@ module Precious
       settings.wiki_options[:allow_editing] = settings.wiki_options.fetch(:allow_editing, true)
       @allow_editing = settings.wiki_options[:allow_editing]
       @critic_markup = settings.wiki_options[:critic_markup]
+      @redirects_enabled = settings.wiki_options.fetch(:redirects_enabled, true)
       @per_page_uploads = settings.wiki_options[:per_page_uploads]
 
       forbid unless @allow_editing || request.request_method == "GET"
@@ -278,7 +279,7 @@ module Precious
         # Renaming preserves format, so add the page's format to the renamed path to retrieve the renamed page
         new_path = "#{rename}.#{Gollum::Page.format_to_ext(page.format)}"
         # Add a redirect from the old page to the new
-        wiki.add_redirect(page.url_path, new_path.gsub(/(^\/)/, ''))
+        wiki.add_redirect(page.url_path, new_path.gsub(/(^\/)/, '')) if @redirects_enabled
 
         page = wiki_page(new_path).page
         return if page.nil?
@@ -553,7 +554,7 @@ module Precious
         mustache :page
       elsif file = wiki.file(fullpath, wiki.ref, true)
         show_file(file)
-      elsif redirect_path = wiki.redirects[fullpath]
+      elsif @redirects_enabled && redirect_path = wiki.redirects[fullpath]
         redirect to("#{encodeURIComponent(redirect_path)}?redirected_from=#{encodeURIComponent(fullpath)}")
       else
         if @allow_editing
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -279,7 +279,7 @@ module Precious
         # Renaming preserves format, so add the page's format to the renamed path to retrieve the renamed page
         new_path = "#{rename}.#{Gollum::Page.format_to_ext(page.format)}"
         # Add a redirect from the old page to the new
-        wiki.add_redirect(page.url_path, new_path.gsub(/(^\/)/, '')) if @redirects_enabled
+        wiki.add_redirect(page.url_path, clean_url(new_path)) if @redirects_enabled
 
         page = wiki_page(new_path).page
         return if page.nil?
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -158,15 +158,15 @@ module Precious
         end
       end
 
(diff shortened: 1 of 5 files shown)


bartkamphorst:
Tests pass locally, but fail on Travis because the new code depends on pending changes in gollum-lib: https://github.com/gollum/gollum-lib/pull/350.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -102,6 +102,12 @@ function preparePage () {
   }
 }
 
+function flashNotice(type, notice) {
```
dometto: Nice! Very clean

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -148,6 +154,11 @@ $(document).ready(function() {
     }
   }
 
+  if (match = new RegExp(/\?redirected\_from=([^?]*)/).exec(window.location.href)) {
```
dometto: This code could be nested under an if statement that checks whether we're viewing a page (one that tests for the existence of `div.markdown-body` or `div.wiki-content`? I forget which)

dometto: Surprising that there isn't a built-in jQuery function for getting query string arguments, but fine as long as this works!

It looks like the regex will only match if `redirected_fom` is the first parameter after the `?` though. Is that too strict?

bartkamphorst: Well, there is [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) but that is not supported by IE. This regex should work cross-browser.

> It looks like the regex will only match if redirected_fom is the first parameter after the ? though.

Would this be better? `[&?]redirected\_from=([^?]*)`

bartkamphorst: `div.markdown-body`. Will move the code there.

bartkamphorst: I've made the regex more lenient.

bartkamphorst:
I've also made the redirect functionality conditional on `@redirects_enabled`, which defaults to true but can be set to false in `config.rb`.

lib/gollum/app.rb
```diff
@@ -274,9 +275,13 @@ class App < Sinatra::Base
           return
         end
         committer.commit
-
+        
         # Renaming preserves format, so add the page's format to the renamed path to retrieve the renamed page
-        page = wiki_page("#{rename}.#{Gollum::Page.format_to_ext(page.format)}").page
+        new_path = "#{rename}.#{Gollum::Page.format_to_ext(page.format)}"
+        # Add a redirect from the old page to the new
+        wiki.add_redirect(page.url_path, new_path.gsub(/(^\/)/, '')) if @redirects_enabled
```
dometto: Alternatively, use the slower but more explicit `clean_url` helper from https://github.com/gollum/gollum/blob/5.x/lib/gollum/helpers.rb#L20

bartkamphorst: Of course, much nicer. Thanks for the suggestion!

bartkamphorst: Done.

asmeurer:
Does this feature work on the GitHub wikis? If so, is it documented somewhere how to use it? As I understand you should have a file named .redirects.gollum, but what should go in the file?

If not, do you know if GitHub plans to update their wiki to include this feature?

dometto:
See the announcement on the top of gollum's [wiki](https://github.com/gollum/gollum/wiki)

asmeurer:
Yes, I did see that, which is why I asked the question.

asmeurer:
@dometto if you don't know the answer to my questions, you can just say so. I won't hold it against you.

bartkamphorst:
@asmeurer 

> Does this feature work on the GitHub wikis?

No.

> If not, do you know if GitHub plans to update their wiki to include this feature?

No, we don't know, since we have no official connections to GitHub. As @dometto suggested, your best bet is to contact GitHub support directly to ask them if they have any plans to support redirects in the future.

PR #1434, opened 2019-10-07
Rake precompile: Add missing requires

jamieforth:
Precompiling currently throws the following error.

`NameError: uninitialized constant Gollum::Markup`

Not sure if this is the correct fix, but requiring gollum and
gollum-lib restores functionality.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -181,20 +181,22 @@ end
 
 desc 'Precompile assets'
 task :precompile do
- require './lib/gollum/views/helpers.rb'
- require './lib/gollum/assets.rb'
- require 'sprockets'
- require 'sprockets-helpers'
- require 'sass'
+  require 'gollum'
+  require 'gollum-lib'
+  require './lib/gollum/views/helpers.rb'
+  require './lib/gollum/assets.rb'
+  require 'sprockets'
+  require 'sprockets-helpers'
+  require 'sass'
   env = Precious::Assets.sprockets
   manifest = Sprockets::Manifest.new(env, './public/assets')
-    Sprockets::Helpers.configure do |config|
-      config.environment = env
-      config.prefix      = Precious::Assets::ASSET_URL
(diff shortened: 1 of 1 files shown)


dometto:
Makes sense! Thanks for the PR

dometto:
FYI: this was due to addition of a new helper which was referencing `Gollum::Markup`. Until now, the views helper didn't need gollum/gollum-lib.

dometto:
@bartkamphorst this has made me realize: the new helper uses the Gollum::Markup class to generate a JS array containing the recognized file extensions. This means that when users add a markup through `config.rb`, they will also have to precompile new assets for everything to work. Is this ok, or should we try to eliminate the need for the asset helpers to rely on Gollum::Markup?

dometto:
@jamieforth also please feel free to weigh in on the above

bartkamphorst:
> should we try to eliminate the need for the asset helpers to rely on Gollum::Markup?

Ideally, yes, but is there another feasible way to achieve what the helper achieves?

dometto:
I added it in an attempt to make the Edit route aware of where to take the user after saving, [here](https://github.com/gollum/gollum/blob/5.x/lib/gollum/public/gollum/javascript/gollum.js.erb#L432). The options I see are:

* the current one
* getting the map of valid formats/extensions via an Ajax call to a new app route
* adding a  route in the app to which we POST the relevant form data (or add it to the query string for a GET route), and which returns the new location (or which just takes you there for a GET route)

bartkamphorst:
Out of the three I like the current strategy the most. 👍

dometto:
There was a fourth option: providing the correct extension to the JS by adding it to the editor format selector in the mustache template. 👍 No Ajax calls, but also no need to precompile when using custom formats. This also obviates the need for the extra requires in the Rakefile. See https://github.com/gollum/gollum/pull/1437

PR #1436, opened 2019-10-09
Show flashNotice instead of JS alert for (un)successful uploads.

bartkamphorst:
This PR adds visual feedback to successful uploads through the Upload button by showing a flash notice. The feedback for unsuccessful uploads is improved by also showing a flash notice instead of a JS alert.

This PR also improves the `flashNotice` helper in the following ways:
* flashNotice will replace an existing flash notice if one already exists.
* flashNotice will add an automatic fade-out to success notices. The fade-out will occur after 5 seconds.
* flashNotice now takes additional optional parameters to add an action button to the notice. This is currently unused, but may come in handy in the future. Example:
`flashNotice("warn", "Careful, you are about to ...", "Are you sure?", "alert('boo')", "danger")`
This will add a danger button to the flash notice with the label "Are you sure?" that will show an alert when clicked.

### DIFF
--- a/lib/gollum/public/gollum/javascript/gollum.js.erb
+++ b/lib/gollum/public/gollum/javascript/gollum.js.erb
@@ -102,10 +102,17 @@ function preparePage () {
   }
 }
 
-function flashNotice(type, notice) {
+function flashNotice(type, notice, button_label, button_function, button_type) {
   // accepted types: info, success, warn, error
-  html = '<p><div class="flash flash-' + type +'"><button class="flash-close js-flash-close" type="button" onclick="parentNode.remove()"><%=rocticon('x')%></button>' + notice + '</div></p>';
+  nested_button_html = '';
+  if (typeof(button_label) !== 'undefined'  && typeof(button_function) !== 'undefined') {
+    button_type = (typeof(button_type) !== 'undefined' && button_type == 'danger') ? ' btn-danger' : '';
+    nested_button_html = '<span class="px-2"><button type="button" class="btn btn-sm' + button_type + '" onclick="' + button_function + '">' + button_label + '</button></span>'
+  }
+  html = '<p><div id="gollum-flash" class="flash flash-' + type +'"><button class="flash-close js-flash-close" type="button" onclick="parentNode.remove()"><%=rocticon('x')%></button>' + notice + nested_button_html + '</div></p>';
+  $('#gollum-flash').remove();
   $('#wiki-content h1').before(html);
+  if (type == 'success') { setTimeout(function() {$('#gollum-flash').fadeOut();}, 5000); }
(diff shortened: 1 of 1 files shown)


lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -102,10 +102,17 @@ function preparePage () {
...
+function flashNotice(type, notice, button_label, button_function, button_type) {
   // accepted types: info, success, warn, error
-  html = '<p><div class="flash flash-' + type +'"><button class="flash-close js-flash-close" type="button" onclick="parentNode.remove()"><%=rocticon('x')%></button>' + notice + '</div></p>';
+  nested_button_html = '';
+  if (typeof(button_label) !== 'undefined'  && typeof(button_function) !== 'undefined') {
+    button_type = (typeof(button_type) !== 'undefined' && button_type == 'danger') ? ' btn-danger' : '';
+    nested_button_html = '<span class="px-2"><button type="button" class="btn btn-sm' + button_type + '" onclick="' + button_function + '">' + button_label + '</button></span>'
+  }
+  html = '<p><div id="gollum-flash" class="flash flash-' + type +'"><button class="flash-close js-flash-close" type="button" onclick="parentNode.remove()"><%=rocticon('x')%></button>' + notice + nested_button_html + '</div></p>';
+  $('#gollum-flash').remove();
```
dometto: Don't quite understand how this plays with the timeout below. Doesn't this just remove any earlier flash notices, potentially before the user could see them?

dometto: I'm especially worried about cases in which, say, a success notice might remove an error or warn notice

bartkamphorst: ` $('#gollum-flash').remove();` makes sure that there is no more than one flash notice at a time. So yes, if you would have two notices following each other in quick succession, the first one would be removed in favor of the second, more recent one.

> I'm especially worried about cases in which, say, a success notice might remove an error or warn notice

Here's the scenario in favor of this behavior. You upload a file, which turns out to already exist. You get an error. You upload your file again, under another name. This time the upload succeeds. The intuitive behavior, I think, is for the more recent success flash message to supersede the earlier error message.

bartkamphorst: Using one at a time is also in line with the Primer CSS style guide:

Screen Shot 2019-10-09 at 17 32 44

dometto: OK :)

dometto: Ok :)

PR #1443, opened 2019-11-14
Track path changes in History. Resolves #1441

### LINKED ISSUE #1441: 5.x Broken history and compare links for asciidoc files
Gollum version:
Gollum 5.0.1b

With an asciidoc page the history links `http://wiki/Home.md/c4b30797a2bd78d6165bb1d08772e2e0b3321105` and compare links are broken because they refer to `<page>.md` (file extension is always .md) and not to `<page>.asciidoc`, if i manually change the URL to `.asciidoc` everything works as expected.

dometto:
Requires https://github.com/gollum/rugged_adapter/pull/39

### DIFF
--- a/lib/gollum/views/compare.rb
+++ b/lib/gollum/views/compare.rb
@@ -20,6 +20,8 @@ module Precious
       def lines
         lines = []
         lines_to_parse = @diff.split("\n")[2..-1]
+        # If the diff is of a rename, the diff header will be one line longer than normal because it will contain a line starting with '+++' to indicate the 'new' filename.
+        # Make sure to skip that header line if it is present.
         lines_to_parse = lines_to_parse[1..-1] if lines_to_parse[0].start_with?('+++')
         lines_to_parse.each_with_index do |line, line_index|
           lines << { :line  => line,
--- a/lib/gollum/views/compare.rb
+++ b/lib/gollum/views/compare.rb
@@ -19,11 +19,10 @@ module Precious
 
       def lines
         lines = []
-        lines_to_parse = @diff.split("\n")[2..-1]
+        lines_to_parse = @diff.split("\n")[4..-1]
         # If the diff is of a rename, the diff header will be one line longer than normal because it will contain a line starting with '+++' to indicate the 'new' filename.
         # Make sure to skip that header line if it is present.
         lines_to_parse = lines_to_parse[1..-1] if lines_to_parse[0].start_with?('+++')
-        puts lines_to_parse.inspect
         lines_to_parse.each_with_index do |line, line_index|
           lines << { :line  => line,
                      :class => line_class(line),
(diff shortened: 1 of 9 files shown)


lib/gollum/views/history.rb
```diff
@@ -33,6 +34,13 @@ def versions
       def editable
         @editable
       end
+
+      private
+
+      def path_for_version(sha)
+        Gollum::PreviewPage.new(@wiki, @pathnames[sha], '', sha).escaped_url_path
```
dometto: A bit annoying that we have to instantiate a new PreviewPage object for each revision. Alternative would be to refactor `gollum-lib` in such a way that the `escaped_url_path` method depends on a class method that we could also invoke here.

bartkamphorst: I would prefer the second option. It seems wasteful to instantiate PreviewPage objects for what is basically a string manipulation operation.

lib/gollum/views/compare.rb
```diff
@@ -19,11 +19,13 @@ def after
 
       def lines
         lines = []
-        @diff.split("\n")[2..-1].each_with_index do |line, line_index|
+        lines_to_parse = @diff.split("\n")[2..-1]
+        lines_to_parse = lines_to_parse[1..-1] if lines_to_parse[0].start_with?('+++')
```
bartkamphorst: Why was this `starts_with('+++')` introduced?

lib/gollum/views/history.rb
dometto: Ok, agreed!

lib/gollum/views/compare.rb
dometto: Allowing diffing of files with different filenames in different revisions (which is the main purpose of this PR) introduced a new edge case: the patch 'header' would be one line longer to indicate the creation of a 'new' file. That header line starts with `+++`. We want to start parsing changes from the next line.

bartkamphorst: Good. Could you add a comment above this line to express this (e.g., `# skip optional creation header` or something)?

dometto:
Still getting some errors having to do with the new signature for the diff method

dometto:
Pushed changes to the adapters that solve the last issue, also encountered a small bug that's now fixed in this PR.

Small question concerning behavior to think about: in the History view, the order in which you select a commit determines whether it becomes the 'old sha' or the 'new sha' for the diff generation. We should make sure this order is respected when reverting a patch, or disallow reverting a patch that's the wrong way round. What do you think, @bartkamphorst?

bartkamphorst:
> We should make sure this order is respected when reverting a patch, or disallow reverting a patch that's the wrong way round. What do you think, @bartkamphorst?

Agreed.

dometto:
> disallow reverting a patch that's the wrong way round

This turns out to be the case.

PR #1445, opened 2019-12-15
Update asset dependencies. Resolves #1435

### LINKED ISSUE #1435: Update to Sprockets 4
See https://github.com/rails/sprockets/issues/565#event-2696215187

This will also allow migrating to `SassC`, as the old ruby `sass` gem is deprecated.

dometto:
This updates a number of asset-related dependencies, including a switch to `sassc`. It works, but downside: `sassc` breaks jruby support due to native extensions. We can choose to make `sassc` and `sprockets` development dependencies, so that jruby is only supported with precompiled assets. But then the jruby travis build would still fail. Alternatively, we have to hack conditional dependencies into gollum: `sass` for jruby, `sassc` for MRI.


ahorek:
what is your platform? sassc should support JRuby.

there're some known problems on Windows / MacOS. Maybe https://github.com/sass/sassc-ruby/pull/173 is related?

dometto:
@ahorek yes, when I tested I was on Windows. I later noticed `sassc` installs on JRuby on OS X without problems. We still haven't decided to upgrade to `sassc` yet, because:

* Default JRuby installations on Windows come without build tools, which means for most people a simple `gem install gollum` wouldn't suffice.
* Creating a multiplatform `.jar` won't be possible, because of the linking to platform-specific `libsass`.

It's a bit annoying that the only drop-in replacement for `sass` uses C-extensions, which (although FFI makes things a little better), inevitably complicates the goal to have one codebase for all platforms.

Alternatively, asset compilation could be moved out of the app itself (except test mode?) and into the `Rakefile` only, so that we don't need to include `sassc` as a hard dependency. That would mean running tests or compiling assets would not be possible on Windows out of the box (without the dev tools), but perhaps that's worth it for being able to keep our dependency chain up to date.

ahorek:
It is unfortunate that the entire ruby-sass ecosystem depends on a single maintainer. Hopefully a new release will be out soon.

PR #1475, opened 2020-03-13
Use tmpdir instead of tempfile for creating temporary repos in the test suite.

bartkamphorst:
Previously, the test suite was broken for older git versions because git (e.g. version 2.14.3 (Apple Git-98)) could not clone to Tempfiles. This PR fixes the issue by cloning to a tmpdir instead of a Tempfile.


bartkamphorst:
Failing tests are unrelated, so merging.

PR #1476, opened 2020-03-13
Update rake to mitigate CVE-2020-8130.

bartkamphorst:
We may also want to fix this in master, depending on when we think the 5.x release will be out.


dometto:
Ok with adding this only to 5.x

PR #1489, opened 2020-03-16
Added abilities for citation rendering in 5.x

TheCedarPrince:
Here, I finally updated the Gemfile to contain the gems for proper citation. 

Pinging @dometto for a review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,16 @@
 source 'https://rubygems.org'
 
 if RUBY_PLATFORM == 'java'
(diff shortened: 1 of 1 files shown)


dometto:
@TheCedarPrince thanks for submitting this! However, I probably wasn't clear enough in explaining how this works. Once we release gollum 5.0 (very, very soon now), you won't have to bother with the Gemfile at all anymore! You can then just:

```bash
gem install gollum pandoc-ruby bibtex-ruby citeproc-ruby csl-styles
```

...and all will work! Gemfile is only needed when running gollum from source, which is meant for development, testing, or experimenting, but not for production.

PR #1493, opened 2020-03-22
Package assets

dometto:


dometto:
A bit annoying that this adds some many files, but I don't see how, without this, we could run with static assets (which is waaay faster) out of the box. We could decide on a different location for the packaged assets, though.

dometto:
MathJax works with non-static assets, but is broken with static assets. @ProgramFan do you have any idea how we can get MathJax to play well with sprockets assets precompilation?

ProgramFan:
I don't know how the precompilation works. In the non-static asserts version, mathjax is served in a certain base URL and the base URL is configured in the page template (not sure which one, but you can always find them in the templates directory). I suspect the base URL may have to change in the case static asserts.

dometto:
@ProgramFan: thanks for the info. I decided that precompiling MathJax was an impossible task. Instead, I've added a special route for MathJax assets that serve them statically. (This route is even used when running gollum with `--no-static`, which applies only to golum's own assets.) This significantly speeds up loading times.

Also made some further small changes:
* Fixed MathJax on Preview pages
* Fixed the Rakefile task for precompilationl (was not using a sane path by default, assets are now dumped into `lib/gollum/public/assets` -- that ok @bartkamphorst ?).

Lots of file changes, but that's due to having moved the MathJax folder. :)

dometto:
@bartkamphorst notice that this PR sets the default environment when running gollum to production (and so you'll get the static assets by default). That means that if you want to test changes to JS/CSS, you have to do `GOLLUM_ENV=test bundle exec gollum`. Is that acceptable? I don't see another way for people to run gollum in production out of the box otherwise (if that's what we want)

dometto:
> @bartkamphorst notice that this PR sets the default environment when running gollum to production (and so you'll get the static assets by default). That means that if you want to test changes to JS/CSS, you have to do GOLLUM_ENV=test bundle exec gollum. 

Using `RACK_ENV` instead of `GOLLUM_ENV`, as that is the standard for rack apps (and the tests were already using `RACK_ENV=test`).

bin/gollum
```diff
@@ -197,8 +197,8 @@ end
 # Read command line options into `options` hash
 begin
   opts.parse!
-rescue OptionParser::InvalidOption
-  puts 'gollum: #{$!.message}'
+rescue OptionParser::InvalidOption => e
+  puts "gollum: #{$!.message}"
```
bartkamphorst: If you're capturing the error as `e`, you might as well use `e` instead of `$!`.

lib/gollum/app.rb
```diff
@@ -110,14 +111,17 @@ class App < Sinatra::Base
     end
 
     namespace '/gollum' do
+      
+      get '/assets/mathjax/*' do
+        env['PATH_INFO'].sub!("/#{Precious::Assets::ASSET_URL}/mathjax", '')
```
bartkamphorst: Will `sub!` always succeed, even if mathjax is not enabled?

lib/gollum/templates/editor.mustache
```diff
@@ -1,4 +1,4 @@
-<div id="gollum-editor" data-escaped-name="{{escaped_name}}" class="{{#is_create_page}}create{{/is_create_page}}{{#is_edit_page}}edit{{/is_edit_page}} {{#allow_uploads}}uploads-allowed{{/allow_uploads}}">
+<div id="gollum-editor" data-escaped-name="{{escaped_name}}" class="{{#is_create_page}}create{{/is_create_page}}{{#is_edit_page}}edit{{/is_edit_page}} {{#allow_uploads}}uploads-allowed{{/allow_uploads}} tex2jax_ignore">
```
bartkamphorst: What does this `tex2jax_ignore` addition do?

dometto: I had to add the MathJax JS to the editor, but that caused a small glitch as MathJax would try to render math in the gollum editor (which didn't succeed, but looked weird). :) `tex2jax_ignore` causes MathJax to ignore text in this element.

bartkamphorst: Ok.

lib/gollum/app.rb
dometto: Yes, it's only replacing out '/gollum/assets/mathjax' from the PATH_INFO. Come to think of it, I might just put in the string `'/gollum/assets/mathjax'` instead of `"/#{Precious::Assets::ASSET_URL}/mathjax"`?

bartkamphorst: Sure, makes sense.

dometto:
Incorporated your feedback and added a test for the mathjax route, will merge if the tests pass!

PR #1497, opened 2020-03-24
Update migration script

dometto:
Support the hyphenate option in the migration script, too. Will require minor changes after https://github.com/gollum/gollum-lib/pull/365 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
 	* May be written in a variety of [markups](#markups).
 	* Can be edited with your favourite system editor or IDE (changes will be visible after committing) or with the built-in web interface.
 	* Can be displayed in all versions, reverted, etc.
+* Gollum strives to be compatible with GitHub wikis (see `--hyphened_tag_lookup`)
 * Gollum supports advanced functionality like:
   * [UML diagrams](https://github.com/gollum/gollum/wiki#plantuml-diagrams)
   * [BibTeX and Citation support](https://github.com/gollum/gollum/wiki/BibTeX-and-Citations)
@@ -124,6 +125,7 @@ Gollum comes with the following command line options:
 | --template-page   | none      | Use _Template in root as a template for new pages. Must be committed. |
 | --emoji           | none      | Parse and interpret emoji tags (e.g. `:heart:`) except when the leading colon is backslashed (e.g. `\:heart:`). |
 | --global-tag-lookup | none   |  Match an internal link to 'Foo' with the first page found with that filename, anywhere in the repository. Provides compatibility with Gollum 4.x. |
+| --hyphened-tag-lookup | none  | Match an internal link to 'Bilbo Baggins' with 'Bilbo-Baggins'. Provides compatibility with Gollum 4.x. |
 | --help            | none      | Display the list of options on the command line. |
 | --version         | none      | Display the current version of Gollum. |
 | --versions        | none      | Display the current version of Gollum and auxiliary gems. |
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
   * Annotations using [CriticMarkup](https://github.com/gollum/gollum/wiki#criticmarkup-annotations)
   * Mathematics via [MathJax](https://github.com/gollum/gollum/wiki#mathematics)
   * [Macros](https://github.com/gollum/gollum/wiki/Standard-Macros)
-	* [Redirects](https://github.com/gollum/gollum/wiki/5.0-release-notes#support-for-redirects)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -17,6 +17,7 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
 	* May be written in a variety of [markups](#markups).
 	* Can be edited with your favourite system editor or IDE (changes will be visible after committing) or with the built-in web interface.
 	* Can be displayed in all versions, reverted, etc.
+* Gollum strives to be compatible with GitHub wikis (see `--hyphened_tag_lookup`)
```
dometto: Ok to mention github here, or do we have an alternative?

bartkamphorst: Seems fair, since it is one of the original selling points of gollum!

bin/gollum-migrate-tags
```diff
@@ -6,6 +6,24 @@ require 'rubygems'
 
 REPO = ARGV[0] || Dir.pwd
 
+def set_const(const, val)
+  already_defined = begin
+    Object.const_get(const)
+    true
```
bartkamphorst: Need help understanding this `true` here.

bin/gollum-migrate-tags
```diff
@@ -6,6 +6,24 @@ require 'rubygems'
...
+  rescue NameError
+    false
+  end
+  Object.const_set(const, val) unless already_defined
+end
+
+def const_true(const)
```
bartkamphorst: Not sure I follow this logic at all I'm afraid. Need some context.

bin/gollum-migrate-tags
```diff
@@ -189,16 +218,21 @@ class ::Gollum::Filter::TagMigrator < Gollum::Filter::Tags
   private
   
   def tag_for_pick(pick, orig_tag, extra, anchor, linking_page_path)
-    pick = if defined?(PREFER_RELATIVE)
+    pick = if const_true(:PREFER_RELATIVE)
       overlapping_path = Pathname.new(linking_page_path).dirname.to_s
-      relative_path = pick.to_s.match(/^\/#{overlapping_path}\/(.+)/)
+      if overlapping_path == '.'
```
bartkamphorst: Ternary would work here well I think:
```ruby
overlapping_path = overlapping_path == '.' ? '' : ::File.join('/', overlapping_path)
```

bin/gollum-migrate-tags
```diff
@@ -189,16 +218,21 @@ class ::Gollum::Filter::TagMigrator < Gollum::Filter::Tags
...
+        overlapping_path = '' 
+      else
+        overlapping_path = ::File.join('/', overlapping_path)
+      end
+      relative_path = pick.to_s.match(/^#{overlapping_path}\/(.+)/)
```
bartkamphorst: Did the Pathname solution I proposed not work properly?

test/test_migrate.rb
```diff
@@ -36,13 +38,30 @@
...
+      RUN_SILENT = true
+      NO_DRY_RUN = true
+      PREFER_RELATIVE = true
+      HYPHENATE = true
+    
+      Dir.chdir(@path) do
+        load script_path
+      end
+    
+      f = File.new(::File.join(@path, 'Home.textile'), 'r')
```
bartkamphorst: `f = ::File.new`.

bin/gollum-migrate-tags
dometto: if `const` is set to `false`, this block would return `false`. But we want to know if the constant has already been defined or not -- not whether the constant evaluates as `true` or `false`.

Motivation: I'm writing a substitute for `defined?` here, as `const` has to be a `String` or `Symbol` here, and `defined?(:some_symbol)` is always true. I wanted a way to not constantly have to do this in the main code: `CONST = bla unless defined?(CONST)`. But on reflection, maybe that's less ugly than this.

The real problem is the use of constant to set application options, but apart from global's it's the only thing I could think of that would allow us to access the options from within the subclassed Filter classes.

bin/gollum-migrate-tags
dometto: I wanted to avoid having to do `bla if defined?(CONST) && CONST` all the time. But again, that may be better than this metacode.

bin/gollum-migrate-tags
dometto: I couldn't get it to work sadly. The problem is `pick` has to start with a `/`, and so is absolute. But `linking_page_path` is relative. This triggers an `ArgumentError` in `Pathname.relative_path_from`. Making `linking_page_path` absolute as well has the consequence that the result of `relative_path_from` will also be absolute. Making `pick` relative as well means you sometimes end up with paths like `../Foo/Bar.md` (that is, *all* generated links become relative to root).

bin/gollum-migrate-tags
dometto: Found a nicer way to do all this :)

PR #1500, opened 2020-03-26
Add RSS Feed

dometto:
Couldn't stand that we were one of the only wikis without an RSS feed on https://en.wikipedia.org/wiki/Comparison_of_wiki_software

Some pointers in the line comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
   * Mathematics via [MathJax](https://github.com/gollum/gollum/wiki#mathematics)
   * [Macros](https://github.com/gollum/gollum/wiki/Standard-Macros)
   * [Redirects](https://github.com/gollum/gollum/wiki#redirects)
-	* [RSS Feed](https://github.com/gollum/gollum/wiki/5.0-release-notes#rss-feed) of latest changes
+  * [RSS Feed](https://github.com/gollum/gollum/wiki/5.0-release-notes#rss-feed) of latest changes
   * ...and [more](https://github.com/gollum/gollum/wiki)
 
 ### SYSTEM REQUIREMENTS
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
   * Annotations using [CriticMarkup](https://github.com/gollum/gollum/wiki#criticmarkup-annotations)
   * Mathematics via [MathJax](https://github.com/gollum/gollum/wiki#mathematics)
   * [Macros](https://github.com/gollum/gollum/wiki/Standard-Macros)
-  * [Redirects](https://github.com/gollum/gollum/wiki/5.0-release-notes#support-for-redirects)
+  * [Redirects](https://github.com/gollum/gollum/wiki#redirects)
+	* [RSS Feed](https://github.com/gollum/gollum/wiki/5.0-release-notes#rss-feed) of latest changes
   * ...and [more](https://github.com/gollum/gollum/wiki)
 
 ### SYSTEM REQUIREMENTS
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -10,6 +10,7 @@ require 'sprockets-helpers'
 require 'octicons'
 require 'sass'
 require 'pathname'
+require 'rss'
 
 require 'gollum'
 require 'gollum/assets'
@@ -44,6 +45,8 @@ module Precious
     register Mustache::Sinatra
     register Sinatra::Namespace
     include Precious::Helpers
+    include Precious::Views::AppHelpers
+    
     dir = File.dirname(File.expand_path(__FILE__))
 
     set :sprockets, ::Precious::Assets.sprockets(dir)
@@ -77,6 +80,8 @@ module Precious
       @critic_markup = settings.wiki_options[:critic_markup]
       @redirects_enabled = settings.wiki_options.fetch(:redirects_enabled, true)
       @per_page_uploads = settings.wiki_options[:per_page_uploads]
+      
+      @wiki_title = settings.wiki_options.fetch(:title, 'Gollum Wiki')
 
       forbid unless @allow_editing || request.request_method == 'GET'
       Precious::App.set(:mustache, {:templates => settings.wiki_options[:template_dir]}) if settings.wiki_options[:template_dir]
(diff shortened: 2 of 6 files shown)


lib/gollum/app.rb
```diff
@@ -44,6 +45,8 @@ class App < Sinatra::Base
     register Mustache::Sinatra
     register Sinatra::Namespace
     include Precious::Helpers
+    include Precious::Views::AppHelpers
```
dometto: needed the `extract_page_dir(file)` helper below. It's a bit ugly to include view helpers here. Ideally, the feed generation would be part of a view. We could do so by registering our own Tilt templates. On the other hand, this works, and I don't know how big a deal it is.

lib/gollum/app.rb
```diff
@@ -111,7 +116,37 @@ class App < Sinatra::Base
     end
 
     namespace '/gollum' do
-      
+      get '/feed/' do
+        url = env['REQUEST_URI'] || "http#{env['HTTPS'] == 'off' ? nil : 's'}://#{env['SERVER_NAME']}#{env['PATH_INFO']}" # Annoyingly, env['REQUEST_URI'] is not available in the tests. In case it sometimes proves to be unavailable in production, fall back to joining together the various parts of the url.
+        url = url.match(/^(.*)gollum\/feed\/$/)[1]
+        latest_changes = "#{url}gollum/latest_changes"
+        changes = wiki_new.latest_changes(::Gollum::Page.log_pagination_options(per_page: 10, page_num: 0))
```
dometto: Now using a hard coded maximum of 10 commits. Could also use `settings.wiki_options.fetch(:pagination_count, 10)` as for the latest changes view.

lib/gollum/app.rb
```diff
@@ -111,7 +116,37 @@ class App < Sinatra::Base
...
+          changes.each do |change|
+            maker.items.new_item do |item|
+              item.link = latest_changes
+              item.title = change.message
+              item.updated = change.authored_date
+              id = change.id
+              files = change.stats.files.map do |files|
+                [files[:old_file], files[:new_file]].compact.map do |file|
+                  f = extract_page_dir(file)
+                  "<li><a href=\"#{url}#{f}/#{id}\">#{f}</a></li>"
```
dometto: Link to each file's URL at this specific revision.

dometto:
This is how it looks in the NetNewsWire RSS reader:

Screenshot 2020-03-26 at 12 49 51

lib/gollum/app.rb
```diff
@@ -77,6 +80,8 @@ class App < Sinatra::Base
       @critic_markup = settings.wiki_options[:critic_markup]
       @redirects_enabled = settings.wiki_options.fetch(:redirects_enabled, true)
       @per_page_uploads = settings.wiki_options[:per_page_uploads]
+      
+      @wiki_title = settings.wiki_options.fetch(:title, 'Gollum Wiki')
```
dometto: There is no switch for this in `bin/gollum`, but someone could use a `config.rb` to set `Precious::App.wiki_options[:title] = 'My Title`. That ok, or should we add a `--title` switch? Downside would be: the RSS feed is currently the only place where the title makes a difference.

bartkamphorst: > but someone could use a config.rb to set Precious::App.wiki_options[:title] = 'My Title'

Perfectly acceptable!

README.md
```diff
@@ -24,7 +24,8 @@ Gollum is a simple wiki system built on top of Git. A Gollum Wiki is simply a gi
   * Annotations using [CriticMarkup](https://github.com/gollum/gollum/wiki#criticmarkup-annotations)
   * Mathematics via [MathJax](https://github.com/gollum/gollum/wiki#mathematics)
   * [Macros](https://github.com/gollum/gollum/wiki/Standard-Macros)
-  * [Redirects](https://github.com/gollum/gollum/wiki/5.0-release-notes#support-for-redirects)
+  * [Redirects](https://github.com/gollum/gollum/wiki#redirects)
+	* [RSS Feed](https://github.com/gollum/gollum/wiki/5.0-release-notes#rss-feed) of latest changes
```
bartkamphorst: Strange indentation.

lib/gollum/app.rb
bartkamphorst: Why would the view need to be a Tilt template?

lib/gollum/app.rb
bartkamphorst: Perfectly acceptable!

lib/gollum/app.rb
bartkamphorst: I'm not a fan of magic numbers, so the fetching `:pagination_count` has my vote.

lib/gollum/app.rb
```diff
@@ -111,7 +116,37 @@ class App < Sinatra::Base
     end
 
     namespace '/gollum' do
-      
+      get '/feed/' do
+        url = env['REQUEST_URI'] || "http#{env['HTTPS'] == 'off' ? nil : 's'}://#{env['SERVER_NAME']}#{env['PATH_INFO']}" # Annoyingly, env['REQUEST_URI'] is not available in the tests. In case it sometimes proves to be unavailable in production, fall back to joining together the various parts of the url.
```
bartkamphorst: I would leave out `Annoyingly`.

bartkamphorst: Instead of `nil` I would prefer `''` since you're working with Strings.

bartkamphorst: Perhaps `"#{env['HTTPS'] == 'off' ? 'http' : 'https'}` is more readable?

lib/gollum/app.rb
bartkamphorst: This would work correctly with `page-file-dir` and `base-dir`?

lib/gollum/app.rb
dometto: The route would then just look like this:

```ruby
get '/feed'
@changes = wiki_new.latest_changes(...)
rss :index
```

And all the logic for generating the actual content would be moved to the view.

lib/gollum/app.rb
dometto: Yes:

1. `extract_page_dir` handles the page_file_dir.
2. `base_dir` should be correctly incorporated into `url` on line 121.

lib/gollum/app.rb
dometto: yes, thanks!

dometto:
Separated the view logic from the route logic in a rudimentary way.

retifrav:
Perhaps I was looking in the wrong place, but I couldn't find the RSS feed in my Gollum. I tried `my.domain/gollum/feed` and `my.domain/feed` - both result in creating a new page.

Do I need to enable RSS feed somewhere in the `config.rb` first?

I have Gollum **5.1.1**.

bartkamphorst:
> Perhaps I was looking in the wrong place, but I couldn't find the RSS feed in my Gollum. I tried my.domain/gollum/feed and my.domain/feed - both result in creating a new page.

@retifrav Add a trailing slash to your url: `my.domain/gollum/feed/`. That should do the trick.

retifrav:
Oh, that was a silly overlook. Thank you!

PR #1504, opened 2020-03-27
Remove a tiny padding in the head div

heavywatal:
`px-1` class has been replaced with `pl-1` so that the right edge of the `div#wiki-wrapper` is aligned.

Current:
Screen Shot 2020-03-27 at 16 49 01

This PR:
Screen Shot 2020-03-27 at 16 49 44


bartkamphorst:
Thanks @heavywatal !

PR #1512, opened 2020-03-29
Restore --bare option. Resolves #1511

### LINKED ISSUE #1511: Citation rendering fails with bare repository (rugged adaptor)
Rendering citations in a bare repository stops working with commit f6f81b39, due to the switch to the rugged adaptor and removal of the --bare option.

dometto:


bin/gollum
```diff
@@ -61,6 +61,9 @@ opts = OptionParser.new do |opts|
   opts.on('-r', '--ref [REF]', 'Specify the branch to serve. Default: \'master\'.') do |ref|
     wiki_options[:ref] = ref
   end
+  opts.on("--bare", "Declare '<git-repo>' to be bare. This is only necessary when using the grit adapter.") do
```
bartkamphorst: @dometto Does #1511 show that `This is only necessary when using the grit adapter.` is false?

dometto: It does. Dang, went too fast -- that notice needs to be removed.

dometto: I've now fixed this on my local 5.x branch that will be merged into `master`.

bartkamphorst: Excellent.

PR #1513, opened 2020-03-29
Add Rhino to dependencies

dometto:
Fortunately, it installs on MRI, so we don't need to make it a conditional dependency.


gollum.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.md LICENSE]
 
-  s.add_dependency 'gollum-lib', '~> 5.0.a'
+  s.add_dependency 'gollum-lib', '~> 5.0'
```
dometto: Also update gollum-lib dependency

PR #1517, opened 2020-03-29
Actually package the sprockets manifest

dometto:
Caught a problem when testing the new release with a local `gem build ...; gem install ...`: sprockets had to precompile the assets on every first page load. Turns out the sprockets manifest wasn't packaged with the gem because it was being gitignored..


bartkamphorst:
Good catch!

PR #1520, opened 2020-03-30
Let Primer.CSS determine H1-H6 styling (fixes #1515).

### LINKED ISSUE #1515: [5.x, CSS] H2 and H3 headings have the same size (1.5em)
H2 and H3 looks indistinguishable to an unfamiliar reader. Try this, for a comparison:

```
# H1
## H2
### H3
#### H4
##### H5
###### H6
```

In Firefox's debug view, I can see that H2 and H3 are indeed the same size – 1.5em.

I think the faint underline on H2 is good as extra styling, but not sufficient as the only difference:

* What you expect is a size difference. What the unfamiliar reader sees, when looking at a H2, is a H3 with a `<hr>` underneath.
* You can't see the difference from 3m away.
* The faint underline is not preserved after pasting into Word (yes, my teammate had enough of markdown).

bartkamphorst:
We now use the same font sizes for H1-H6 as Github does:

Screen Shot 2020-03-30 at 09 15 38
Screen Shot 2020-03-30 at 09 15 20

### DIFF
--- a/lib/gollum/public/gollum/stylesheets/template.scss.erb
+++ b/lib/gollum/public/gollum/stylesheets/template.scss.erb
@@ -167,20 +167,15 @@ a {
   }
 
   h3 {
-    font-size: 1.5em;
   }
 
   h4 {
-    font-size: 1.2em;
   }
 
   h5 {
-    font-size: 1em;
   }
 
   h6 {
-    color: #777;
-    font-size: 1em;
   }
 
   p, blockquote, ul, ol, dl, table, pre {


dometto:
Great, less code on our side! looks fine to me, but why are you leaving the empty CSS definitions for h1-6 in place?

bartkamphorst:
> Great, less code on our side! looks fine to me, but why are you leaving the empty CSS definitions for h1-6 in place?

Just to indicate the proper place for future modifications, should those be needed. It also signals explicitly that gollum is not adding any styling. But I can remove them if you wish!

dometto:
> Just to indicate the proper place for future modifications, should those be needed. It also signals explicitly that gollum is not adding any styling. But I can remove them if you wish!

Nope that actually makes sense! Was just worried not having them in was magically breaking something, which would have been a sign of something Bad. :)

PR #1522, opened 2020-03-30
Improve CSS styling on 5.x using Primer.CSS. Fixes 1515, 1516, 1518

bartkamphorst:

PR #1525, opened 2020-03-30
Update static assets

dometto:
After changes in #1522


dometto:
Steps:

1. `git rm -r lib/gollum/public/assets`
1. `bundle exec rake precompile`
1. `git add lib/gollum/public/assets`
1. `git commit`

PR #1535, opened 2020-04-01
Check page.nil? before page.wiki

heavywatal:
So that /history/wrong-name gets redirected to / successfully.

### DIFF
--- a/test/test_app.rb
+++ b/test/test_app.rb
@@ -647,6 +647,11 @@ EOF
     assert_match /meta name="robots" content="noindex, nofollow"/, last_response.body
   end
 
+  test 'history/NO-EXIST redirects to Home' do
+    get '/gollum/history/NO-EXIST'
+    assert_equal last_response.status, 302
+  end
+
   def app
     Precious::App
   end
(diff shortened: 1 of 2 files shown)


bartkamphorst:
Would you mind adding a regression test for this PR as well?

test/test_app.rb
```diff
@@ -647,6 +647,11 @@ def nfd utf8
     assert_match /meta name="robots" content="noindex, nofollow"/, last_response.body
   end
 
+  test 'history/NO-EXIST redirects to Home' do
+    get '/gollum/history/NO-EXIST'
+    assert_equal last_response.status, 302
```
dometto, line 652: Perhaps better:

```ruby
follow_redirect!
assert last_response.ok?
assert_equal last_response.path, '/Home' # Not sure about .path here, but you get the idea
```

dometto:
Thanks for all the bug finds and fixes, @heavywatal!

test/test_app.rb
heavywatal: Thanks. I have just pushed a fix accordingly.

PR #1536, opened 2020-04-01
Replace redirect_to() with 'redirect to()`

heavywatal:
Because `redirect_to()` does not respect `base_path` while `redirect to()` does.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -188,8 +188,8 @@ module Precious
               @etag         = page.sha
               mustache :edit
           else
-            newfile = URI.encode(::File.join(@path, @name).sub(%r{^/}, ''))
-            redirect to("/gollum/create/#{newfile}")
+            newfile = ::File.join('gollum/create', @path, @name)
+            redirect to(encodeURIComponent(newfile).gsub('%2F', '/'))
           end
         end
 
@@ -436,12 +436,12 @@ module Precious
       end
 
       post '/compare/*' do
-        @file     = URI.encode(params[:splat].first)
+        @file     = encodeURIComponent(params[:splat].first).gsub('%2F', '/')
         @versions = params[:versions] || []
         if @versions.size < 2
-          redirect to("/gollum/history/#{@file}")
+          redirect to("gollum/history/#{@file}")
         else
-          redirect to("/gollum/compare/%s/%s...%s" % [
+          redirect to("gollum/compare/%s/%s...%s" % [
               @file,
               @versions.last,
               @versions.first]
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -188,7 +188,8 @@ module Precious
               @etag         = page.sha
               mustache :edit
           else
-            redirect to("/gollum/create/#{@name}")
+            newfile = URI.encode(::File.join(@path, @name).sub(%r{^/}, ''))
+            redirect to("/gollum/create/#{newfile}")
           end
         end
 
@@ -435,7 +436,7 @@ module Precious
       end
 
       post '/compare/*' do
-        @file     = params[:splat].first
+        @file     = URI.encode(params[:splat].first)
         @versions = params[:versions] || []
         if @versions.size < 2
           redirect to("/gollum/history/#{@file}")
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -189,7 +189,7 @@ module Precious
               mustache :edit
           else
             newfile = ::File.join('gollum/create', @path, @name)
-            redirect to(encodeURIComponent(newfile).gsub('%2F', '/'))
+            redirect to(encodeURIComponent(newfile))
           end
         end
 
@@ -436,7 +436,7 @@ module Precious
       end
 
       post '/compare/*' do
-        @file     = encodeURIComponent(params[:splat].first).gsub('%2F', '/')
+        @file     = encodeURIComponent(params[:splat].first)
         @versions = params[:versions] || []
         if @versions.size < 2
           redirect to("gollum/history/#{@file}")
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -188,7 +188,7 @@ module Precious
               @etag         = page.sha
               mustache :edit
           else
-            redirect_to("/create/#{encodeURIComponent(@name)}")
+            redirect to("/#{encodeURIComponent(@name)}")
           end
         end
 
@@ -438,9 +438,9 @@ module Precious
         @file     = encodeURIComponent(params[:splat].first)
         @versions = params[:versions] || []
         if @versions.size < 2
-          redirect_to("/history/#{@file}")
+          redirect to("/gollum/history/#{@file}")
         else
-          redirect_to("/compare/%s/%s...%s" % [
+          redirect to("/gollum/compare/%s/%s...%s" % [
               @file,
               @versions.last,
               @versions.first]
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -188,7 +188,7 @@ module Precious
               @etag         = page.sha
               mustache :edit
           else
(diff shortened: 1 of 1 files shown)


lib/gollum/app.rb
```diff
@@ -188,7 +188,7 @@ class App < Sinatra::Base
               @etag         = page.sha
               mustache :edit
           else
-            redirect_to("/create/#{encodeURIComponent(@name)}")
```
heavywatal, line 191: I removed `/create` because gollum tries to create a `no-exist` file **within the `/create` directory** when it is redirected to `/create/no-exist`. Redirecting to `/no-exist` lets gollum do the job. Is it possible to follow two-step redirection in the test?

https://github.com/gollum/gollum/blob/edc7d0b50bdb4a6477ad1bccef7651af293d2e94/test/test_app.rb#L304-L310

heavywatal: My bad. Redirecting to `/gollun/create/no-exist` is the way to go.

dometto: Understood. Are you managing to create a test guarding against the, I assume, bug with base-path?

lib/gollum/app.rb
```diff
@@ -435,12 +435,12 @@ class App < Sinatra::Base
       end
 
       post '/compare/*' do
-        @file     = encodeURIComponent(params[:splat].first)
```
dometto, line 438: Can you explain why `encodeURIComponent` is no longer necessary? Do we still need to use it elsewhere?

heavywatal: I am still struggling to understand this. First, I replaced `redirect_to` with `redirect to` because the latter respects `base_path` as in the other parts in the program. Then I found that `encodeURIComponent` converts `/` to `%2F` in a URL path, which makes the address bar very ugly. But just removing it was a bad idea, sorry, because it is necessary for example when a path contains whitespace and so on. Now I found that ruby's standard `URI.encode()` makes better results in most cases: slashes are left as they are, spaces are converted to `%20`. Why is `encodeURIComponent` implemented in gollum?

heavywatal: Now I found that `URI.encode` is obsolete.
https://ruby-doc.org/stdlib-2.7.0/libdoc/uri/rdoc/URI/Escape.html

dometto: > Now I found that URI.encode is obsolete.

Well, that settles it -- but perhaps the original reason for not using it was that `URI.encode` only allows encoding entire, properly formed URIs? Not sure about that.  But anyway, maybe we can use this: https://apidock.com/ruby/ERB/Util/url_encode ?

I think that may allow us to remove the entire `uri_encode_component.rb` file

heavywatal: That sounds really nice. But shall we make that another task? I will revert the things here to some ugly-but-works state to stick to the redirection problem.

dometto: 👍

dometto:
I see two more instances of `redirect_to` rather than `redirect to` in `app.rb`. Are they affected by the same bug?

heavywatal:
Yes, and this PR addresses all the three `redirect_to` and the preceding `encodeURIComponent`.

lib/gollum/app.rb
```diff
@@ -435,12 +436,12 @@ class App < Sinatra::Base
...
+        @file     = encodeURIComponent(params[:splat].first).gsub('%2F', '/')
```
dometto, line 439: This will mean the `%2F` will show up on the page if `@file` is used in the template -- that's why it's arguably better to take care of 'cleaning up' the URL at the point of the the redirect call. Maybe just leave that out completely for now, and figure out how to clean up the URLs in a separate PR where we tackle the issue of which encoding method to use?

lib/gollum/app.rb
```diff
@@ -188,7 +188,8 @@ class App < Sinatra::Base
               @etag         = page.sha
               mustache :edit
           else
-            redirect_to("/create/#{encodeURIComponent(@name)}")
+            newfile = ::File.join(@path, @name).sub(%r{^/}, '')
```
dometto, line 191: Not sure about the `.sub` here -- what if you're trying to create a page in a subdirectoy, e.g. `Subdir/Foo`? Is the fact that the path contains slashes causing a problem?

heavywatal: `Subdir/Foo` is fine because `%r{^/}` only matches a leading slash. The problem is `/Subdir/Foo`, which is redirected to `/gollum/create//Subdir/Foo`. See the next line: `to("gollum/create/#{encodeURIComponent(newfile)}")`. Removing `.sub` results in the following failure:
```
/Users/watal/fork/gollum/test/test_app.rb:308:in `block (2 levels) in <top (required)>'
<"/gollum/create/E"> expected but was
<"/gollum/create/%2FE">
```
https://github.com/gollum/gollum/blob/5a8750a9754f46d3bf76b66eb1e03051cbbb29a3/test/test_app.rb#L305-L308

heavywatal: If `::File.join(@path, @name)` always starts with `/`, then the following code is much simpler:
```rb
newfile = ::File.join(@path, @name)
redirect to("gollum/create#{encodeURIComponent(newfile)}")
```

heavywatal:
As for test, I guess a new context like "Frontend with base_path" needs to be added. If so, should I do it in this PR?

lib/gollum/app.rb
dometto: `::File.join(@path, @name)` only starts with a `/` if `@path` starts with a `/`. If `@path` can fail to contain a `/` (or can be `nil`), maybe the following works:

```ruby
redirect to("gollum/create#{::File.join('/', @path, @name)}")
```

Or alternatively, use [this helper](https://github.com/gollum/gollum/blob/master/lib/gollum/helpers.rb#L20). That does the same as you `sub`, it's just a little easier to maintain.

heavywatal: Perfect! It also indicates that replacing `%2F`with `/` has been considered safe and clean so far.

dometto:
> As for test, I guess a new context like "Frontend with base_path" needs to be added. If so, should I do it in this PR?

Yes please! This may not be non-trivial though -- have a look at how `base_path` is implemented [here](https://github.com/gollum/gollum/blob/master/bin/gollum#L290). If it's not possible to get this to work in the test context, then I would say too bad -- as long as the redirect targets are tested without base_path.

heavywatal:
Thank you for the primer to the implementation. But it makes almost no sense to me since actually I am quite new to Ruby and Rack. Is it related to the problem that `base_path` cannot be configured with `Precious::App.set(:wiki_options, {})` and the command line option `-b /base_path` is required? Its value seems to be overwritten here:

https://github.com/gollum/gollum/blob/5a8750a9754f46d3bf76b66eb1e03051cbbb29a3/lib/gollum/app.rb#L90-L91

dometto:
In each "Context", you'll see a method like this:

```ruby
  def app
    Precious::App
  end
```

Instead, for your new `context`, try this:

```ruby
  def app
    class MapGollum
      def initialize(base_path)
        @mg = Rack::Builder.new do

          map "/#{base_path}" do
            run Precious::App
          end
          map '/' do
            run Proc.new { [302, { 'Location' => "/#{base_path}" }, []] }
          end
          map '/*' do
            run Proc.new { [302, { 'Location' => "/#{base_path}" }, []] }
          end

        end
      end
  end
``` 

The `MapGollum` class is defined in `bin/gollum`, so it's not great to have to redefine it here. But if you get this test to work, I can refactor myself (requiring `MapGollum` in both places from the same source).

dometto:
@heavywatal I've added a test here: https://github.com/gollum/gollum/pull/1540

heavywatal:
Wow. I just started working on the test, but it would have been impossible for me to write one like yours. Excellent work! Thank you @dometto for your patience and effort.

PR #1538, opened 2020-04-02
Fix migrate script path argument. Resolves #1543

dometto:
Also fix tiny bug in `gollum --version`

### DIFF
--- a/bin/gollum-migrate-tags
+++ b/bin/gollum-migrate-tags
@@ -4,8 +4,6 @@ require 'optparse'
 require 'pathname'
 require 'rubygems'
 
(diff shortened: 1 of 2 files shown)


bin/gollum-migrate-tags
```diff
@@ -89,6 +87,8 @@ rescue OptionParser::InvalidOption
   exit
 end
 
+REPO = ARGV[0] || Dir.pwd
```
dometto, line 90: Doing this *after* optparser parses the options means that `ARGV[0]` is whatever argument *remains* that is not recognized by optparse.

PR #1539, opened 2020-04-03
Update layout.mustache

sam-340453:
Fix loading of MathJax when run with a base path.


dometto:
Dang, nice find! Thanks!

dometto:
Added a test in https://github.com/gollum/gollum/pull/1540

PR #1540, opened 2020-04-03
Add tests for base_path

dometto:
Add tests for https://github.com/gollum/gollum/pull/1536/files and https://github.com/gollum/gollum/pull/1539


dometto:
Had to move the definition of the `MapGollum` class to `lib/app.rb`.

PR #1543, opened 2020-04-05
Simulate 'git show' when a single version is posted from /history/

heavywatal:
This PR depends on https://github.com/gollum/rugged_adapter/pull/44 .


lib/gollum/app.rb
```diff
@@ -462,7 +462,12 @@ class App < Sinatra::Base
       post '/compare/*' do
         @file     = clean_url(encodeURIComponent(params[:splat].first))
         @versions = params[:versions] || []
-        if @versions.size < 2
+        if @versions.size == 1
+          wikip  = wiki_page(params[:splat].first)
+          commit = wikip.wiki.repo.rev_parse(@versions.first)
+          @versions.push commit.parent_ids.first
+        end
+        if @versions.empty?
```
dometto, line 470: This could be an `elsif`

lib/gollum/app.rb
```diff
@@ -462,7 +462,12 @@ class App < Sinatra::Base
       post '/compare/*' do
         @file     = clean_url(encodeURIComponent(params[:splat].first))
         @versions = params[:versions] || []
-        if @versions.size < 2
+        if @versions.size == 1
+          wikip  = wiki_page(params[:splat].first)
+          commit = wikip.wiki.repo.rev_parse(@versions.first)
```
dometto, line 467: Just for my understanding: the idea is to show the difference between `@version.first` and the latest commit?

lib/gollum/public/assets/.sprockets-manifest-3edc7e21737459b21fbf150f5dccd15c.json
```diff
@@ -1 +1 @@
...
...size":646,"digest":"31ae5d3282bbb8e7b7c3c9917e9fb68e3315a6b4a75da6cec48d21b8846905c4","integrity":"sha256-Ma5dMoK7uOe3w8mRfp+2jjMVprSnXabOxI0huIRpBcQ="},"print-512498c368be0d3fb1ba105dfa84289ae48380ec9fcbef948bd4e23b0b095bfb.css":{"logical_path":"print.css","mtime":"2020-03-30T23:45:16+09:00","size":75,"digest":"512498c368be0d3fb1ba105dfa84289ae48380ec9fcbef948bd4e23b0b095bfb","integrity":"sha256-USSYw2i+DT+xuhBd+oQomuSDgOyfy++Ui9TiOwsJW/s="},"app-3360f8faf2f9cd3020463278cdb5512b0109241647bbf55407e64ea11759572d.js":{"logical_path":"app.js","mtime":"2020-04-02T20:28:13+09:00","size":136035,"digest":"3360f8faf2f9cd3020463278cdb5512b0109241647bbf55407e64ea11759572d","integrity":"sha256-M2D4+vL5zTAgRjJ4zbVRKwEJJBZHu/VUB+ZOoRdZVy0="}},"assets":{"app.js":"app-3360f8faf2f9cd3020463278cdb5512b0109241647bbf55407e64ea11759572d.js","editor.js":"editor-0dbae28c53db5a73d666d30504946671eb09948ffdaf1fdd9d1c8d856d4c0ba1.js","app.css":"app-b205e593a30f1cc0054e2e9ed9fc8af3658d8ef4a62b9708c20f204560deefb7.css","criticmarkup.css":"criticmarkup-31ae5d3282bbb8e7b7c3c9917e9fb68e3315a6b4a75da6cec48d21b8846905c4.css","print.css":"print-512498c368be0d3fb1ba105dfa84289ae48380ec9fcbef948bd4e23b0b095bfb.css"}}
```
dometto, line 1: The new line is much longer than the old line -- it could be you still have multiple versions of `app.js` etc. in the static assets folder. I've found the best way to generate a new pack of static assets is to follow these steps:

`https://github.com/gollum/gollum/blob/master/CONTRIBUTING.md#updating-static-assets`

lib/gollum/app.rb
heavywatal: No. The difference between `@version.first` and its parent.

lib/gollum/app.rb
heavywatal: I don't think so. if `@version.size == 1`, the parent of `@version.first` is pushed to `@version` and, we want to enter the next `else` block. This will not happen with `elsif @versions.empty?` here.

lib/gollum/app.rb
dometto: Ah, right! What about the edge case where there is no parent?

heavywatal: Good catch. I didn't consider that case at all because it cannot happen in my repo with an empty commit at the beginning. As far as I know, this code fails if the page is created at the very first commit. Even in this case, users can get the same information by clicking the commit ID on the right in the Page History page. So, leaving the error will be an easy option, but redirecting to the commit page will be more user-friendly.

lib/gollum/public/assets/.sprockets-manifest-3edc7e21737459b21fbf150f5dccd15c.json
heavywatal: Thank you. Done.

lib/gollum/app.rb
```diff
@@ -464,8 +464,11 @@ class App < Sinatra::Base
         @versions = params[:versions] || []
         if @versions.size == 1
           wikip  = wiki_page(params[:splat].first)
-          commit = wikip.wiki.repo.rev_parse(@versions.first)
-          @versions.push commit.parent_ids.first
+          adapter_commit = wikip.wiki.repo.commit(@versions.first)
+          impl_commit = adapter_commit.commit
```
bartkamphorst, line 468: What is `impl` short for here?

heavywatal: implementation. It can be `Rugged::Commit` or `RJGit::Commit`. They happen to have `parents` method in common. Ideally this part should be hidden in adapter.

dometto: As discussed in https://github.com/gollum/rugged_adapter/pull/44 I think it would be better to implement the behavior in this PR using the existing gollum-lib API, and then find a separate way of exploiting the `rev_parse` efficiency improvement throughout the code base.

heavywatal: OK. What is the easiest way to find a parent commit with the current gollum-lib API?

heavywatal: FYI, this PR no longer depends on `rev_parse()` as of last commit.

dometto: > FYI, this PR no longer depends on rev_parse() as of last commit.

Ah, sorry, didn't realize that. It seems like you've found the only way of doing that, actually. In that case, we might want to abstract that logic to gollum-lib -- and/or add a `parent` method to the `Gollum::Git::Commit` class in the adapters. In any case, we want to limit direct use of adapter classes in gollum. What do you think, @bartkamphorst?

bartkamphorst: Agreed, gollum should be agnostic about adapters. Adding `Gollum::Git::Commit#parent` seems like the best way forward.

lib/gollum/app.rb
```diff
@@ -464,11 +464,8 @@ class App < Sinatra::Base
         @versions = params[:versions] || []
         if @versions.size == 1
           wikip  = wiki_page(params[:splat].first)
-          adapter_commit = wikip.wiki.repo.commit(@versions.first)
-          impl_commit = adapter_commit.commit
-          impl_parent = impl_commit.parents.first
-          adapter_parent = Gollum::Git::Commit.new(impl_parent)
-          @versions.push(adapter_parent.id)
+          commit = wikip.wiki.repo.commit(@versions.first)
+          @versions.push(commit.parent.id)
```
bartkamphorst, line 468: Can `#parent` return `nil`? If so, there might be cases where this statement would try to get the id for `NilClass`.

dometto: Good point. We need some `nil` checking :)

lib/gollum/app.rb
dometto: That's been implemented now thanks to @heavywatal

lib/gollum/app.rb
```diff
@@ -465,7 +465,12 @@ class App < Sinatra::Base
         if @versions.size == 1
           wikip  = wiki_page(params[:splat].first)
           commit = wikip.wiki.repo.commit(@versions.first)
-          @versions.push(commit.parent.id)
+          parent = commit.parent
+          if parent.nil?
+            redirect to("#{@file}/#{@commit.id}")
```
dometto, line 470: Just to make sure -- this is the method that takes account of the base path, right? If so, I think this is ready to merge!

heavywatal: Yes, it redirects to `/base_path/path/to/page/sha1`.

dometto:
I don't understand why the JRuby test is still failing. It's using version `0.5.2` of the rjgit adapter, which should have the new method..

dometto:
Go the tests to pass. Great stuff, thanks @heavywatal!

heavywatal:
Thank you for your support and patience! What was the cause of the Jruby failure, just for future reference, cache or something?

dometto:
I somehow released a version which did not yet contain the new method. Not sure how it happened, maybe the wrong `.gem` file was pushed accidentally. I just yanked the 'bad' release from rubygems and did a new release, that fixed it

PR #1547, opened 2020-04-08
Fix a selector in flashNotice()

heavywatal:
The original selector `#wiki-content h1` matches multiple times when the page content has h1 headers, resulting in multiple notice in a page, and what is worse, `$('#gollum-flash').fadeOut()` works only for the first one. This PR makes `flashNotice()` create only one banner at the top of wiki-content.


bartkamphorst:
Makes perfect sense! Thanks @heavywatal !

PR #1549, opened 2020-04-09
Fix `data-file-path` in overview pages

yuttie:
This correct the `data-file-path` attribute's value so that the value includes the ancestor directories to a file.

This is a fix for a bug that, for example, deleting `abc/def.md` actually deletes `def.md`.

### DIFF
--- a/lib/gollum/templates/overview.mustache
+++ b/lib/gollum/templates/overview.mustache
@@ -18,7 +18,7 @@
 						<span class="pr-2">{{{icon}}}</span>
 						<span><a href={{url}}>{{name}}</a></span>
 						{{#allow_editing}}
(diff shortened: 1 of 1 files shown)


yuttie:
I'm wondering whether also `data-confirm` attribute should be fixed or not...

dometto:
Thanks, nice catch! Wish we had a way of testing for these kinds of bugs 😕 

> I'm wondering whether also data-confirm attribute should be fixed or not...

Don't have a strong preference, either is fine with me!

bartkamphorst:
I think the `data-confirm` is fine as is. If you're browsing files in `Home / foo / bar` you and want to delete `File.md` you intuitively expect that to be `/foo/bar/File.md`. But nice catch on the bug @yuttie, especially since the DOM is updated as if the file was deleted. And thanks for the PR as well!

yuttie:
@bartkamphorst Thank you for merging my PR!  I'm glad to contribute such a nice app even though it is one-line fix :)

bartkamphorst:
@yuttie Absolutely; great to have you contributing. Let us know if you find anything else!

Issue #1339, opened 2018-10-19
5.x: utf-8 error thrown in Overview page (rugged)


dometto:
Rugged adapter: when there are pages in the wiki with utf-8 filenames:

<img width="1438" alt="screen shot 2018-10-20 at 00 52 33" src="https://user-images.githubusercontent.com/147111/47247304-237a2d00-d403-11e8-9a28-d167edbc2b1a.png">

I do not really understand why this is happening, but the cause appears to be [here](https://github.com/gollum/gollum/commit/3b9af34e0dacf6980eccccb3d82da269a43c62cb#diff-c3330a4adf9600567f00ccb96269d088R60), in the recent changes to the pages/overview view.

bartkamphorst:
I don't see this locally (RJGit). Does it happen with any utf-8 characters? Do you mind adding `--versions` info?

dometto:
Sorry, should've mentioned that it's rugged-specific in the body, too

dometto:
Resolved in https://github.com/gollum/gollum/pull/1346 and https://github.com/gollum/gollum-lib/pull/307

PR #1550, opened 2020-04-10
Fix folder URL escaping in overview

flowerysong:
The overview currently doesn't do any escaping for folder URLs, which can result in incorrect HTML like ```<a href=/gollum/overview/Mail Routing/test subdirectory/>test subdirectory</a>```

This change escapes these URLs the same way file URLs are escaped, turning the above example into ```<a href=/gollum/overview/Mail%20Routing/test%20subdirectory/>test subdirectory</a>```

### DIFF
--- a/lib/gollum/views/overview.rb
+++ b/lib/gollum/views/overview.rb
@@ -49,6 +49,7 @@ module Precious
               # result contains a folder
               folder_name = result_path.split('/').first
               folder_path = @path ? "#{@path}/#{folder_name}" : folder_name
+              folder_path = ERB::Util.url_encode(folder_path).gsub('%2F', '/').force_encoding('utf-8')
               folder_url  = "#{overview_path}/#{folder_path}/"
               files_and_folders << {name: folder_name, icon: rocticon('file-directory'), type: 'dir', url: folder_url, is_file: false}
             elsif result_path != '.gitkeep'


dometto:
Thanks for catching this!

> This change escapes these URLs the same way file URLs are escaped

Can you point us towards where the file URLs are being escaped in this way? I'm wondering if this would merit it's own helper method.

bartkamphorst:
Yes, good catch, but I'm not terribly excited about the `.force_encoding('utf-8')` in the fix. Do we really need this? Also, why escape `folder_path` but not `folder_url`?

flowerysong:
File URL escaping is in https://github.com/gollum/gollum-lib/blob/3163c3638eacba9b63cd9e977f96cb6c7d407089/lib/gollum-lib/file.rb#L120

The `force_encoding` is necessary because of #1339 

`folder_path` is only used to construct `folder_url`, and the rest of `folder_url` should already be escaped when we get it from the `overview_path` helper so we don't want to introduce double escaping. (From a practical standpoint it should never matter in this case, but it's more correct.)

dometto:
Sorry for the wait @flowerysong. Not sure if we have access to these helpers in the views, but would it be an idea to use:

```ruby
clean_url(encodeURIComponent(path))
```
Like [here](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L216)

dometto:
Ping @flowerysong

tpoliaw:
I had a look at this before realising there was already a PR for it. I think an alternative fix could be to quote the url in the `overview.mustache` template. Mustache already escapes variables by default.

heavywatal:
Oops, I've submitted #1581 after seeing #1572 without seeing this thread. I think the main problem here is missing double-quotes in `overview.mustache`.

flowerysong:
@dometto I don't really care how it's fixed, nor do I know enough about the framework you're using to have a useful opinion on what's correct. This change makes the construction of folder URLs match the construction of file URLs, which was enough to fix my problem.

dometto:
I've merged #1581, thanks to @heavywatal. I'll close this then, please reopen if the issue persists.

PR #1551, opened 2020-04-12
Don't uglify JS in development env

svoop:
Leave JS human readable when running Gollum with `RACK_ENV=development`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -181,6 +181,7 @@ end
 
 desc 'Precompile assets'
 task :precompile do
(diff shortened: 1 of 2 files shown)


dometto:
Thanks, I like it! Does the check als succeed when you run `bundle exec rake precompile`?

svoop:
@dometto Good point! For the precompile task to work, it has to require the app as well. I've pushed an updated commit.

dometto:
Thanks!

PR #1555, opened 2020-04-16
Add quick access to diff of each commit in the history

heavywatal:
This PR provides `/commit/<SHA>` interface to show diffs of a given commit for each of the related files as `git show` and GitHub do. Links to such diffs are also added to the pages in `/history/` and `/last_changes/`.

Screen Shot 2020-04-15 at 13 29 17
Screen Shot 2020-04-16 at 21 28 37


lib/gollum/public/gollum/stylesheets/template.scss.erb
```diff
@@ -650,4 +650,8 @@ div.pagination a.disabled {
   .search-context li:nth-child(n+4) {
     display: none;
    }
-}
\ No newline at end of file
+}
+
+.commit .sha {
+  font-family: $font-console;
```
bartkamphorst, line 656: Could you check if `Primer.CSS` doesn't already have an alternative way of setting fonts? I really want to minimize `template.scss` and rely on Primer as much as possible. Thanks!

lib/gollum/views/commit.rb
```diff
@@ -0,0 +1,111 @@
...
+      def authored_date
+        @commit.authored_date
+      end
+
+      def message
+        @commit.message
+      end
+
+      def files
+        puts @diff
```
bartkamphorst, line 25: Debug statement?

lib/gollum/views/commit.rb
```diff
@@ -0,0 +1,111 @@
...
+          {
+            path: matched[0],
+            lines: lines(diff)
+          }
+        end
+      end
+
+      private
+
+      def lines(diff)
```
bartkamphorst, line 39: @dometto Do we want methods like this in `gollum` or preferably in `gollum-lib`?

lib/gollum/views/commit.rb
heavywatal: Oops. Removed in the next commit.

lib/gollum/public/gollum/stylesheets/template.scss.erb
heavywatal: Got it. I found `text-mono` class and reverted all the changes in assets.

bartkamphorst: Terrific, thanks!

lib/gollum/views/commit.rb
dometto: Fine to have them in `gollum` since we already have them [here](https://github.com/gollum/gollum/blob/master/lib/gollum/views/compare.rb). My main comment is thus that a) either these methods which are common to the `Commit` and `Compare` view should be in their own Module (like [here](https://github.com/gollum/gollum/blob/master/lib/gollum/views/has_user_icons.rb) for example), or the `Commit` view should just be a subclass of the `Compare` view.

dometto:
Nice feature!

lib/gollum/views/commit.rb
bartkamphorst: 👍 Yes, let's try to avoid code duplication.

heavywatal: Thank you for the suggestion. Now I learned `require_relative` and class inheritance in Ruby :)

lib/gollum/app.rb
```diff
@@ -506,6 +506,19 @@ class App < Sinatra::Base
...
       end
 
+      get %r{
+        /commit/  # match any URL beginning with /show/
+        (\w+)     # match the SHA1
+      }x do |version|
+        @version = version
+        wiki = wiki_new
+        @commit = wiki.repo.commit(version)
+        parent = @commit.parent
```
dometto, line 516: Is it possible for `@commit` to be `nil` at this point? In that case, we need another `nil`-check, unfortunately

lib/gollum/templates/latest_changes.mustache
```diff
@@ -8,20 +8,22 @@
...
 			<span class="flex-auto col-1 text-gray-light">{{date}}</span>
 			<span class="flex-auto col-7">{{message}}<br/>
 			{{#files}}
 			  <span class="flex-auto col-2">{{#renamed}}{{renamed}} -> {{/renamed}}<a href="{{link}}">{{file}}</a></span><br/>
 			{{/files}}
 			</span>
-			<span class="pl-4 float-right">[{{id7}}]</span>
-		</div>
+			<span class="pl-4 float-right">
+				<a href="{{base_url}}/gollum/commit/{{id}}" class="btn btn-outline text-mono">{{id7}}</a>
```
dometto, line 22: It would be nice to be able to use the `page_route` helper, which is defined [here](https://github.com/gollum/gollum/blob/master/lib/gollum/views/helpers.rb#L56), but I don't think it works in the mustache templates at the moment. :(

lib/gollum/views/compare.rb
```diff
@@ -17,9 +17,9 @@ def after
         @versions[1][0..6]
       end
 
-      def lines
+      def lines(diff = @diff)
```
dometto, line 20: Didn't know you could use an instance variable in a method definition like this! If it works, fine!

dometto:
@bartkamphorst do you think we should add this to a `5.1` release? Should we make a new `rc` or `5.1` branch?

lib/gollum/templates/latest_changes.mustache
heavywatal: We can use it in the corresponding `views/*.rb` to make a new variable and reference it from mustache.

lib/gollum/app.rb
heavywatal: What should gollum show to the users when some invalid sha1 is given?

lib/gollum/views/compare.rb
heavywatal: No problem. The original behavior of `/compare/` is conserved.

lib/gollum/app.rb
dometto: I think in that case a `mustache :error` would be appropriate, like [here](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L502-L503). Message could be something like `"Invalid commit: #{@version}"`.

lib/gollum/templates/latest_changes.mustache
dometto: That's maybe not a bad idea! @bartkamphorst what do you think about that development pattern? If it makes sense, we could gradually start doing that for all the places in the templates where we are now manually concatenating stuff to `base_path`.

dometto:
Sorry for the long wait -- I think this PR is ready to merge?

> @bartkamphorst do you think we should add this to a 5.1 release? Should we make a new rc or 5.1 branch?

Ping!

bartkamphorst:
> do you think we should add this to a 5.1 release? 

For me personally that would be fine, because this would be a shiny new feature for 5.1. @heavywatal may be waiting for the feature to be released sooner than 5.1 though, or not?

heavywatal:
It's fine with me because I am not relying on the tags and releases.

bartkamphorst:
OK! Let's "save" it for the upcoming feature release then. 👍

dometto:
Thanks @heavywatal! Great work.

About the use of `{{base_url}}` in template pages, @heavywatal suggsted:

> We can use it in the corresponding views/*.rb to make a new variable and reference it from mustache.

This seems like a nice solution. In fact, it seems that there are several places where we still reference `{{base_url}}` in the templates:

```
lib/gollum/templates/layout.mustache:	var baseUrl = '{{base_url}}';
lib/gollum/templates/layout.mustache:  <script type="text/javascript" src="{{base_url}}/{{mathjax_config}}"></script>
lib/gollum/templates/layout.mustache:  <script defer src="{{base_url}}/gollum/assets/mathjax/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
lib/gollum/templates/history.mustache:					<span class="pl-4 float-right">[<a href="{{base_url}}/{{filename}}/{{id}}" title="View commit">{{id7}}</a>]</span>
lib/gollum/templates/navbar.mustache:	    <a class="btn" href="{{base_url}}/">Home</a>
lib/gollum/templates/search.mustache:      <span class="text-bold"><a href="{{base_url}}/{{name}}">{{name}}</a></span>
```

@heavywatal would it be an idea to try and refactor these lines in the way you suggested?

heavywatal:
OK, I'll take a look and submit a PR for the refactoring maybe in this weekend.

Issue #1563, opened 2020-05-08
Faulty syntax highlighting


julianbarg:
**Describe the bug**
In editing mode, when there are two internal links in a paragraph, both internal links _and_ all the text between them will be highlighted in blue (rather than just the internal links themselves).

**To Reproduce**
Paste the following text into an entry in editing mode:
`We create one internal link, e.g., [[test|testpage#this-section]]. Then we write some text. After that text, there is another internal link, such as [[test 2|testpage#that-section]].`

**Expected behavior**
Will display (bold section is what is displayed in blue): "We create one internal link, e.g., **test. Then we write some text. After that text, there is another internal link, such as test 2**." Obviously, expected behavior would be for it to display: "We create one internal link, e.g., **test**. Then we write some text. After that text, there is another internal link, such as **test 2**."

**Screenshots**
![image](https://user-images.githubusercontent.com/40502144/81406237-351b4680-9107-11ea-87e5-14243b49423d.png)

**Environment Info**

Gollum 5.0.1
Running on: x86_64-linux with Ruby version 2.7.1
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.0.4
Markdown rendering gem: kramdown
Other renderering gems:
org-ruby 0.9.12

dometto:
Confirmed, thanks! Working on a fix.

dometto:
Fix in https://github.com/gollum/gollum/pull/1566

julianbarg:
Thanks for the quick fix!

PR #1566, opened 2020-05-14
Fix gollum link tag highlighting

dometto:
See https://github.com/gollum/gollum/issues/1563

The problem was that previously, the regex would be too greedy, and match everything until the final `]]` on the line. So the following string matched in its entirety: `[[link]] this is still highlighted]]`


dometto:
Or would it be better to only re-generate static assets when we do a new release?

bartkamphorst:
I think this is fine. 👍

PR #1576, opened 2020-06-05
Add auto save functionality

mivok:
As discussed in #527, this adds auto saving to gollum using local storage. It's a direct conversion of the code I posted there, but with jquery used instead of pure javascript, and the code added to the relevant pieces of gollum.editor.js.

### DIFF
--- a/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
+++ b/lib/gollum/public/gollum/javascript/editor/gollum.editor.js
@@ -51,9 +51,36 @@
       }).insertAfter(textarea);
       textarea.css('display', 'none');
 
+      // Create a div for the autosave status message
+      var autoSaveMessage = document.createElement('div');
+      $(autoSaveMessage).addClass('div position-fixed bottom-0 left-0');
+      $('body').append(autoSaveMessage);
+
+      var autoSaveTimer = null;
+
       // NOTE: This requires the page to have only one 'gollum-editor-body'.
       var editor = ace.edit(editDiv[0], {rtlText: true});
       window.ace_editor = editor;
+
+      // Check to see if we have any autosaved text and show a message to
+      // restore it if present.
+      var storageKey = 'gollum_autorecover_' + window.location;
+      var savedText = localStorage.getItem(storageKey);
+
+      if (savedText) {
+        var savedMsg = document.createElement('div');
+        var restoreButton = document.createElement('button');
+        $(restoreButton).addClass('btn btn-sm primary flash-action')
+          .text('Restore Text')
+          .click(function() {
+            editor.getSession().setValue(savedText);
+            $(savedMsg).remove()
+          });
+        $(savedMsg).addClass('flash flash-full')
+          .text('Autosaved text is available. Click the button to restore it.')
+          .append(restoreButton);
+        $('body').prepend(savedMsg);
+      }
 
       editor.setTheme("ace/theme/tomorrow");
       editor.setKeyboardHandler();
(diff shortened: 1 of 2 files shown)


lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -51,10 +51,37 @@
       }).insertAfter(textarea);
       textarea.css('display', 'none');
 
+      // Create a div for the autosave status message
```
bartkamphorst, line 54: I think this `div` should just go in the editor template.

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -51,10 +51,37 @@
...
+      var autoSaveMessage = document.createElement('div');
+      $(autoSaveMessage).addClass('div position-fixed bottom-0 left-0');
+      $('body').append(autoSaveMessage);
+
+      var autoSaveTimer = null;
+
       // NOTE: This requires the page to have only one 'gollum-editor-body'.
       var editor = ace.edit(editDiv[0], {rtlText: true});
```
bartkamphorst, line 62: @dometto Any way to get the current rtlText setting, rather than hardcoding it to true here?

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -51,10 +51,37 @@
...
       textarea.css('display', 'none');
 
+      // Create a div for the autosave status message
+      var autoSaveMessage = document.createElement('div');
+      $(autoSaveMessage).addClass('div position-fixed bottom-0 left-0');
+      $('body').append(autoSaveMessage);
+
+      var autoSaveTimer = null;
+
       // NOTE: This requires the page to have only one 'gollum-editor-body'.
```
bartkamphorst, line 61: There is only one gollum-editor-body, but there may be up to three more text fields for the `_Footer`, `_Header` and `_Sidebar`. How difficult would it be to add the content from those text fields to the local storage as well?

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -51,10 +51,37 @@
...
       var editor = ace.edit(editDiv[0], {rtlText: true});
       window.ace_editor = editor;
 
+      // Check to see if we have any autosaved text and show a message to
+      // restore it if present.
+      var storageKey = 'gollum_autorecover_' + window.location;
+      var savedText = localStorage.getItem(storageKey);
+
+      if (savedText) {
```
bartkamphorst, line 70: Ideally, I'd like to have this `div` and the `button` in the editor template as well, and just activate/unhide it when relevant. That way we don't have these CSS classes added in the JS.

mivok: That makes sense. I'll make the appropriate changes and push an update.

bartkamphorst: Thanks!

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
mivok: Probably not too difficult. I'll look into adding those too.

mivok:
I've just pushed the two requested changes.

The first moves the html elements to the mustache templates. Because I originally put the banner notification ("you have autosaved text to be restored") at the top right under the body element, and now it has to go next to the editor itself, the banner has been moved to be right above the editor window. There are probably a few places this could go though.

The second change adds autosaving for the subpages. Because the subpages don't use the ace editor and just use textareas, the events are slightly different, and I moved the autosave handling into a function so it can be called from both the ace and textarea event handlers. All element selection for the subpage textareas is done with jquery and so if they don't exist then nothing will happen. The same goes for loading and clearing text from local storage - it should work correctly if there isn't anything in local storage for the subpage key.

bartkamphorst:
Terrific work, thanks @mivok !

Issue #1546, opened 2020-04-06
Compare Revisions redirect to error message when Edit option is disabled


DanSman:
**Describe the bug**

#1545

> I have also another question about the editing mode. I disabled it and I can't compare versions without having the edit mode set on true. Is it possible to compare different versions without having the edit mode activated? Should I do another issue about it? I don't wanna bother you so much :)

**To Reproduce**
1. Add `  :allow_editing => false,` to your config.rb
2. Click on 'Page History'
3. Compare Revisions

**Expected behavior**
To see the different versions in comparison.

**Version**
Gollum 5.0.1
Running on: x86_64-linux with Ruby version 2.7.1
Using:
rugged 0.99.0
gollum-rugged_adapter 0.99.4
gollum-lib 5.0.3
Markdown rendering gem: github/markdown
Other renderering gems:
none

dometto:
The problem is [here](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L108), coupled with the fact that the History view is `POST`ing the revision data to the compare view. Not sure what the best way to fix is. Would be a bit ugly to hardcode an exception for the `compare` view on L108. Also I don't see a super easy way to make the compare view use GET instead (although this is certainly possible).

dometto:
Fixed in #1578 thanks to @tpoliaw! Many thanks

PR #1578, opened 2020-06-10
Let compare versions be used in no-edit mode 

tpoliaw:
First pass at letting compare work in non-edit mode - potential fix for #1546.

Based on comments in the issue, it made more sense for compare to be a get as nothing is changed on the server and this lets it past the forbidden check for POST requests. The routes have to be rearranged to allow the more specific regex to match the request before the general one accepts everything.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -459,7 +459,30 @@ module Precious
         mustache :latest_changes
       end
 
-      post '/compare/*' do
+      get %r{
+        /compare/ # match any URL beginning with /compare/
+        (.+)      # extract the full path (including any directories)
+        /         # match the final slash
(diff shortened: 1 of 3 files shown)


dometto:
Thanks @tpoliaw! This looks like a good fix to me. @heavywatal has recent experience working on the compare route -- any issues you can think of?

heavywatal:
Looks amazing to me. Quite reasonable. I also briefly checked this PR branch works well with my local instance.
Now two entries for `get /compare/` look like hanging fruits to be merged and refactored, but let's leave it to another PR.

dometto:
>  Now two entries for get /compare/ look like hanging fruits to be merged and refactored, but let's leave it to another PR.

Good point! But yes, I'll merge this for now.

PR #1580, opened 2020-06-14
Reduce direct use of base_url

heavywatal:
As proposed in #1555, `page_route()` and `clean_url()` have been enhanced and applied to replace raw `base_url`s in mustache templates and ruby sources.

There may be some more places to be refactored, but I want somebody to review the changes before it gets too difficult to track.

### DIFF
--- a/lib/gollum/helpers.rb
+++ b/lib/gollum/helpers.rb
@@ -19,7 +19,7 @@ module Precious
     # Remove all double slashes
     def clean_url(url)
       return url if url.nil?
-      url.gsub('%2F', '/').gsub(/^\/+/, '').gsub('//', '/')
+      url.gsub('%2F', '/').gsub(%r{/{2,}}, '/').gsub(%r{^/}, '')
     end
 
     def forbid(msg = "Forbidden. This wiki is set to no-edit mode.")
--- a/lib/gollum/templates/layout.mustache
+++ b/lib/gollum/templates/layout.mustache
@@ -14,7 +14,7 @@
 
   <script>
   var criticMarkup = '{{critic_markup}}';
-	var baseUrl = '{{base_url}}';
+	var baseUrl = '{{page_route}}';
 	var uploadDest = 'uploads';
 	var perPageUploads = '{{per_page_uploads}}';
 	if (perPageUploads == 'true') {
--- a/lib/gollum/templates/layout.mustache
+++ b/lib/gollum/templates/layout.mustache
@@ -8,7 +8,7 @@
   {{#sprockets_stylesheet_tag}}app{{/sprockets_stylesheet_tag}}
   {{#sprockets_stylesheet_tag}}print print{{/sprockets_stylesheet_tag}}
 
-  {{#css}}<link rel="stylesheet" type="text/css" href="{{custom_path}}/custom.css" media="all">{{/css}}
+  {{#css}}<link rel="stylesheet" type="text/css" href="{{custom_css}}" media="all">{{/css}}
   {{#noindex}}<meta name="robots" content="noindex, nofollow" />{{/noindex}}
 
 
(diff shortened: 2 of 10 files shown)


dometto:
Looks good on first sight but I'll have a closer look later , the devil is in the detail here :)

lib/gollum/helpers.rb
```diff
@@ -19,7 +19,7 @@ def strip_page_name(name)
     # Remove all double slashes
     def clean_url(url)
       return url if url.nil?
-      url.gsub('%2F', '/').gsub(/^\/+/, '').gsub('//', '/')
+      url.gsub('%2F', '/').gsub(%r{/{2,}}, '/').gsub(%r{^/}, '')
```
dometto, line 22: :+1:

lib/gollum/views/helpers.rb
```diff
@@ -53,8 +53,15 @@ def self.included(base)
         end
       end
 
-      def page_route(page)
-        "#{base_url}/#{page}".gsub(/\/{2,}/, '/') # remove double slashes
+      def page_route(page = nil)
+        clean_url(@base_url, page)
```
dometto, line 57: Nice!

lib/gollum/views/helpers.rb
```diff
@@ -53,8 +53,15 @@ def self.included(base)
...
+      end
+
+      def clean_url(*url)
+        url.compact!
```
dometto, line 61: Why is this `compact!` call necessary -- and if it is, shouldn't it happen after the `nil` check on the next line?

lib/gollum/templates/layout.mustache
```diff
@@ -8,13 +8,13 @@
...
 
-  {{#css}}<link rel="stylesheet" type="text/css" href="{{custom_path}}/custom.css" media="all">{{/css}}
+  {{#css}}<link rel="stylesheet" type="text/css" href="{{custom_css}}" media="all">{{/css}}
   {{#noindex}}<meta name="robots" content="noindex, nofollow" />{{/noindex}}
 
 
   <script>
   var criticMarkup = '{{critic_markup}}';
-	var baseUrl = '{{base_url}}';
+	var baseUrl = '{{page_route}}';
```
dometto, line 17: I find `page_route` here a bit confusing, as `base_url` is conceptually independent from pages. Could we keep the `base_url` helper, perhaps as an alias of `custom_path`, and use it here?

lib/gollum/views/overview.rb
```diff
@@ -52,7 +52,7 @@ def files_folders
               folder_url  = "#{overview_path}/#{folder_path}/"
               files_and_folders << {name: folder_name, icon: rocticon('file-directory'), type: 'dir', url: folder_url, is_file: false}
             elsif result_path != '.gitkeep'
-              file_url = "#{@base_url}/#{result.escaped_url_path}"
+              file_url = page_route(result.escaped_url_path)
```
dometto, line 55: Much nicer! 👍

lib/gollum/views/search.rb
```diff
@@ -15,6 +15,7 @@ def results
           end
           [a[:filename_count], a[:count]] <=> [b_filename_count, b_count]
         end.reverse.slice((@page_num - 1) * @max_count, @max_count)
+        sorted.each {|x| x[:href] = page_route(x[:name])}
```
dometto, line 18: Can't we add the `:href` attribute within the `sort` block above (e.g. immediately after line 14)? Not sure that would work, but if it does, it would be more efficient

lib/gollum/views/helpers.rb
heavywatal: Note that `url` is an array that receives variable-length arguments. For example, `clean_url(nil)` and `clean_url("base", nil)` result in `url = [nil]` and `url = ["base", nil]`, respectively. So `compact!` is enough, and the next line `return url if url.nil?` should be `return url if url.empty?` instead.

lib/gollum/templates/layout.mustache
heavywatal: > I find `page_route` here a bit confusing

Agreed. But I may have a better idea. We can replace the javascript code in the following lines with a template variable and remove `baseUrl`, i.e., `uploadDest + {{url_tail}}`.
https://github.com/gollum/gollum/blob/6e6d9d8c1fbeca072c2b7a14208d808757bd5b44/lib/gollum/templates/layout.mustache#L20-L22

lib/gollum/views/search.rb
heavywatal: I don't think so. Each element in an array can be referenced many times in the `sort` block.

lib/gollum/templates/layout.mustache
heavywatal: No, `baseUrl` seems to be widely used in `gollum.js.erb`. Let's just restore `base_url` for now.

heavywatal: Committed. I kept `custom_path` separate from `base_url` because some user might happen to be using it to customize the path to `custom.css` and `custom.js`.

dometto:
Thanks @heavywatal!

heavywatal:
Thank you for the review, @dometto !

Issue #1572, opened 2020-05-28
Overview links are broken for subdirectories with spaces


Fastidious:
**Describe the bug**
Pages can be created under a subdirectory with spaces, but the subdirectory link itself is broken.

**To Reproduce**
1. Edit a page, add ```[[test directory/testpage]]```.
2. Click on link, create page.
3. Go to Overview, and see the `test directory`. 
4. Only the `test` portion will be clickable.

**Expected behavior**
Spaces converted to `%20` or other, subdirectories in working state.

**Environment Info**
```
madhatter:~ fastidious$ gollum --version
Gollum 5.0.1
```

bartkamphorst:
@dometto I think I have a fix by encoding the `folder_path` variable [here](https://github.com/gollum/gollum/blob/master/lib/gollum/views/overview.rb#L52) like so:
```ruby
folder_url  = "#{overview_path}/#{ERB::Util.url_encode(folder_path)}/"
```
I just wonder if there isn't already a more elegant way to do this?

PR #1581, opened 2020-06-14
Address the issue #1572

heavywatal:
Address the issue #1572 just by adding the missing double-quotes in an `<a>` tag.


bartkamphorst:
Nice fix. I assume the `data-file-path={{url}}` [here](https://github.com/gollum/gollum/pull/1581/files#diff-5094a6394aeac0fa8487f6e8ef16eb76R21) should probably also be quoted then, correct @heavywatal ?

heavywatal:
Good catch!

dometto:
Nice catches!

Issue #1532, opened 2020-03-31
the gemspec file is missing an rdoc dependancy


greyltc:
[Bundling](https://bundler.io/) gollum doesn't work unless the rdoc dep is explicitly is included.

Here's the error when trying to run the bundle:
```
$ BUNDLE_GEMFILE=/opt/gollum/Gemfile bundle exec ruby /opt/gollum/bin/gollum
Traceback (most recent call last):
	11: from /opt/gollum/bin/gollum:7:in `<main>'
	10: from /opt/gollum/bin/gollum:7:in `require'
	 9: from /opt/gollum/lib/gollum.rb:8:in `<top (required)>'
	 8: from /opt/gollum/lib/gollum.rb:8:in `require'
	 7: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup.rb:10:in `<top (required)>'
	 6: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup.rb:25:in `<module:GitHub>'
	 5: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup.rb:101:in `<module:Markup>'
	 4: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup.rb:101:in `instance_eval'
	 3: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markups.rb:2:in `<module:Markup>'
	 2: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markups.rb:2:in `require'
	 1: from /opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup/rdoc.rb:2:in `<top (required)>'
/opt/gollum/vendor/bundle/ruby/2.7.0/gems/github-markup-3.0.4/lib/github/markup/rdoc.rb:2:in `require': cannot load such file -- rdoc (LoadError)
```

dometto:
Isn't rdoc standardly available on a ruby installation?

greyltc:
fixed by https://github.com/gollum/gollum/pull/1533

greyltc:
> Isn't rdoc standardly available on a ruby installation?

I don't know.  

[...]

PR #1582, opened 2020-06-15
Add rdoc dependency

andydna:
Given
  I clone gollum from: https://github.com/gollum/gollum.git

Running
  ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
  Fedora 32

When
  I follow the instructions at:
  https://github.com/gollum/gollum#running-from-source

Then
  It should Just Work

But
  $ bundle exec bin/gollum
  bundler: failed to load command: bin/gollum (bin/gollum)
  LoadError: cannot load such file -- rdoc

So
  I fixed it

But
  I don't know how to decide what version to put in the gemspec, so I didn't put any

### DIFF
--- a/gollum.gemspec
+++ b/gollum.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.md LICENSE]
 
+  s.add_dependency 'rdoc'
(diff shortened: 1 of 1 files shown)


bartkamphorst:
Thanks for chipping in! Unfortunately, your fix should be proposed with the gem actually requiring `rdoc`, which is `github/markup`. See https://github.com/gollum/gollum/issues/1532 for more.

Issue #1590, opened 2020-07-08
Home button is broken


plujon:
The home button in a vanilla instance of gollum now merely loads the same url.  It does not go "home".

I suspect https://github.com/gollum/gollum/pull/1580 .

Perhaps gollum should have a test that the home button works.

bartkamphorst:
@heavywatal Could you have a look too see if this issue with the home button is indeed related to #1580?

heavywatal:
Thank you for the report, and sorry for the inconvenience. I have just submitted a quick fix.

> Perhaps gollum should have a test that the home button works.

Agreed. Many parts in gollum seems untested.

dometto:
Just merged @heavywatal's fix. It'll be contained in the next release of gollum (or you can pull in the changes if you're running from source).

> Agreed. Many parts in gollum seems untested.

Yes, the problem is we currently don't have tests for the UI elements and Javascript. We can test basic route functionality, but to test what happens when you press a button, we would need to add tests using a framework like https://github.com/teamcapybara/capybara

dometto:
Fix released in 5.1

PR #1593, opened 2020-07-12
Address home button issue #1590

heavywatal:
A trailing slash is added. This fixes the behavior when `base_url` is `nil`.


dometto:
Thanks @heavywatal!

Issue #1584, opened 2020-06-25
error fetching images uploaded through the editor when the path contains spaces


acavalin:
**Describe the bug**
Uploaded image is not displayed because gollum can't find it and redirects to "create new page".

**To Reproduce**
1. Create a new page with spaces `/hello world/foo bar.md` and save it
2. Edit the page, drag&drop an image with spaces `jon doe.png`, save
3. Viewing the page you can see a broken picture
4. Copy image location and paste link in a new browser tab
5. You are being redirected to the creation of a new page

**Expected behavior**
When an image is uploaded, it has to be downloadable and thus viewable.

**Screenshots**
![sshot](https://i.ibb.co/wYQfD0H/sshot.png)

The page:
~~~markdown
nice page [[/uploads/hello%20world/foo%20bar/jon doe.png]]
~~~

* Image location: `/uploads/hello%20world/foo%20bar/jon%20doe.png`
* Redirected to: `/gollum/create/uploads/hello world/foo bar/jon doe.png`

**Environment Info**

~~~shell
# $GEM_HOME/bin/gollum --version
Gollum 5.0.1

# ruby -v
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]

# start command
thin start --ssl --ssl-key-file /app/ssl.key --ssl-cert-file /app/ssl.crt -p $APP_PORT -R /app/config.ru
~~~

~~~ruby
# config.ru
#!/usr/bin/env ruby
%w{ rubygems digest yaml gollum/app rack/ssl }.each{|l| require l }

USERS = YAML.load_file '/app/users.yml'
PASS  = USERS['default_password']

use Rack::SSL # force ssl

use Rack::Auth::Basic, "Accesso riservato" do |user, pass|
  Precious::App.set :logged_user, user if USERS.keys.include?(user.strip) && Digest::MD5.hexdigest(pass) == PASS
end

Precious::App.set :port          , ENV['APP_PORT'].to_i
Precious::App.set :gollum_path   , ENV['APP_DATA']
Precious::App.set :default_markup, :markdown
Precious::App.set :wiki_options  , {
  allow_editing:    true,       # --no-edit
  allow_uploads:    true,       # --allow-uploads
  critic_markup:    true,       # --critic-markup
  css:              true,       # --css => uses commited /custom.css
  emoji:            true,       # --emoji
  h1_title:         true,       # --h1-title
  js:               true,       # --js => uses commited /custom.js
[...]

PR #1595, opened 2020-07-12
Address the uploads issue #1584

heavywatal:
'%20' in filenames are translated back to ' ' in both editor and file creation.


lib/gollum/app.rb
```diff
@@ -230,14 +230,15 @@ class App < Sinatra::Base
...
 
         if wiki.per_page_uploads
-          # remove base_url and gollum/* subpath if necessary
-          dir = request.referer.
-                  sub(request.base_url, '').
-                  sub(/.*gollum\/[-\w]+\//, '')
+          dir = request.referer.sub(request.base_url, '')
           # remove base path if it is set
-          dir = dir.sub(wiki.base_path, '') if wiki.base_path
+          dir.sub!(/^#{wiki.base_path}/, '') if wiki.base_path
```
dometto, line 235: Why the added `'/'`? Does this work when `base_path` already ends with a slash?

heavywatal: `/this/` is not a string but a regex. With the leading `^`, I just wanted to ensure that only `base_path` at the beginning is replaced.

dometto: Aah sorry for misreading!

Issue #1602, opened 2020-08-01
use wrong place for mathjax.config.js


fanhongtao:
Beacuse  #1572 , I run Gollum by [RUNNING FROM SOURCE](https://github.com/gollum/gollum#running-from-source).

* When I write a page in the root directory, Gollum uses `mathjax.config.js` in the root.
* When I write a page in the a sub-directory, Gollum tries to use `mathjax.config.js` in that sub-directory. 

If I run Gollum 5.0.1, Gollum always uses `mathjax.config.js` in the root.

After read the git logs, I think the problem maybe introduced [Use page_route() for custom.css, custom.js, and MathJax ](https://github.com/gollum/gollum/commit/6e6d9d8c1fbeca072c2b7a14208d808757bd5b44#diff-ae57601ec9f43453506032bb4295c44d)

And I fix the problem on my computer by change `layout.mustache` from 

```html
<script type="text/javascript" src="{{mathjax_config}}"></script>
```

to 

```html
<script type="text/javascript" src="/{{mathjax_config}}"></script>
```

Don't know if it's appropriate.

 @dometto

heavywatal:
Thank you for the report. You are right. I have submitted a quick fix for this.

dometto:
Fixed by @heavywatal in #1604. Will do a gollum bugfix release soon

dometto:
Actually, fixing this required another change: https://github.com/gollum/gollum/pull/1605

dometto:
Released in 5.1

PR #1604, opened 2020-08-01
Ensure page_route starts with a leading slash

heavywatal:
It addresses the issue #1602.
It also solves the issue #1590 differently from #1593.

### DIFF
--- a/lib/gollum/views/helpers.rb
+++ b/lib/gollum/views/helpers.rb
@@ -54,9 +54,7 @@ module Precious
       end
 
       def page_route(page = nil)
-        base_url = @base_url.nil? ? '/' : @base_url
-        base_url.prepend('/') unless base_url.start_with?('/')
-        clean_url(base_url, page)
+        clean_url("/#{@base_url}", page)
       end
 
       def clean_url(*url)
--- a/lib/gollum/views/helpers.rb
+++ b/lib/gollum/views/helpers.rb
@@ -54,7 +54,9 @@ module Precious
       end
 
       def page_route(page = nil)
-        clean_url('/' + @base_url, page)
+        base_url = @base_url.nil? ? '/' : @base_url
+        base_url.prepend('/') unless base_url.start_with?('/')
+        clean_url(base_url, page)
       end
 
       def clean_url(*url)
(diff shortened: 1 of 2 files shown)


lib/gollum/views/helpers.rb
```diff
@@ -54,7 +54,9 @@ def self.included(base)
       end
 
       def page_route(page = nil)
-        clean_url(@base_url, page)
+        base_url = @base_url.nil? ? '/' : @base_url
+        base_url.prepend('/') unless base_url.start_with?('/')
+        clean_url(base_url, page)
```
dometto, line 59: Does this really yield a string that starts with `'/'`? It seems the problem is that `clean_url` [removes all leading slashes](https://github.com/gollum/gollum/blob/master/lib/gollum/helpers.rb#L18). In that case, we could do:

```ruby
clean_url(@base_url ? @base_url : '/', page).prepend('/')
```

Or (faster):

```ruby
"/#{clean_url(@base_url ? @base_url : '/', page)}"
```

But perhaps the thing to do is first to write a test that defines the correct behavior?

heavywatal: There are two `clean_url()` in gollum: one you mentioned and another just bellow `page_rouge()`.
https://github.com/gollum/gollum/blob/1d67565450679cb350937887eea608f48da57508/lib/gollum/views/helpers.rb#L56-L67
The latter one is used, and it just removes duplicated slashes. Maybe the function should be renamed. Or gollum should provide one and only way to handle `base_url` and `page_route` things, someday...

dometto: Aah, thanks for the explanation!

dometto: I would still like to avoid the `.prepend` call, as modifications to Strings are expensive in Ruby. What about simply:

```ruby
         clean_url("/#{@base_url}", page)
```

This would automatically handle the `nil` case:

```ruby
2.6.3 :001 > a = nil
 => nil 
2.6.3 :002 > "/#{a}"
 => "/" 
```

And since double slashes would be removed anyway by `clean_url`, I think we don't need to worry about the case in which `base_url.start_with?('/')`?

heavywatal: Great! This is a good lesson for me.

dometto:
Thanks @heavywatal!

PR #1608, opened 2020-08-11
contrib/automation/gollum-post: A script to post files to Gollum in a…

basking2:
…n automated way.

I use a version of this script to publish current documentation many CI/CD builds of various microservices into a common Gollumn instance. This allows the primary authors to document along side their code but distribute and publish in a centralized way.


dometto:
Thanks for the contribution @basking2!

PR #1612, opened 2020-08-25
Filter _Template content. Resolves #1603 and #1640.

### LINKED ISSUE #1603: Is it possible to use variables in _Template?
For example:

`_Template.md`

```markdown
Date: {{YYYY-MM-DD}}

# {{page_name}}

Body

...
```

bartkamphorst:
This is a proposed solution to the question in #1603 . Basically, we just allow users to create a `Gollum::TemplateFilter` in their `config.rb` and let them manipulate the contents of `_Template` however they want at runtime. What do you think about this approach @dometto? A very simple example of such a filter could be the following:
```ruby
module Gollum
  class TemplateFilter
    def self.filter(data)
      return data.gsub!('{{current_date}}', Time.now.strftime("%d/%m/%Y"))
    end
  end
end
```

(Where `_Template.md` would obviously need to include something like `Date: {{current_date}}`.)

### DIFF
--- a/lib/gollum.rb
+++ b/lib/gollum.rb
@@ -33,4 +33,25 @@ module Gollum
       super(message || "Cannot write #{@dir}/#{@attempted_path}, found #{@dir}/#{@existing_path}.")
     end
   end
+  
+  class TemplateFilter
+    @@filters = {}
+
+    def self.filters
+      @@filters
+    end
+
+    def self.add_filter(pattern, &replacement)
+      @@filters[pattern] = replacement
+    end
+
+    def self.apply_filters(data)
+      template = data
+      @@filters.each do |pattern, replacement|
+        template = template.gsub(pattern, replacement.call)
+      end
+      template
+    end
+  end
+
 end
--- a/lib/gollum.rb
+++ b/lib/gollum.rb
@@ -46,11 +46,10 @@ module Gollum
     end
 
     def self.apply_filters(data)
-      template = data
       @@filters.each do |pattern, replacement|
-        template = template.gsub(pattern, replacement.call)
+        data.gsub!(pattern, replacement.call)
       end
-      template
+      data
     end
   end
 
--- a/lib/gollum.rb
+++ b/lib/gollum.rb
@@ -37,10 +37,6 @@ module Gollum
   class TemplateFilter
     @@filters = {}
 
-    def self.filters
-      @@filters
-    end
-
     def self.add_filter(pattern, &replacement)
       @@filters[pattern] = replacement
     end
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -638,7 +638,7 @@ module Precious
 
     def load_template
       template_page = wiki_page('/_Template')
-      template_page = (templalate_page.page != nil) ? template_page.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
+      template_page = (template_page.page != nil) ? template_page.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
       template_page = Gollum::TemplateFilter.apply_filters(template_page) unless Gollum::TemplateFilter.filters.empty?
       template_page
     end
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -352,11 +352,11 @@ module Precious
 
       get '/create/*' do
         forbid unless @allow_editing
-        @template_page = load_template if settings.wiki_options[:template_page]
         wikip = wiki_page(params[:splat].first)
         @name = wikip.name
         @ext  = wikip.ext
         @path = wikip.path
+        @template_page = load_template(@path) if settings.wiki_options[:template_page]
         @allow_uploads = wikip.wiki.allow_uploads
         @upload_dest   = find_upload_dest(wikip.fullpath)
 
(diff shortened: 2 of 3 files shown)


dometto:
I like the way this works on the user side, and the customizability! See comment.

lib/gollum/app.rb
```diff
@@ -355,6 +355,9 @@ class App < Sinatra::Base
         if settings.wiki_options[:template_page] then
           temppage = wiki_page('/_Template')
           @template_page = (temppage.page != nil) ? temppage.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
+          if defined?(Gollum::TemplateFilter)
+            @template_page = Gollum::TemplateFilter.filter(@template_page)
```
dometto, line 359: Quite a lot happening here in the middle of a Sinatra route. Would it make sense to:

1. Change `TemplateFilter` to (e.g.) `TemplatePage`
2. Add the logic for checking for/loading the template data to the `TemplatePage` class (for instance, to the `.load` method)
3. Simply call `@template_page = TemplatePage.load` here

dometto: Also, another approach to the API could be the following. Perhaps it would be more user friendly? Not sure.

```ruby
TemplatePage.add_filter('{{current_date}}') do
  Time.now.to_s
end
```

The add filter method would do something like:

For example:
```ruby
def self.add_filter(pattern, &replacement)
  @@filters[pattern] = replacement
end
```
And then in the `filter` method:

```ruby
@@template_filter.each |pattern, replacement|
data.gsub!(pattern, replacement.call)
end
```

bartkamphorst:
I processed your comments and used your API proposal for the new version. Filters can now be added in `config.rb` like so:
```ruby
Gollum::TemplateFilter.add_filter('{{current_date}}') do
  Time.now.strftime("%d/%m/%Y")
end

Gollum::TemplateFilter.add_filter('{{current_time}}') do
  Time.now.strftime("%T")
end
```

lib/gollum.rb
```diff
@@ -33,4 +33,25 @@ def initialize(dir, existing, attempted, message = nil)
...
+    end
+
+    def self.add_filter(pattern, &replacement)
+      @@filters[pattern] = replacement
+    end
+
+    def self.apply_filters(data)
+      template = data
+      @@filters.each do |pattern, replacement|
+        template = template.gsub(pattern, replacement.call)
```
bartkamphorst, line 51: Note that `gsub!` would not work because `gsub!` returns `nil` when the call does not succeed (which could happen if a filter is defined but not included in `_Template`).

dometto:
Very cool! Looks super elegant now.

lib/gollum/app.rb
```diff
@@ -639,6 +636,13 @@ def show_file(file)
       end
     end
 
+    def load_template
+      template_page = wiki_page('/_Template')
+      template_page = (template_page.page != nil) ? template_page.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
+      template_page = Gollum::TemplateFilter.apply_filters(template_page) unless Gollum::TemplateFilter.filters.empty?
```
dometto, line 642: we could leave out the `unless` call here, but it's also fine like this.

bartkamphorst: Yes, you're right. In that case I can also leave out the `getter` method `:filters`, cleaning it up even further. 👍

bartkamphorst:
This latest commit adds support for relative templates (e.g., `subdir/_Template.md`), with fallback to `/_Template.{ext}`. Relates to #1640 . @dometto Please let me know if you think this behavior is desirable.

dometto:
I think this is a nifty feature, yes, and the implementation is nice and clean -- so by all means!

tevino:
Can't wait for the release

tevino:
Found no document about this nifty feature, I just documented it with an example here:

https://github.com/gollum/gollum/wiki/Home/_compare/257834c879687ac819a0993019acdedca60c0e48

@bartkamphorst @dometto PTAL as I barely understand the difference between a proc and a block, the example works, but I'm not sure if it's the best practice.

Thanks for the work!

tevino:
@bartkamphorst This may sound straightforward to you: is it possible to access `@page.name` within a template filter?

lib/gollum/app.rb
```diff
@@ -639,6 +636,11 @@ def show_file(file)
       end
     end
 
+    def load_template(path)
+      template_page = wiki_page(::File.join(path, '_Template')).page || wiki_page('/_Template').page
+      template_page ? Gollum::TemplateFilter.apply_filters(template_page.raw_data) : nil
```
tevino, line 641: @bartkamphorst @dometto How about passing `path` to `apply_filters` so that it could be accessed in the `replacement`?

This makes a filter like `{{page_name}}` possible.

dometto: This makes sense to me, if you're wiling to open a PR for this @tevino! What do you think @bartkamphorst?

dometto:
Thanks for adding the documentation @tevino!

lib/gollum/app.rb
tevino: @dometto I just submitted a PR #1818, please help to review it, thanks!

PR #1618, opened 2020-09-20
Don't escape page path. Resolves #1615

### LINKED ISSUE #1615: Pages with some non alpha numeric characters are not edited
**Describe the bug**
If you create a page name with non alpha numeric characters, you can not edit it afterwards. 

**To Reproduce**
1. Create a new page named  C++
2. Edit it afterwards, it will have no effect. You will only get the original page you created as if no edit happened.

**Environment Info**
Gollum 5.1.1
That used to be fine prior version 5.

dometto:
Strangely, it seems that `params[:page]` is already escaped -- perhaps by Sinatra? When you inspect `params` when you get the page `/Samwise%20Gamgee.md`, `params[:page]` will already be `Samwise Gamgee.md`. The fact that we were calling unescape on the already escaped string caused #1615: "C++.md" was being escaped into "C  .md".

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -590,7 +590,6 @@ module Precious
     end
 
     get '/*' do
-      puts params.inspect
       show_page_or_file(params[:splat].first)
     end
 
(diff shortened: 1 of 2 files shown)


bartkamphorst:
>  it seems that params[:page] is already escaped -- perhaps by Sinatra?

Do you mean decoded?

dometto:
> Do you mean decoded?

Yes, although the call we're doing (and that is superfluous) is `CGI.unescape`.

bartkamphorst (approved):
Looks good. 👍

PR #1619, opened 2020-09-20
Normalize preview data

### LINKED ISSUE #1617: YAML front matter doesn't work in editing preview
**Describe the bug**
When using the YAML front matter, the page preview shows a YAML parsing error.

**To Reproduce**
1. Create a new page or edit an existing page
2. Use YAML front matter (for example to set the page title)
3. Click the preview tab in the gollum editor
4. YAML parsing error is shown at the top of the preview
5. If changes are saved, the page is rendered as expected

**Expected behavior**
Page preview should look like the final rendering of the page, with YAML front matter taken into account.

**Screenshots**
**Editing the page:**
image

**Preview of page while editing:**
image

**View of page once saved:**
image

**Environment Info**
```sh
gollum@red-markdown2:~$ gollum --versions
Gollum 5.0.1
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.0.4
Other renderering gems:
org-ruby 0.9.12
```

bartkamphorst:
This fixes #1617 by ensuring that the data passed into the PreviewPage is normalized (removing any `\r` occurrences). 

On second thought, this is a "gollum-only" fix; the alternative would be to call `normalize` from within `gollum-lib`. Alternative fix is [here](https://github.com/gollum/gollum-lib/pull/386).

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -421,7 +421,7 @@ module Precious
       post '/preview' do
         wiki           = wiki_new
         @name          = params[:page] ? strip_page_name(CGI.unescape(params[:page])) : 'Preview'
-        @page          = wiki.preview_page(@name, params[:content], params[:format])
+        @page          = wiki.preview_page(@name, wiki.normalize(params[:content]), params[:format])
         ['sidebar', 'header', 'footer'].each do |subpage|
           @page.send("set_#{subpage}".to_sym, params[subpage]) if params[subpage]
         end
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -355,6 +355,9 @@ module Precious
         if settings.wiki_options[:template_page] then
           temppage = wiki_page('/_Template')
           @template_page = (temppage.page != nil) ? temppage.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
+          if defined?(Gollum::TemplateFilter)
+            @template_page = Gollum::TemplateFilter.filter(@template_page)
+          end
         end
         wikip = wiki_page(params[:splat].first)
         @name = wikip.name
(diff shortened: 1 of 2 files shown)


lib/gollum/app.rb
```diff
@@ -355,6 +355,9 @@ class App < Sinatra::Base
         if settings.wiki_options[:template_page] then
           temppage = wiki_page('/_Template')
           @template_page = (temppage.page != nil) ? temppage.page.raw_data : 'Template page option is set, but no /_Template page is present or committed.'
+          if defined?(Gollum::TemplateFilter)
+            @template_page = Gollum::TemplateFilter.filter(@template_page)
```
dometto, line 359: How does this relate to normalization?

bartkamphorst: Shoot, remnant of other PR. 😨

dometto:
Think I prefer this fix to the gollum-lib one! I don't think it's the job of the `PreviewPage` class to ensure that the data that's passed into it fits a certain mould. We still need the gollum-lib PR for the typo though :)

lib/gollum/app.rb
bartkamphorst: But perhaps it doesn't matter, depending on whether you prefer this PR or the one in gollum-lib. :)

bartkamphorst:
> I don't think it's the job of the PreviewPage class to ensure that the data that's passed into it fits a certain mould.

Maybe not, but this fix uses a gollum-lib method that so far hasn't been used in gollum, only in gollum-lib.

dometto:
Ah, thanks for the explanation! In that case both PRs seem about equally good to me.. not sure what to decide, maybe the gollum-lib PR is less work (since we won't need to do a bugfix release of gollum)?

bartkamphorst:
Since gollum-lib already is calling `#normalize` in several places (e.g., `update_page` and `add_to_index`) I think the PR in gollum-lib makes the most sense. Closing this one.

PR #1634, opened 2020-12-05
Fix bug when link to a page doesn't work if it has ? in its title

ViChyavIn:


dometto:
Thanks for the pr!

lib/gollum/views/helpers.rb
```diff
@@ -61,7 +61,9 @@ def clean_url(*url)
         url.compact!
         return nil if url.empty?
 
-        ::File.join(*url).gsub(%r{/{2,}}, '/')
+        _url = ::File.join(*url)
+        _url = _url.gsub(%r{/{2,}}, '/')
```
dometto, line 65: The following would be faster:

```ruby
_url.gsub!(%r{/{2,}}, '/')
_url.gsub!(%r{\?}, '%3F')
_url
```

dometto (changes requested):
Small comment related to code performance, otherwise this looks good!

lib/gollum/views/helpers.rb
ViChyavIn: Done! https://github.com/ViChyavIn/gollum/commit/7643cd3b421bc0b4b19a27a5a583c76ae32c1b64

dometto:
Thanks!

bartkamphorst:
Do we have tests for these cases?

dometto:
@bartkamphorst no, forgot about that. @ViChyavIn any chance you could submit a regression test guarding against this bug? Sorry for the lateness of this request!

ViChyavIn:
@dometto Added test here: https://github.com/ViChyavIn/gollum/commit/87821d56fa2a9700dbf9bf90bfb24c4b44c2cadb . Should I open another pull request? And please review my test: it seems to work but it's my first time writing tests so I'm not sure if it's written properly.

dometto:
Thanks @ViChyavIn, yes that looks good! A PR would be very welcome.

ViChyavIn:
Done!

Issue #1624, opened 2020-11-06
Add support to extract user's name and email from HTTP headers (injected by authenticating reverse proxy)


pdinc-oss:
This feature request will allow F5, Apache HTTPD, etc reverse proxies to inject authoritative HTTP headers on every request to inform Gollum as to the identity of the remote user.

--enable-authentication [MODE]

Where mode HTTP_HEADER is expected

The headers will follow the convention described in https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

- GIT_AUTHOR_NAME is the human-readable name in the “author” field.
- GIT_AUTHOR_EMAIL is the email for the “author” field.
- GIT_COMMITTER_NAME sets the human name for the “committer” field.
- GIT_COMMITTER_EMAIL is the email address for the “committer” field.

Later the names can become configurable.

Please assign to us, patch will be forthcoming.

pdinc-oss:
- there really should be a header name override config option.
- it currently ignores the committer headers

pdinc-oss:
We deployed this to our production system, will provide feedback after a week of usage.

dometto:
hi @pdinc-oss, wondering if you've had any success with this. :)

pdinc-oss:
> hi @pdinc-oss, wondering if you've had any success with this. :)

yes, it is working quite well - no bugs in 16 days. Looking for feedback on it at this point.

bartkamphorst:
> Looking for feedback on it at this point.

Interesting approach! Looking forward to learning more about this.

pdinc-oss:
> > Looking for feedback on it at this point.
> 
> Interesting approach! Looking forward to learning more about this.

What can we help explain?

bartkamphorst:
Sorry, I meant to say that I'm awaiting your pull request. 😄

pdinc-oss:
> Sorry, I meant to say that I'm awaiting your pull request. 😄

https://github.com/gollum/gollum/pull/1636

dometto:
Closing in favor of discussion in #1636. @pdinc-oss it still seems to me like using omnigollum might be the best fit for your use case, but if there's something I'm missing feel free to let us know.

pdinc-oss:
We had to fork it, as we are only using gollum - no additional software.

PR #1636, opened 2020-12-10
initial draft solution for (#1624) extracting git credentials from HTT…

pdinc-oss:
…P headers

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -112,6 +112,12 @@ MSG
   opts.on('--no-edit', 'Disable the feature of editing pages.')  do
     wiki_options[:allow_editing] = false
   end
+  opts.on('--enable-authentication [MODE]', [:HTTP_HEADER], 'Enable authentication.',
+    'If set to \'HTTP_HEADER\', Gollum fetch author information from HTTP Headers.') do |mode|
+    wiki_options[:authentication]      = true
+    wiki_options[:authentication_mode] = mode
+    wiki_options[:allow_editing]       = false
+  end
   opts.on('--follow-renames', 'Follow pages across renames in the History view. Default: true.') do
     wiki_options[:follow_renames] = true
   end
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -97,8 +97,23 @@ module Precious
     end
 
     before do
+      @authentication = settings.wiki_options.fetch(:authentication, false)
+      @auth_editing_override = false
+      if @authentication
+        author  = session['gollum.author']
+        if author.nil? || author.empty?
+          if settings.wiki_options[:authentication_mode] == :HTTP_HEADER
+            unless request.env["HTTP_GIT_AUTHOR_NAME"].nil? || request.env["HTTP_GIT_AUTHOR_EMAIL"].nil? || request.env["HTTP_GIT_AUTHOR_NAME"].empty? || request.env["HTTP_GIT_AUTHOR_EMAIL"].empty?
+              session['gollum.author'] = { name: request.env["HTTP_GIT_AUTHOR_NAME"], email: request.env["HTTP_GIT_AUTHOR_EMAIL"] }
+              author  = session['gollum.author']
+              @auth_editing_override = true
+            end
+          end
+        end
+      end
+
       settings.wiki_options[:allow_editing] = settings.wiki_options.fetch(:allow_editing, true)
-      @allow_editing = settings.wiki_options[:allow_editing]
+      @allow_editing = settings.wiki_options[:allow_editing] || @auth_editing_override
       @critic_markup = settings.wiki_options[:critic_markup]
       @redirects_enabled = settings.wiki_options.fetch(:redirects_enabled, true)
       @per_page_uploads = settings.wiki_options[:per_page_uploads]


dometto:
Thanks for the PR @pdinc-oss. It looks like the actual authentication occurs outside of gollum. Probably your reverse proxy handles authentication, and sets the relevant HTTP headers upon success?

I am wondering whether it would not be better to implement this using a Rack middleware. Here's a minimal example of how that would look (tested, it works):

```ruby
  class Auth
    def initialize(app)
      @app = app
    end
    
    def call(env)
      if env['HTTP_GIT_AUTHOR_NAME'] && env['HTTP_GIT_AUTHOR_EMAIL']
        env['rack.session'] = {} unless env['rack.session']
        env['rack.session']['gollum.author'] = {
          name:  env['HTTP_GIT_AUTHOR_NAME']
          email: env['HTTP_GIT_AUTHOR_EMAIL']
        }
      end
      @app.call(env)
    end
  end
```

and then:


```ruby
  class App < Sinatra::Base
    use Auth # possibly conditional, use Auth if ... ?
    ...
```

This could easily be expanded so as to make the exact conditions under which to set the session variables (e.g. the name of the http headers) dynamic.

Besides being arguably simpler and more of a racky/ruby-y approach, it also has the advantage of disentangling the setting of committer name and email from the question of whether `@auth_editing_override` should be set to true. The user should be able to set the committer info without overriding that.

Let me know if that would be a good solution for you, and if you would still like to contribute a solution of that form.

pdinc-oss:
Honestly, it is a lot to digest, being unfamiliar with the technology stack. My instant thought goes towards usability - How would a systems administrator download Gollum and configure it?

dometto:
>  How would a systems administrator download Gollum and configure it?

By running gollum with `--config` and adding the code I pasted, or suitably altered code (it probably needs minor changes, e.g. wrapping the block in `module Precious`), to `config.rb`. The only thing that couldn't be accomplished in that way would be the `@auth_editing_override` logic. That would still require a PR that would:

1. Implement command line switch for `--enable-authentication` (no need for `MODE` I think)
2. Change [this line](@ https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L101) to something like:

```ruby
 @allow_editing = settings.wiki_options[:allow_editing] || (settings.wiki_options[:authentication] && session['gollum.author'])
```

pdinc-oss:
> > How would a systems administrator download Gollum and configure it?
> 
> By running gollum with `--config` and adding the code ...

This is where the issue we were trying to address starts. It is too much to expect of a SA for a common place method of A&A. A limited set of "authentication plugins" should be available out of the box.

> command line switch for `--enable-authentication` (no need for `MODE` I think)

The concept of `MODE` was in line with a standardization of the out of the box authentication capabilities. It would also encourage a disciplined, modular, and hopefully extensible, approach to doing so. Then things like this will have a reference starting point for those that need to extend the authentication options beyond the out of the box shipped modes.

> Here's a minimal example of how that would look (tested, it works):
> 
> ```ruby
>   class Auth
>     def initialize(app)
>       @app = app
>     end
>     
>     def call(env)
>       if env['HTTP_GIT_AUTHOR_NAME'] && env['HTTP_GIT_AUTHOR_EMAIL']
>         env['rack.session'] = {} unless env['rack.session']
>         env['rack.session']['gollum.author'] = {
>           name:  env['HTTP_GIT_AUTHOR_NAME']
>           email: env['HTTP_GIT_AUTHOR_EMAIL']
>         }
>       end
>       @app.call(env)
>     end
>   end
> ```
> 
> and then:
> 
> ```ruby
>   class App < Sinatra::Base
>     use Auth # possibly conditional, use Auth if ... ?
>     ...
> ```
> 
> This could easily be expanded so as to make the exact conditions under which to set the session variables (e.g. the name of the http headers) dynamic.

I am a big fan of doing things the right way or by convention. Without extensive formal documentation (showing the right ways) or examples (establishing convention) to reference - the new implementations will not comply.

I am willing to put in the effort to learn and and make the patch, where it accomplishes al of the principles above. It will be a slow and steep learning curve.

dometto:
>  A limited set of "authentication plugins" should be available out of the box.

Just to make sure you are aware, there is https://github.com/arr2036/omnigollum (also [Docker images](https://hub.docker.com/r/hotmaps/gollum) including omnigollum). I understand that many not cover all your requirements -- although it would be interesting to know why not.

pdinc-oss:
> there is https://github.com/arr2036/omnigollum (also [Docker images](https://hub.docker.com/r/hotmaps/gollum) including omnigollum). I understand that many not cover all your requirements -- although it would be interesting to know why not.

Do you know when is it targeted for merging into mainline gollum?

bartkamphorst:
> Do you know when is it targeted for merging into mainline gollum?

There are no plans to merge these two projects at this time.

dometto:
Closing this for now. Proposals for adding omnigollum/auth to the Docker image (or providing support for one) should be discussed at https://github.com/gollum/docker or at https://github.com/arr2036/omnigollum

pdinc-oss:
> Closing this for now. Proposals for adding omnigollum/auth to the Docker image (or providing support for one) should be discussed at https://github.com/gollum/docker or at https://github.com/arr2036/omnigollum

This pull request **_is not_** about merging the two projects. Please re-open with constructive requests for change.

dometto:
@pdinc-oss sorry for the confusion. I was thinking: omnigollum implements everything you need (in effect, omnigollum just *is* the middleware-based approach I set out in the comments above). If there is a feature you are missing, could you please explain what it is and its use case? Thanks!

PR #1637, opened 2020-12-10
Update gems

dometto:


Gemfile
```diff
@@ -1,10 +1,10 @@
 source 'https://rubygems.org'
-
+gem 'gollum-lib', :path => '~/Code/gollum_project/gollum-lib/'
```
bartkamphorst, line 2: Please remove this local `:path` to gollum-lib from the Gemfile.

gollum.gemspec
```diff
@@ -23,18 +23,18 @@ Gem::Specification.new do |s|
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = %w[README.md LICENSE]
 
-  s.add_dependency 'gollum-lib', '~> 5.0'
+#  s.add_dependency 'gollum-lib', '~> 5.0'
```
bartkamphorst, line 26: Why is this line commented out?

gollum.gemspec
```diff
@@ -23,18 +23,18 @@ Gem::Specification.new do |s|
...
   s.add_dependency 'kramdown', '~> 2.3'
   s.add_dependency 'kramdown-parser-gfm', '~> 1.1.0'
   s.add_dependency 'sinatra', '~> 2.0'
   s.add_dependency 'sinatra-contrib', '~> 2.0'
-  s.add_dependency 'mustache', ['>= 0.99.5', '< 1.0.0']
+  s.add_dependency 'mustache-sinatra', '~> 1.0'
```
bartkamphorst, line 31: Did this gem change names?

dometto: Not exactly, but the Sinatra helpers that we were using were abstracted to a separated gem (also maintained by the mustache org), which in turn requires `mustache`.

gollum.gemspec
dometto: That's part of requiring my local copy of gollum-lib (for the purpose of testing the octicons). Thanks for catching these!

bartkamphorst: I figured as much! :)

gollum.gemspec
bartkamphorst: 👍

Issue #1620, opened 2020-10-02
Delete file doesn't work


CrustyAuklet:
**Describe the bug**
Using the "overview" mode to view files, if I click the trash can icon to delete a file it initially appears to be removed but persists on page reload.

**To Reproduce**
I want to delete a file that has been uploaded. In our case the file is `uploads/sheet1.png`. Using the "overview" mode to see all the files, the I then navigate to the `uploads` folder. If I then press the "trash can" button next to the file to delete it, gollum asks if I am sure with a popup prompt. After clicking "yes" the file appears to be deleted. But if I then refresh the page, the file is still there.

If I ssh into the server and run `git log` in the gollum directory I see no activity related to the delete attempt.

**Expected behavior**
If I press the trash can icon next to the file it should be removed

**Screenshots**
![image](https://user-images.githubusercontent.com/9755578/94887253-b5ca3e80-042a-11eb-9418-00a5c0f5c36e.png)

**Environment Info**
```
gollum@red-markdown2:~$ gollum --versions
Gollum 5.0.1
Running on: x86_64-linux-gnu with Ruby version 2.7.0
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.0.4
Markdown rendering gem: kramdown
Other renderering gems:
org-ruby 0.9.12
```

dometto:
Hi @CrustyAuklet! I've tried but can't reproduce this. There's been a couple of patch-level updates to gollum-lib, and one patch-level update to gollum. Could you update and let us know if you still encounter this bug? If so, could you let us know:

1. with what arguments you're running gollum
1. whether this occurs with all files or only some
2. if the latter, which files?
  * either point us to an example repository
  * or tell us about the directory structure/filename of the files which display the problem

Thanks!

CrustyAuklet:
@dometto the bug still persists, we are running 5.1.2 now. 

It happens with any type of file, specifically for us PDFs and PNG files.

we have files in a single "uploads" folder, with all the markdown pages in the root wiki directory  
example:
```
~/wiki
├── uploads
│  ├── picture.png
│  └── document.pdf
├── WikiPage1.md
└── WikiPage2.md
```
  


we are running gollum using systemd, here is the systemd file and gollum settings file:
```
gollum@wiki-server:~$ cat /etc/systemd/system/gollum.service
[Unit]
Description=Gollum wiki server
After=network.target
After=syslog.target

[Service]
Type=simple
User=gollum
[...]

PR #1639, opened 2020-12-13
Don't add base path to file url for deletion. Resolves https://github.com/gollum/gollum/pull/1620

dometto:


lib/gollum/views/overview.rb
```diff
@@ -4,6 +4,7 @@ module Precious
   module Views
     class Overview < Layout
       attr_reader :results, :ref, :allow_editing, :newable
+      HIDDEN_PATHS = ['.gitkeep']
```
dometto, line 7: This has nothing to do with the issue at hand, but I came across this `'.gitkeep'` string hardcoded in the method below and thought we should perhaps abstract it. Look ok, @bartkamphorst?

PR #1643, opened 2020-12-15
Reset h2 counters. Resolves #1642

### LINKED ISSUE #1642: Missing css definition breaks a header numbering (proposed fix)
When using a header counting feature (header_enum: decimal) a weird behavior is observed.

* Multiple h2 headers have the same number assigned (e.g. "1.1. header2a", "1.1. header2b", etc)
* Level 3 headers are numbered incorrectly ("1.0.1 header3a","1.0.2 header3b" - note the zero digit in the middle).

Confirmed on firefox 83.0 (64 bit) on Linux.

I tracked this down to app-b205e593a30f1cc0054e2e9ed9fc8af3658d8ef4a62b9708c20f204560deefb7.css, which misses the definition of "h1{counter-reset:h2}.header-enum" (accodring to css definitions - body resets h1, h2 resets h3, h3 resets h4 and so on - but nothing resets the h2 counter).

After patching my local css-file with the missing definition, header numbering works as expected.

Could you please verify the problem and fix a css-file permanently ?

dometto:
Still need to run `bundle exec rake precompile`.

### DIFF
--- a/lib/gollum/public/gollum/stylesheets/template.scss.erb
+++ b/lib/gollum/public/gollum/stylesheets/template.scss.erb
@@ -53,19 +53,21 @@ a {
 /* Markdown body */
 
 .header-enum {
-  h2 {counter-reset: h3}
-  h3 {counter-reset: h4}
-  h4 {counter-reset: h5}
-  h5 {counter-reset: h6}
+  h1 {counter-reset: h2; counter-increment: h1}
+  h2 {counter-reset: h3; counter-increment: h2}
+  h3 {counter-reset: h4; counter-increment: h3}
+  h4 {counter-reset: h5; counter-increment: h4}
+  h5 {counter-reset: h6; counter-increment: h5}
+  h6 {counter-increment: h6;}
 
   --header-enum-style: decimal;
   
-  h1:before {counter-increment: h1; content: counter(h1, var(--header-enum-style)) ". ";}
-  h2:before {counter-increment: h2; content: counter(h1, var(--header-enum-style)) "."  counter(h2, var(--header-enum-style)) ". ";}
-  h3:before {counter-increment: h3; content: counter(h1, var(--header-enum-style)) "." counter(h2, var(--header-enum-style)) "." counter(h3, var(--header-enum-style)) ". ";}
-  h4:before {counter-increment: h4; content: counter(h1, var(--header-enum-style)) "." counter(h2, var(--header-enum-style)) "." counter(h3, var(--header-enum-style)) "." counter(h4, var(--header-enum-style)) ". ";}
-  h5:before {counter-increment: h5; content: counter(h1, var(--header-enum-style)) "." counter(h2, var(--header-enum-style)) "." counter(h3, var(--header-enum-style)) "." counter(h4, var(--header-enum-style)) "." counter(h5, var(--header-enum-style)) ". ";}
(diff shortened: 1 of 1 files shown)


dometto:
(Copying the below from the issue)

I just noticed that with header numbering enbled, the 'link' icon to the left of the header is in the wrong position:

Screenshot 2020-12-15 at 11 36 03

If anyone has any suggestions on how to fix this, that'd be very helpful!

dometto:
> I just noticed that with header numbering enbled, the 'link' icon to the left of the header is in the wrong position:

I've come up with a fix to the issue. Result looks like this:

Screenshot 2020-12-16 at 17 17 56

However, it's not very pretty as it just overrides the `margin-left` for the relevant `anchor` elements. And it doesn't end up in the exact same place as the anchor for un-numbered headings, as the position for those is fixed by `primer` at `-20px`. And I don't see how we can get the anchor to be in that position as:

1. h1-6 have prefixes of different character lengths (2 for h1, 4 for h2, etc: `'1.`',` '1.1`', etc.).
2. the character width (`ch`) of each character is different for h1-h6 (because of the different font sizes)

So we would need a way to convert `ch` to `px` for each heading level. This is probably not strictly impossible, but I wonder if we should complicate things that much. With the naked eye it's difficult to see the difference in the left placement of the anchor for numbered an un-numbered headings.

dometto:
@bartkamphorst what do you think of the (sub-ideal) fix for the anchor placement when header counting is enabled?

dometto:
Ping @bartkamphorst :)

dometto:
@benjaminwil might I ask you to have a short look at this problem, since you seem much more adept at CSS than we are? No worries if you have no time, of course!

The problem is basically that when header enumeration is enabled on a page, the header link anchor is out of position:

Screenshot 2020-12-15 at 11 36 03

I've tried to come up with a fix to this but as I explain above, it's rather [ugly](https://github.com/gollum/gollum/pull/1643/files#diff-b6173e8a97c4947696e685cd73f5f67b90335a5a28455983e3dbf0b95ca7a3b7R92-R98). If you could advice on whether you think there's a less hacky way of achieving this, that would be much appreciated!

benjaminwil:
Hey! I'm not convinced that this would be that much better, but here's my advice.

If you set the `::before` elements to `display:inline-block;` all of the `.anchor elements should at least appear in the same place on the `y` axis.

A quick mock up:

```
h1,
h2,
h3,
h4,
h5,
h6 {
  &::before {
    display: inline-block;
  }
}

.anchor {
  margin-left: -1.5em;
}
```

This got me pretty far, locally. And all of the link anchors seemed to be aligned properly for me. Hope this helps.

bartkamphorst:
@dometto Can you proceed with @benjaminwil 's suggestion?

dometto:
Unfortunately not! This is not @benjaminwil's fault but I don't seem to be able to get his snippet to work. :) @benjaminwil if it's not too much work, could I ask you to include this in one of your styling PRs? All your help is hugely appreciated!

benjaminwil:
Sure I can look more deeply into this!

benjaminwil:
I'll make a separate PR for it.

bartkamphorst:
Superseded by https://github.com/gollum/gollum/pull/1673 .

PR #1648, opened 2020-12-31
Add redirect to rename commit

ViChyavIn:
*To work, this pull request https://github.com/gollum/gollum-lib/pull/390 needs to be merged first.*

This change will save redirects additions in commits where renaming is saved.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -303,12 +303,13 @@ module Precious
           redirect to("/#{page.escaped_url_path}")
           return
         end
-        committer.commit
-        
+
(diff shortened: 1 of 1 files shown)


bartkamphorst:
> This change will save redirects additions in commits where renaming is saved.

I'm having a hard time understanding the use case here. Could you elaborate on what this PR means to achieve?

ViChyavIn:
Let's say I've renamed a page.

WIthout this change latest changes would look like this:

image

Otherwise:

image

So basically it adds redirect addition to the commit containing renaming. I think it makes more sense this way.

PR #1649, opened 2021-01-03
Fix binary diffs unable to be viewed in history. Resolves #1650

### LINKED ISSUE #1650: Bug: binary diffs unable to be viewed in history
**Describe the bug**
Viewing binary files diffs is impossible (causes error).

I've opened [a pull request](https://github.com/gollum/gollum/pull/1649) to fix the issue.

**To Reproduce**
1. Create git repo and run Gollum
2. Add a binary file
3. Attempt to view the commit in history

**Expected behavior**
Viewing the diff should not cause an error.

**Environment Info**
```
Gollum 5.1.2
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.0.6
Other renderering gems:
RedCloth 4.2.9
org-ruby 0.9.12
wikicloth 0.8.3
```

ViChyavIn:


lib/gollum/views/commit.rb
```diff
@@ -23,12 +23,24 @@ def message
       end
 
       def files
-        files = @diff.split(%r{^diff --git a/.+ b/.+$}).reject(&:empty?)
+        files = @diff.force_encoding(Encoding::UTF_8).scan(%r{
```
dometto, line 26: Nice and clean refactor! Only thing I'm wondering about is: does this still implement the equivalent of `.reject(&:empty?)`? Or do we no longer need it?

dometto: Also, are the changes in this file just a refactor, or do they implement (slightly) different behavior?

dometto: Ah, I think I see it now: we're precisely not rejecting empty diffs to support changes to binary files?

dometto (commented):
A couple of questions but I like the refactor and the idea of supporting binary files. A test case would be great!

dometto:
Thanks for identifying all these bugs and working on the fixes, @ViChyavIn! Much appreciated!

lib/gollum/views/commit.rb
ViChyavIn: > does this still implement the equivalent of .reject(&:empty?)
I don't think we need it anymore because we use a scan method, so there won't be empty strings.

ViChyavIn: > Also, are the changes in this file just a refactor, or do they implement (slightly) different behavior?

Mainly it's a refactor, but I also [added another `gg` class](
https://github.com/gollum/gollum/pull/1649/commits/c93a46034b220e99a5ff1c134f883e29436fe583) for `\ No newline...` and `Binary files...` messages and [put `...` as placeholder for line numbers.](https://github.com/gollum/gollum/pull/1649/files#diff-09095607da570d2efe65488d68a6b0d6d3cfd6c5661d489855835572abd4946fL78-R85)

ViChyavIn:
Added some tests in https://github.com/gollum/gollum/pull/1649/commits/479cf620bb26c055c7b62da778dfa53db3c11ab0!

dometto:
Thanks! Getting two Travis failures unfortunately.

On MRI/'normal ruby:
```
Failure:  /latest_changes should include the :pagination_count commit.
```

On JRuby:

```
<[{:class=>"gg",

  :ldln=>"...",

  :line=>"Binary files /dev/null and b/Mordor/eye.jpg differ",

  :rdln=>"..."}]> expected but was

<[{:class=>"gg", :ldln=>"...", :line=>"Binary files differ", :rdln=>"..."}]>
```

Are you also getting the MRI error locally? For the JRuby error, I recommend just making the test more lenient. It looks like JRuby is just not adding the information about the binary filenames to the diff, so you could just test whether the string matches something like `/Binary files (.*)? differ/`.

ViChyavIn:
It turned out I put a wrong SHA1 in the test, my bad!

> `Failure:  /latest_changes should include the :pagination_count commit.`

I got the same result. I added one more commit to examples/lotr.git but I didn't expect it would break the latest changes test. I rewrote it a bit so adding more commits to the test repo won't break anything again.

PR #1653, opened 2021-01-08
Add option to show browser's local time

NikitaIvanovV:
This PR adds an option to show browser's local time when viewing history, commit, latest_changes and clicking `When was this page last modified?` button.

I used the client-side approach: I pass time in `datetime` attribute of `<time>` tag and then JS creates a Date object, which will be in borwser's local time. To support `strftime` (used for formatting Time in latest_changes and history) on the client, I had to include [Datejs library](https://github.com/datejs/Datejs). I hope adding another library for such a minor thing is not a big problem?


ViChyavIn:
Oh no, I forgot to add tests again. Will do tomorrow!

dometto:
@ViChyavIn this looks good to me. Perhaps you could change the documentation strings a little:

`Show browser's local time instead of server's.`

...is a bit unclear. If I understand the point of this PR correctly, it would perhaps be better phrased by:

`Use the browser's local timezone instead of the server's for displaying dates."

@bartkamphorst any thoughts?

dometto:
@ViChyavIn it also looks like one of the tests is failing on JRuby:

https://travis-ci.org/github/gollum/gollum/jobs/754610273

Could you investigate? Thanks!

ViChyavIn:
@dometto Done!

lib/gollum/views/history.rb
```diff
@@ -18,18 +20,21 @@ def versions
...
+            :href        => page_route("gollum/commit/#{v.id}"),
+            :href_page   => page_route("#{filename}/#{v.id}"),
+            :num         => i,
+            :selected    => @page.version.id == v.id,
+            :author      => v.author.name.respond_to?(:force_encoding) ? v.author.name.force_encoding('UTF-8') : v.author.name,
+            :message     => v.message.respond_to?(:force_encoding) ? v.message.force_encoding('UTF-8') : v.message,
+            :date_full   => authored_date,
+            :date        => authored_date.strftime(DATE_FORMAT),
+            :datetime    => authored_date.utc.iso8601,
+            :date_format => DATE_FORMAT,
```
dometto, line 35: Is it really necessary to pass along `:date_full` and `:date` in addition to the datetime and date_format? I mean: can't the browser just compute the former, given the latter?

Also, since `:date_format` is always going to be the same, doesn't it make more sense to implement it as a method on a view helper `HasTime` that could be included in all the views that require it, along with other required methods?

dometto (commented):
Thanks @ViChyavIn. One more suggestion for refactoring in the comments. Sorry for not mentioning it earlier.

lib/gollum/views/history.rb
dometto: Ping @ViChyavIn

ViChyavIn: @dometto Oh, I'm terribly sorry! Forgot about this PR.

I pass `:date_full` because it has been there before and someone might rely on them in their custom mustache files.

Let me explain how it works right now. Gollum always sends `<time>` tags with time in server timezone. If Gollum is set to print time in server timezone, no JS is ran and the tags remain untouched. If set to print in local timezone, [JS takes UTC time from `datetime` attribute](https://github.com/gollum/gollum/pull/1653/files#diff-a50b01f9b9463206ae82d9bd65cff7f5313e728e992e3770cd6130cddabebfd0R165-R174) and replaces contents of `<time>` with computed local time. I decided to use this approach to several reasons:
1. I failed to understand how to convert UTC Date object to another Date object in server timezone
2. The fact that `<time>` tags already have contents by default becomes a sort of a fallback in case JS is either fails to load or disabled in a browser.
3. Another approach where we always compute time string would require passing server timezone to the client, which can increase reponse body size in the worst case (`<time>` still has content).

As for `HasTime`, [here](https://github.com/ViChyavIn/gollum/blob/2b290e5fa8ad2f30f40d111536c93140b88b6072/lib/gollum/views/page.rb#L15) `DATE_FORMAT` is different. Shoud I still make the helper for the rest where the constant is the same?

dometto: Thanks for the explanation. I didn't notice `DATE_FORMAT` was different for the `Page` view. I guess that defeats the purpose of a helper!

dometto:
Again, thanks for the contribution and patience, @NikitaIvanovV!

PR #1659, opened 2021-01-21
Update Primer CSS to v15.2.0

benjaminwil:
This patch updates [Primer CSS](https://primer.style) to the latest version.

Upgrading Primer enables us to make changes to existing Gollum layouts that will fare better on mobile devices.

I opted to change the Primer library file extension from `.scss` to `.css` and include it as a separate file in the `<head>`. I did this because we don't need to re-compile the compiled CSS into Gollum's `app.css` file. I think it makes sense to load the vendor CSS first and separate it from the project's CSS.

If there is a historical reason for Gollum to do this, I could not find evidence of it. I'm of course happy to discuss this change further.


benjaminwil:
I've rebased this pull request against the latest `master`.

@dometto Once this pull request can be merged, I can submit an additional PR that improves Gollum's presentation on mobile devices.

dometto:
Done! Looking forward to the next one

dometto:
> I opted to change the Primer library file extension from .scss to .css and include it as a separate file in the <head>. I did this because we don't need to re-compile the compiled CSS into Gollum's app.css file. I think it makes sense to load the vendor CSS first and separate it from the project's CSS.

On gollum `v5.2` I just noticed this error in the javascript/developer console:

```
The resource from “http://127.0.0.1:4567/gollum/assets/primer-5cc5a370910e6ca6e735bcdca2c72d9172ac308429f68ae64f99f2adb12dc79b.css” was blocked due to MIME type (“text/plain”) mismatch (X-Content-Type-Options: nosniff).
```

...and indeed, the `primer` file is not part of the gollum precompiled assets. I only just noticed this as the SCSS is actually loaded fine -- it still seems to be part of `app.css` (so defined in `app.scss`)! So I think your attempt to separate primer from `app.css` was unsuccessful :) Probably fixing this would require a separete `primer.scss` file to be added?

This is not a big issue at all since primer is still working, but it would be nice to get ` {{#sprockets_stylesheet_tag}}primer{{/sprockets_stylesheet_tag}}` working for the next release, as I agree with you it would be an improvement!

benjaminwil:
I see why I made this error. I've got a PR ready to correct it. Thank you!

PR #1660, opened 2021-01-21
Make sidebar layout mobile-friendly

benjaminwil:
Gollum's sidebar did not render very well for me on my mobile device. Using Primer CSS utilities, I was able to make a mobile-friendly grid with minimal changes.

Please see the individual commit messages for more details rationale for each change.

Here's a screenshot of the (unchanged) desktop layout resulting from these commits:

Screenshot from 2021-01-20 21-31-54

And here's a screenshot of the mobile layout with a sidebar and footer enabled:

mobile view of Gollum wiki content

### DIFF
--- a/lib/gollum/templates/wiki_content.mustache
+++ b/lib/gollum/templates/wiki_content.mustache
@@ -9,7 +9,7 @@
 	    {{{toc_content}}}
 	  </div>
 	  {{/has_toc}}
-	  <div id="wiki-body" class="gollum-{{format}}-content overflow-hidden {{#left_bar}}pl-4{{/left_bar}}">
+	  <div id="wiki-body" class="gollum-{{format}}-content overflow-hidden">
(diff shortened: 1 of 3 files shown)


bartkamphorst:
This is really great @benjaminwil ! Styling looks better and I'm learning a lot from your detailed comments. 👍 I really appreciate your efforts. The only thing I was confused about was the naming of the `bar_float` method, since I associated `bar` with `sidebar`. How would you feel about changing that to `wiki-float` or `body-float` instead?

benjaminwil:
Oh, that is a very good comment. I see why my naming was confusing now.

There was a `bar_side` method being defined already. So I have renamed my method to be `body_side`, which always returns the opposite of `bar_side`. I think this is more clear.

Thank you for your review.

dometto:
Closing and reopening this to see if that fixes the Travis tests

PR #1661, opened 2021-01-21
Remove unused _layout.scss file

benjaminwil:
The SCSS being removed in this commit is not being imported or used. Please disregard this pull request if I am mistaken.

To verify, I added a `background-color: red;` definition to the `#wiki-body` styles. On reload, the background colour did not change. The background colour only changed after I added an `@import "_layout";` call to the `template.scss.erb` file.

I also grepped through the compiled CSS files for any style definitions that contained `#wiki-` and I did not find any.


bartkamphorst:
I recall having the same thought, but changing my mind at some point. Unfortunately I can't remember why exactly; maybe the styling of the sidebar?

benjaminwil:
I just revisted this. I am now very confident that this file is not being referenced.

To test:

1. Take any of the CSS selectors referenced in `_layout.scss`. For example, `#wiki-sidebar`.
2. Do a project-wide search for the selector.
3. Note any results that are in the compiled CSS assets.

If this selector is being styled, the selector will show up inside of the compiled CSS files.

The `_layout.scss` file is not being referenced from a main SCSS file, and thus is not included when being compiled.

I performed this test with `#wiki-sidebar`, `#wiki-footer`, `and #wiki-content`. None of these selectors appeared in the compiled CSS file at all. I am confident that this means that elements with this IDs are not being styled by the SCSS in `_layout.scss` at all.

PR #1662, opened 2021-01-21
Update reference to test git bare repositories

benjaminwil:
The CONTRIBUTING document referenced a file that has since been moved.


dometto:
The test failures in this PR and @benjaminwil's other PRs are because we need to release a new version of `gollum-lib` that supports recent changes to gollum itself, I think.

@benjaminwil: many many thanks for the PRs! We'll try to fix the tests and review soon.

benjaminwil:
Let me know if there's anything I can help with regarding cutting a new release of `gollum-lib`; I'm happy to support.

PR #1663, opened 2021-01-24
Only escape HTML from breadcrumbs. Resolves #1658

### LINKED ISSUE #1658: Gollum 5.1.2 displays Chinese path name as URI
Some of my wiki pages are in a directory named `数学`,  which means `maths` in English.

* In Gollum 5.1, the path is displayed as `数学`
  gollum-5 1

* In Gollum 5.1.2, the path is displayed as `%E6%95%B0%E5%AD%A6`
   gollum-5 1 2

After search in the git logs, I find it was introduced by  [Guard against malicious file names](https://github.com/gollum/gollum/commit/137728cdabc0f60859fcd30404ad2b8fff6ef715#diff-9a4bb08f6de673dd18e80f29cd9ed26211e3ba56ace35db0210b3cc9dafcc4e7)

Obviously, we can not just use `CGI.unescape` to solve the problem.

---

Test code:

```ruby
#!/usr/bin/ruby

require 'cgi'
puts CGI.escape("数学")
```

benjaminwil:
These commits respond to a bug reported, #1658.

In investigating the bug, I found opportunities to add additional tests around retaining and displaying unicode characters. See the commit messages for more detailed information.

### DIFF
--- a/lib/gollum/views/page.rb
+++ b/lib/gollum/views/page.rb
@@ -32,7 +32,7 @@ module Precious
         path.descend do |crumb|
           element = "#{crumb.basename}"
           next if element == @page.title
-          breadcrumb << %{<li class="breadcrumb-item"><a href="#{overview_path}/#{crumb}/">#{CGI.escape(element.to_s)}</a></li>}
+          breadcrumb << %{<li class="breadcrumb-item"><a href="#{overview_path}/#{crumb}/">#{CGI.escapeHTML(element.to_s)}</a></li>}
         end
         breadcrumb << %{</ol></nav>}
         breadcrumb.join("\n")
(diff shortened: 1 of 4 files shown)


bartkamphorst:
Looks good to me. @dometto Is this ready to be merged?

bartkamphorst:
Thanks @benjaminwil ; great to have the test suite extended like this. 👍

dometto:
Many thanks @benjaminwil!

lib/gollum/views/page.rb
```diff
@@ -265,6 +260,10 @@ def table(data)
         result << "</tr>\n</table>\n"
       end
 
+      def title
```
dometto, line 263: Looks like making this method private is triggering test failures on `test_page_view.rb:143` where this is called:

```ruby
 actual = @view.title
```

I think you can change that to the following to get around this:

``ruby
actual = @view.send(:title)
```

benjaminwil: Thank you, I completely missed this.

Instead of testing the now-private `title` method I added additional tests for the public `page_header` method that is exposed on the frontend.

So now `h1_title = false` and `h1_title = true` cases are explicitly tested against `page_header` instead of `title`.

dometto:
Good solution! Thanks so much

PR #1667, opened 2021-02-04
Add JS polyfill for IE compatibility

### LINKED ISSUE #1664: MSIE support
Under MSIE 11, a javascript exception occurs on page load because Array.prototype.includes() is used.

The exception can be avoided with the following polyfill, inserted at the top of layout.mustache, or by avoiding calls to [].includes

    if (!Array.prototype.includes) {Array.prototype.includes = function (x) { return 0 <= this.indexOf(x) }}

If Gollum is not supposed to support MSIE, perhaps that could be documented somewhere.

garetethan:
Resolves #1664. Add Array.prototype.includes() polyfill to prevent
error in Internet Explorer. Verified to work in Internet Explorer 11
on Windows 10.


dometto:
Thanks for the PR and the verification @garetethan! One small request: could you add [a mustache comment](https://mustache.github.io/mustache.5.html#Comments) to the code indicating its purpose?

bartkamphorst:
Would it be cleaner and more expressive if the actual code was placed in a separate `polyfills.js` file, and referenced/loaded in the template?

garetethan:
Yes, I can add a comment. I am attempting to move the JS into a separate file in `lib/gollum/public/gollum/javascript`, but am having trouble figuring out how to reference it from `layout.mustache`. I don't see how I can reference a file that is part of Gollum, rather than one in the wiki repository that is being served. Can you point me in the right direction here?

benjaminwil:
Hey @garetethan, in `layout.mustache` you can see the `app.js` and `editor.js` files being referenced using the `sprockets_javascript_tag`. 

One way to go would be to add a new JavaScript tag for polyfills to the layout file, then make sure it is compiled like the other JS files (the Sprockets asset compilation step is described in CONTRIBUTING.md). The other way to go would be to add a new `polyfills.js` file to list of assets in `app.js` and then recompile the assets without changing `layout.mustache` at all.

Let me know if I’ve misinterpreted your question.

bartkamphorst:
> The other way to go would be to add a new polyfills.js file to list of assets in app.js and then recompile the assets without changing layout.mustache at all.

This is probably the best way to go, since we'll want the polyfills to be present in pretty much all contexts.

garetethan:
I've added polyfills.js to `/lib/gollum/public/gollum/javascript`, required it in `app.js`, recompiled the assets, and re-verified that it works for me in IE 11 on Windows 10. How does it look?

lib/gollum/views/layout.rb
```diff
@@ -50,6 +50,10 @@ def mathjax_js
         page_route("gollum/assets/mathjax/MathJax.js?config=TeX-AMS-MML_HTMLorMML")
       end
 
+      def polyfill_path
```
bartkamphorst, line 53: Why do we need this?

garetethan: Oh, sorry, my mistake. Removed.

dometto:
Thanks @garetethan and @benjaminwil for the support!

PR #1670, opened 2021-02-13
Update README.md

dometto:
I requested but was never assigned a CVE, so think we might as well remove this notice.


bartkamphorst:
Is there a way to make a complaint about this?

dometto:
Probably, but not sure if it's worth the effort. I've had similar experiences with requests for CVE :(

Issue #1672, opened 2021-02-17
can't satisfy 'octicons (~> 8.5)', already activated 'octicons-12.0.0'


DanSman:
Hey

I've just update to the new version. Not sure that's really a bug on your side but I only use ruby for gollum so I thought I'll try my best here.

After the update I get the following problem:
```
Feb 17 12:58:27 laptop bash[113552]: /usr/lib/ruby/2.7.0/rubygems/specification.rb:1396:in `block in activate_dependencies': can't satisfy 'octicons (~> 8.5)', already activated 'octicons-12.0.0' (Gem::LoadError)
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems/specification.rb:1388:in `each'
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems/specification.rb:1388:in `activate_dependencies'
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems/specification.rb:1370:in `activate'
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems.rb:302:in `block in activate_bin_path'
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems.rb:301:in `synchronize'
Feb 17 12:58:27 laptop bash[113552]:         from /usr/lib/ruby/2.7.0/rubygems.rb:301:in `activate_bin_path'
Feb 17 12:58:27 laptop bash[113552]:         from /home/dan/.gem/ruby/2.7.0/bin/gollum:23:in `<main>'
```

I tried v8.5 and 12 both aren't working. 

Is there anything else I could do?

Thanks

dometto:
Ah, hope I didn't break this in the new version!

Are you running from source with Bundler? In that case try removing
Gemfile.lock and installing again?

DanSman:
Where can I find the file?

I installed it like [your wiki](https://github.com/gollum/gollum#installation).

DanSman:
Hm.

Maybe I leave it open. `gem clean` deletes
```
Successfully uninstalled gollum-lib-5.0.6
```
which gives me the error message.

dometto:
[...]

PR #1673, opened 2021-02-18
Fix heading counters and anchor icon presentation. Resolves #1642

### LINKED ISSUE #1642: Missing css definition breaks a header numbering (proposed fix)

benjaminwil:
This pull request was requested in #1643 in order to fix #1642. (I also cherry-picked the first commit from #1643, then spent some time investigating an alternative solution to the anchor CSS presentation.)

To summarize the issues being solved:

- Heading counters were not being reset when `header_enum: true` was set in a page's frontmatter metadata.
- Link anchor icons (:link:) were not appearing in the correct place and were overlapping the heading text on `:hover`.

I also took this opportunity to re-scope `.anchor` related SCSS.

I also ensured that the `.anchor` icons continue to work if `header_enum: false`.

Here is a screenshot that demonstates where `.anchor` icons appear in relation to their heading:

Screenshot from 2021-02-17 22-34-08

The icons appear in the same place on the page's Y axis whether the heading is an `<h1>` or an `<h6>`.


bartkamphorst:
This is really, really great @benjaminwil ! I understand why you wanted to do this in a separate PR. 👍

dometto:
Thanks again @benjaminwil! This is really great, so much cleaner! There's a merge conflict because of one of the statically compiled files. You could just leave those out and I can do a final precompile before doing the new release.

dometto:
I might do a new release of gollum after merging this, in order to solve https://github.com/gollum/gollum/issues/1672. We can always do another release for forthcoming features (like the button re-labeling). But please let me know if anything forthcoming absolutely needs to be part of the next release!

benjaminwil:
> But please let me know if anything forthcoming absolutely needs to be part of the next release!

No, nothing else should absolutely need to be part of the next release. I am trying to make these changes as isolated as possible. :+1:

> You could just leave those out and I can do a final precompile before doing the new release.

I believe not merging new compilations could break Gollum's frontend for anyone attempting to run off of `master`, so I will re-commit the compiled assets.

I intend to do this later today or tomorrow.

benjaminwil:
I've recompiled the assets. I think this should be good to merge now.

dometto:
Amazing! I'll do a quick gem release of gollum to include your recent updates, and to fix #1672

PR #1676, opened 2021-02-21
Move "Page History" into button group

benjaminwil:
If the user has edit permissions, the "History" button is shown in a button group with "Edit" and "Rename". If the user does not have edit permissions, it's shown by itself.

This is a small change, but it seems more semantic to me: All of the "Edit", "Rename", "Upload" and "History" buttons should be displayed in the same button group because they are all actions that relate to the current page. All other buttons ("New", "Overview", "Latest Changes") do not.

This commit also renames the label from "Page History" to "History".

### DIFF
--- a/lib/gollum/templates/navbar.mustache
+++ b/lib/gollum/templates/navbar.mustache
@@ -18,7 +18,7 @@
     </div>
 
     <div class="TableObject-item hide-sm hide-md">
-      <div class="BtnGroup" style="display: flex;">
+      <div class="BtnGroup d-flex">
         {{#overview}}
           <a
             class="btn BtnGroup-item btn-sm"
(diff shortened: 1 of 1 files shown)


lib/gollum/templates/navbar.mustache
```diff
@@ -41,21 +41,19 @@
...
-          class="btn btn-sm"
-          href="{{history_path}}/{{escaped_url_path}}"
-          id="minibutton-history"
-        >
-          Page History
-        </a>
-      </div>
-    {{/history}}
+    <div class="TableObject-item px-2">
+      <div class="BtnGroup" style="display: flex;">
```
bartkamphorst, line 45: Is there a class for `display: flex` instead? Otherwise I think I'd prefer to have this styling in our custom CSS, rather than in the mustache template itself.

benjaminwil: Ah, it looks like there is a utility class: `d-flex`. 

I will update and rebase this PR to fix.

benjaminwil:
This pull request has been rebased against `master`.

bartkamphorst:
Looks fine to me. @dometto ?

bartkamphorst (approved):

dometto (approved):

PR #1677, opened 2021-02-21
Fix sidebar and footer padding

### LINKED ISSUE #1674: Minor styling issue: footer and sidebar touching on (nearly) empty page
Absolutely love the improvements to the styling you've made, @benjaminwil! When giving gollum a test drive before cutting a new gem release, I just noticed a minor styling issue.

Previously, this is what an almost empty page with sidebar and footer would look like:

Screenshot 2021-02-20 at 11 23 33

So the existing situation is arguably already not super neat. But after recent changes, the footer's width has been extended, so we now get the slightly weirder (or maybe it's just weird in a different way:)) looking following result:

Screenshot 2021-02-20 at 11 23 54

It's a very minor issue as it literally only happens when the page is entirely empty or only has a single line on it. But perhaps we should put some padding between the footer and sidebar? If that's somehow complicated because the way our CSS is currently structured, I'm also happy to keep things as they are for the next gem release!

benjaminwil:
Resolves #1674.

### Padding between sidebar and footer containers

@dometto reported that there was no spacing between the optional sidebar and footer containers. This would look a bit funny in edge cases, specifically if the sidebar was longer than the main wiki content or both the main wiki content and the sidebar were empty.

After:

Screenshot from 2021-02-20 18-26-57

### Padding inside of sidebar and footer containers

While investigating the above issue, I noticed that the footer and sidebar also had very different content padding definitions, which looked off on mobile devices--though I think it is technically off on all devices.

You can see the padding change in the above screenshot, but it's much clearer on mobile:

Before:

Screenshot from 2021-02-20 18-43-54

After:

Screenshot from 2021-02-20 18-30-25

### DIFF
--- a/lib/gollum/templates/wiki_content.mustache
+++ b/lib/gollum/templates/wiki_content.mustache
@@ -22,8 +22,8 @@
 	        {{{content}}}
 	      </div>
 	      {{#has_sidebar}}
-          <div id="wiki-sidebar" class="Box Box--condensed float-md-{{body_side}} col-md-3 px-4">
-	        <div id="sidebar-content" class="gollum-{{sidebar_format}}-content markdown-body">
+          <div id="wiki-sidebar" class="Box Box--condensed float-md-{{body_side}} col-md-3">
+	        <div id="sidebar-content" class="gollum-{{sidebar_format}}-content markdown-body px-4">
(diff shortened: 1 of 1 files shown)


bartkamphorst:
Looks good!

Issue #1381, opened 2019-05-10
5.x: Replace gollum's custom dialogs with Box--overlay (still experimental in primer)


bartkamphorst:
Gollum currently uses custom dialog modals (upload, rename, new). When the switch to Primer CSS has been merged, we should look into deprecated these custom modals for Primer's [Box--overlay](https://styleguide.github.com/primer/components/box-overlay/). At the moment this package is still experimental.

bartkamphorst:
@benjaminwil Any thoughts about using this still experimental Primer feature [Box--overlay](https://github.com/gollum/gollum/pull/1671) to replace our custom modals?

benjaminwil:
I haven't looked closely at the existing modals yet, but I do think that this would be a reasonable change.

While `Box--overlay` is still experimental and the implementation may require some tweaks in the future, I would think the box styles themselves will not be removed from Primer.

I think this change would also allow us to remove some amount of custom CSS, which is always preferrable for a project like Gollum, in my opinion.

I can look more deeply into this sometime soon.

bartkamphorst:
[Box overlays](https://primer.style/css/components/box-overlay) are still in `Alpha` but since they have been around a while, it may be interesting to use them anyway for `v.6.0`.

bartkamphorst:
I just noticed browser support for [pop-overs](https://web.dev/blog/popover-baseline). Maybe we can incorporate those instead of relying on Box overlays?

PR #1679, opened 2021-02-22
Add I18n interface for use in Mustache templates

benjaminwil:
This pull request adds the `i18n` gem, which is a very standard Ruby internationalization library.

The Gollum frontend uses Mustache templates. This means we could not use `I18n.t` directly, as these templates do not take arbitrary Ruby calls like ERB or HAML does. Because of this, we needed to add an Mustache helper that can give required i18n data to the templates any time that HTML is being compiled from the Mustache templates. 

I ended up making an interface, `t`, which captures a subset of the I18n strings, where the scope is the current class. Here's a concrete example of how the `t` interface can be used:

```yaml
# lib/locales/en.yml
en:
  fruit:
    apple: Apple
    orange: Orange
    mango: Mango
    dealers_choice: %{my_choice}
```
```ruby
# lib/views/fruit.rb
class Fruit < Mustache
  include LocaleHelper

  def my_choice
    "Strawberry"
  end
end
```
```mustache
{{! lib/templates/fruit.mustache }}
<ul class="fruits">
  <li>{{ t.apple }}</li>
  <li>{{ t.orange }}</li>
  <li>{{ t.mango }}</li>
  <li>{{ t.dealers_choice }}</li>
</ul>
```

If the `fruit.mustache` template uses the `Fruit` view class, then the following HTML would be rendered for that template:

```html
<ul class="fruits">
  <li>Apple</li>
  <li>Orange</li>
  <li>Mango</li>
  <li>Strawberry</li>
</ul>
```


test/gollum/views/test_helpers.rb
```diff
@@ -0,0 +1,137 @@
+require_relative "../../helper"
+require_relative "../../../lib/gollum/views/helpers"
+
+require "minitest/autorun"
+require "minitest/spec"
```
benjaminwil, line 5: `minitext/autorun` and `minitest/spec` are not used in the rest of Gollum's test suite. So as you can see, the tests here are very different than the existing test suite.

I am open to changing these tests to better match the existing suite. However, I do feel that spec-style tests are easier to read and write. I'm curious to hear your thoughts on them, and whether it would be worth refactoring some other tests to use the `spec` DSL.

test/helper.rb
```diff
@@ -5,6 +5,7 @@
 require 'mocha/setup'
 require 'fileutils'
 require 'minitest/reporters'
+require 'minitest/spec'
```
benjaminwil, line 8: `minitest/spec` is not used in the rest of Gollum's test suite. So as you can see, the tests included in this PR are very different than the existing test suite.

I am open to changing these tests to better match the existing suite. However, I do feel that spec-style tests are easier to read and write. I'm curious to hear your thoughts on them, and whether it would be worth refactoring some other tests to use the spec DSL.

dometto: I'm not against using spec format for any newly added tests, as long as it plays well with code coverage! There is no reason for using the older `assert` based tests other than that's how we inherited them. They're more difficult to read, but otherwise function well. I think it would be a lot of effort to migrate them all to spec format, so not sure that's worth it!

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+          if i18n_value.is_a? Hash
+            [i18n_key, autofill(i18n_value)]
+          else
+            yaml_variable_regex = /\%\{[\w]+\}/
+            yaml_arguments = i18n_value.scan(yaml_variable_regex).uniq
+
+            next [i18n_key, i18n_value] if yaml_arguments.empty?
+
+            yaml_arguments.each do |argument|
+              non_word_character_regex = /[^\w]/
```
dometto, line 42: It would be slightly more efficient not to assign this variable, and just call `.gsub(/[^\w]/`,'')` . I think the regex is readable enough without the variable assignment :)

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+              method_name = argument.gsub(non_word_character_regex, "")
```
dometto, line 43: `''` is more efficient than `""` (at least in older versions of ruby -- maybe this has changed recently?)

dometto (approved):
I really like this approach!

If I understand correctly, this approach would work only for the mustache templates? Any English in the JavaScript would presumably require a different treatment (or better: we should not be including hardcoded English in JavaScript).

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: I will have to do more thorough research, but my understanding is that the performance loss is negligible in recent Rubies. 

But really you are right, I have just done this more as a habit, as some projects I work on require double-quotes for all strings.

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: I agree. I will revisit this and resolve when finished. Thank you.

benjaminwil:
I did not realize we hardcoded English in JavaScript. I feel like internationalizing that could be even less effort than this was, though. I will look into it.

lib/gollum.rb
```diff
@@ -14,6 +15,9 @@
 module Gollum
   VERSION = '5.1.2'
 
+  I18n.available_locales = [:en]
+  I18n.load_path = Dir[File.expand_path("lib/gollum/locales") + "/*.yml"]
```
bartkamphorst, line 19: We prefer to namespace common built-in classes with `::`, so `::Dir` and `::File` respectively, just so they will not inadvertently clash with classes from external dependencies with the same name.

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+          end
+        }.to_h
+      end
+
+      # Returns the current class name in a format that is acceptable in YAML.
+      # To summarize its function:
+      #
+      #     NameOfConstant => name_of_constant
+      #
+      def locale_klass_name
```
dometto, line 62: If this gets called multiple times per template it may be good to use this pattern:

```ruby
@locale_klass_name ||= ...
```

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+            yaml_arguments = i18n_value.scan(yaml_variable_regex).uniq
+
+            next [i18n_key, i18n_value] if yaml_arguments.empty?
+
+            yaml_arguments.each do |argument|
+              non_word_character_regex = /[^\w]/
+              method_name = argument.gsub(non_word_character_regex, "")
+
+              begin
+                i18n_value.gsub!(argument, self.public_send(method_name))
```
bartkamphorst, line 46: I think we should check here for the return value of `method_name`. If it happens to be nil, then `gsub!` will raise a `TypeError`(no implicit conversion of nil into String).

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+      # When a translation value with an argument has no matching method, we
+      # then return that value transformed to include the `no_method_message`
+      #
+      def autofill(yaml)
+          yaml.map { |i18n_key, i18n_value|
+          if i18n_value.is_a? Hash
+            [i18n_key, autofill(i18n_value)]
+          else
+            yaml_variable_regex = /\%\{[\w]+\}/
+            yaml_arguments = i18n_value.scan(yaml_variable_regex).uniq
```
dometto, line 37: Wondering if alternatively, you could just call this here:

```ruby
i18n_value.gsub!(/\%\{[\w]+\}/) do |argument|
  # Handle block
end
```

...and whether this would be faster. I guess that depends on how much effort is saved by the `.uniq` call?

lib/gollum/views/helpers/locale_helpers.rb
bartkamphorst, line 62: Am I correct to assume this regexp in `locale_klass_name` is run every time any built-in String is rendered? If so, can we preprocess or cache these names instead?

bartkamphorst (changes requested):

dometto:
> I did not realize we hardcoded English in JavaScript. I feel like internationalizing that could be even less effort than this was, though. I will look into it.

I'm not *sure* that we do, but it wouldn't surprise me if that's happening in the Gollum Dialog JS, for example! So https://github.com/gollum/gollum/issues/1381 might solve that anyway :)

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: Nice catch. Thank you.

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: Hm, at first I thought the `uniq` call was necessary. But now I realize it is unlikely that a single translation string would contain a variable/method name in it more than once.

I am updating the method to match your comment.

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: Yes! I am now memoizing this method.

lib/gollum.rb
benjaminwil: `I18n` is actually the external dependency. But I've added the `::` prefix regardless. 👍

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: That makes sense. I have added a guard claus, `next if method_name.nil?` before the `begin` block.

benjaminwil:
I've rebased against `master` and addressed all the feedback so far.

The most significant changes are in the `#autofill` method.

benjaminwil:
I just revisited this and found that the `#autofill` method was too monolithic.

I think my revision makes it easier to understand.

I've re-requested your review @dometto.

lib/gollum/views/helpers/locale_helpers.rb
```diff
@@ -0,0 +1,81 @@
...
+          elsif has_arguments?(i18n_value)
+            fill_argument_content(i18n_key, i18n_value)
+          else
+            [i18n_key, i18n_value]
+          end
+        }.to_h
+      end
+
+      def fill_argument_content(i18n_key, i18n_value)
+        i18n_value.gsub!(/\%\{[\w]+\}/) do |argument|
```
dometto, line 44: Since you're using the same regex again in `#has_arguments?`, you could store it in an appropriately named constant. But the regex is very readable, I'm fine with it this way!

dometto (approved):
Again, apologies for my unresponsiveness. Agreed, the refactor is much nicer! Really like this feature. A merge error has cropped up meanwhile, unfortunately, but if you take care of that please feel free to merge!

lib/gollum/views/helpers/locale_helpers.rb
benjaminwil: I decided to add a constant for this regexp. Thanks for the feedback!

benjaminwil:
@bartkamphorst Looks like I'm unable to merge due to your previous "Changed requested" review. Could you re-review?

bartkamphorst (approved):
Great feature, great PR!

PR #1681, opened 2021-02-22
Make Keybinding and Markup selects smaller

dometto:
**EDIT**: oops, accidentally included the same screenshot twice. Now fixed!

Looks like the update to Primer caused the Markup and Keybinding select elements in the editor view to become larger than they were:

Screenshot 2021-02-22 at 17 21 03

This PR makes them smaller again so they fit on the same row:

Screenshot 2021-02-22 at 17 19 42

But as you can see there is not much space between these fields and the buttons to the left. Is this acceptable styling, or should we e.g. move the selects down a row?

### DIFF
--- a/lib/gollum/templates/editor.mustache
+++ b/lib/gollum/templates/editor.mustache
@@ -50,13 +50,13 @@
 
     <div id="gollum-editor-format-selector">
       <label for="format">Keybinding</label>
-      <select id="keybinding" name="keybinding" class="form-select">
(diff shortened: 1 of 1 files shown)


benjaminwil:
I think this is acceptable. Thanks for catching this.

Maybe in the future these could move to a new line. Or maybe even below the editor even since they are likely used less frequently.

dometto:
Thanks @benjaminwil. Both sound like good options! Open to future revisions, but I'll just merge this for now then :+1:

PR #1682, opened 2021-02-23
Fix jruby compare rename

dometto:
Was getting an error on JRuby when comparing a rename commit in the Latest Changes or Page History views. It turns out JGit handles rename commits slightly differently from Rugged.

Here's `diff` and `lines_to_compare` on MRI:

```ruby
"diff --git a/Bar.md b/Bar.md\nnew file mode 100644\nindex 0000000..e69de29\n--- /dev/null\n+++ b/Bar.md\n"
["--- /dev/null", "+++ b/Bar.md"]
"diff --git a/Foo.md b/Foo.md\ndeleted file mode 100644\nindex e69de29..0000000\n--- a/Foo.md\n+++ /dev/null"
["--- a/Foo.md", "+++ /dev/null"]
```

And here on JRuby:

```ruby
"diff --git a/Foo.md b/Bar.md\nsimilarity index 100%\nrename from Foo.md\nrename to Bar.md"
["rename to Bar.md"]
```

This PR makes sure diffs consisting of JGit-style rename lines are appropriately skipped by the `compare.rb` view. @bartkamphorst is this an acceptable way of handling the edge case?

### DIFF
--- a/lib/gollum/views/compare.rb
+++ b/lib/gollum/views/compare.rb
@@ -20,7 +20,7 @@ module Precious
       def lines(diff = @diff)
         lines = []
         lines_to_parse = diff.split("\n")[3..-1]
-        lines_to_parse = lines_to_parse[2..-1] if lines_to_parse[0].start_with?('---')
+        lines_to_parse = lines_to_parse[2..-1] if lines_to_parse[0] =~ /^(---|rename to )/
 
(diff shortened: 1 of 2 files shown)


bartkamphorst (approved):
Nothing to add. 👍

PR #1683, opened 2021-02-23
Fix styling for Latest Changes flexbox

dometto:
This brings the flexbox on the latest changes view in line with that on the page history view, which also has `col-5` rather than `col-7`.

Before:

Screenshot 2021-02-23 at 14 28 53

After:

Screenshot 2021-02-23 at 14 27 50


benjaminwil:
My apologies again for missing style changes that came with the Primer CSS update.

It didn't occur to me that there would be so many little consequences, and I'm new enough to using Gollum that I hadn't come across these view shifts yet.

dometto:
No worries @benjaminwil! I'm very happy to have the Primer update, such small bugs are bound to slip in whenever we update a major dependency. :)

PR #1685, opened 2021-02-25
Don't include primer.css separate from app.css

benjaminwil:
In #1659 I made a change to how we include Primer CSS in Gollum. When doing this I made an error.

It seems that Sprockets does not want to include a `.css` file using our `sprockets_stylesheet_tag` helper. This is okay, since I made an error where `primer` is still included in the `app.scss`.

This change is maybe preferable, since it's one less explicit network request.

Thanks to @dometto for discovering this error.


dometto:
Thanks!

PR #1697, opened 2021-03-07
Fix nav outline styles

### LINKED ISSUE #1692: Navbar button highlighting issues
**Describe the bug**

* The navbar buttons and search bar light up on focus. However, for the search bar and the buttons on the far left and far right, this highlighting is clipped. Have a look at the screenshots to see what I mean. 

* The search bar's height is different from the buttons. This is easily fixed by adding the class `input-sm` to the search bar. (Note that the clipping of the Home button highlight becomes worse without any padding from the (currently) larger input element.)

* The Rename button is the only button that does not show the highlighting on focus. It's unclear to me why.

**To Reproduce**

1. Go to the landing page
2. Click on the search bar
3. Observe the  clipped highlighting

**Expected behavior**

I would expect the search bar to have the same height as the buttons. I would expect highlighting to be visible all around the buttons and search bar.

**Screenshots**

Screenshot 2021-03-02 at 21 26 03
Screenshot 2021-03-02 at 21 26 14

Screenshot 2021-03-02 at 21 31 24
Screenshot 2021-03-02 at 21 44 02

**Environment Info**

```
Gollum 5.2.1
Running on: java with Ruby version 2.6.5
Using:
rjgit 5.7.0.2
gollum-rjgit_adapter 0.6
gollum-lib 5.1
Other renderering gems:
none
```

benjaminwil:
Fixes #1692.

There was a display issue, where navbar items's outline styles were being cut off due to the parent `<nav>` element's margin and padding.

Fortunately, we can do away with the navbar wrapper div entirely. It was not doing anything important except defining its children as `TableObject` items. But we can just do this on the `<nav>` itself.

### DIFF
--- a/lib/gollum/templates/navbar.mustache
+++ b/lib/gollum/templates/navbar.mustache
@@ -54,21 +54,21 @@
 
       {{#allow_editing}}
         {{#allow_uploads}}
-          <a
+          <button
             class="btn BtnGroup-item btn-sm hide-sm hide-md
               minibutton-upload-page"
           >
             Upload
-          </a>
+          </button>
         {{/allow_uploads}}
 
         {{#editable}}
-          <a
+          <button
             class="btn BtnGroup-item btn-sm hide-sm hide-md
               minibutton-rename-page"
           >
             Rename
-          </a>
+          </button>
           <a
             class="btn BtnGroup-item btn-sm hide-sm hide-md"
             href="{{edit_path}}/{{escaped_url_path}}"
@@ -80,7 +80,6 @@
       {{/allow_editing}}
     </div>
   </div>
-
 
   {{#allow_editing}}
     {{#editable}}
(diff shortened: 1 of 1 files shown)


bartkamphorst:
Thanks @benjaminwil , I can confirm that this resolves the clipping issue. 👍 What remains, however, is that the 'Rename' button does not highlight (on click and hold). Its neighboring button does. Any ideas?

benjaminwil:
So both "Upload" and "Rename" are `<a>` elements without a `href`, which browsers identify as something not worth highlighting in the tab index.

Changing these `<a>` elements to `<button>` elements solves the issue. A commit is incoming.

benjaminwil:
I've rebased this PR against `master`, and I've resolved the issue you highlighted above, @bartkamphorst.

bartkamphorst:
Excellent, thanks @benjaminwil !

PR #1700, opened 2021-03-20
Delete dublicate DublicatePageError error

ViChyavIn:
DublicatePageError is definded in both gollum and gollum-lib. They don't have the same arguments, which causes an error when we raise the exception when gollum is required. The fix is just removing the exception from gollum while leaving it in gollum-lib.

### DIFF
--- a/lib/gollum.rb
+++ b/lib/gollum.rb
@@ -17,22 +17,6 @@ module Gollum
   def self.assets_path
     ::File.expand_path('gollum/public', ::File.dirname(__FILE__))
   end
-
-  class Error < StandardError;
-  end
-
-  class DuplicatePageError < Error
-    attr_accessor :dir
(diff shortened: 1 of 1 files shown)


dometto:
Thanks for the PR. Seems sensible! We define the error class in gollum-lib [here](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib.rb#L45).

ViChyavIn:
@dometto Yes, I'm aware that the exception is defined there. Is there anything I need to change to get the branch merged?

dometto:
No sorry, that was just for future reference! Was just waiting to give others the opportunity to respond to this PR if they want, but basically I think this is ready to merge!

benjaminwil (approved):
I think that removing these error classes is safe.

PR #1701, opened 2021-03-20
Fix focused button border shown wrong in dialogs

ViChyavIn:
A fix for this bug:

image

I didn't understand why the overflow was set to hidden in the first place but the commit was made in 2014 so I guess the edit have become out of date by today.


benjaminwil:
Hi @ViChyavIn, thank you for this contribution.

`overflow: visible;` is the default value for the `overflow` property, so we could technically just get rid of the entire `#gollum-dialog-dialog-buttons` definition block. I checked this locally and had success.

Would you like to make this additional change and then make a second commit to [update the static assets](https://github.com/gollum/gollum/blob/master/CONTRIBUTING.md#updating-static-assets) so that users running against `master` have up-to-date assets?

benjaminwil (approved):
Regardless, this change is acceptable! Thanks for the contribution.

ViChyavIn:
@benjaminwil Done!

benjaminwil:
Thank you!

Issue #1704, opened 2021-03-27
MathJax.js not found


jqdg:
I am running gollum with the `--mathjax` flag, but I can't get any math to render. In the web inspector in the browser, I see that it can't find the `MathJax.js` file. In the page source I see
```html
<script defer src="/gollum/assets/mathjax/MathJax.js%3Fconfig=TeX-AMS-MML_HTMLorMML"></script>
```
but `MathJax.js` is not in the `assets` directory (other JavaScript files are). 

Is there some other setup step that I am not performing in order to get MathJax to work? I tried searching the docs but couldn't find anything relevant.

Output of `gollum --versions` (I tried with both Ruby 3.0 and 2.7, same behavior):
```
Gollum 5.2.1
Running on: x86_64-darwin18 with Ruby version 3.0.0
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.1
Markdown rendering gem: kramdown
Other renderering gems:
RedCloth 4.3.2
org-ruby 0.9.12
asciidoctor 2.0.12
wikicloth 0.8.3
```

dometto:
Confirmed bug, fix in #1706. Will merge and do a bugfix release ASAP, after which `gem update gollum` should fix the problem.

dometto:
Resolved in `v5.2.2`

PR #1706, opened 2021-03-27
Fix query ? in mathjax script path

dometto:
Resolve #1704

PR #1707, opened 2021-03-27
Add integration tests with capybara

dometto:
Thought it might be desirable to have integretion/acceptance tests to prevent bugs like #1704. This uses capybara to make sure no JS errors (or actually: no unexpected errors) are thrown. In the future, we could also add more demanding tests for gollum's frontend (e.g. to make sure an update to ACE or other JS assets don't break the editor).

@benjaminwil @bartkamphorst do you think tests like these are desirable? If so, do we want to get them to work on travis, or just run them locally?

### DIFF
--- a/test/integration/test_js_errors.rb
+++ b/test/integration/test_js_errors.rb
@@ -1,0 +1,43 @@
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'helper'))
+require 'selenium-webdriver'
+require 'capybara'
+require 'capybara/dsl'
+
+def console_logs(page, level = :severe, &block)
+  yield page.driver.browser.manage.logs.get(:browser).select{|log| log.level == level.to_s.upcase }
+end
+
+def expected_errors
+  [
+    %r{Refused to apply style from 'http:.*/gollum/create/custom.css'}
+  ]
+end
+
+context 'Frontend with matjax' do
+  include Capybara::DSL
+  
+  setup do
+    @path = cloned_testpath("examples/lotr.git")
+    @wiki = Gollum::Wiki.new(@path)
+    Precious::App.set(:gollum_path, @path)
+    Precious::App.set(:wiki_options, {mathjax: true})
+    Capybara.app = Precious::App
+    Capybara.default_driver = :selenium_chrome
+    Capybara.server = :webrick
+  end
+  
+  test 'expected asset errors' do
+    visit '/'
+    console_logs(page) do |logs|
+      expected_errors.each do |error|
+        assert logs.find {|log| log.message.match?(error) }
+      end
+      assert logs.size == expected_errors.size # No other errors
+    end
+  end
+  
+  teardown do
+    Capybara.reset_sessions!
+    Capybara.use_default_driver
+  end
+end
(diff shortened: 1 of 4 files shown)


dometto:
Note: this should currently fail as long #1706 is not merged yet.

benjaminwil:
I think using Capybara and having it configured in CI would be useful. 

I am a little bit confused by the committed test as written. Why are we testing for the presence of an error unrelated to the MathJax configuration?

Is it possible to set up the test so we can expect 0 errors? Or could we test for the absence of a known MathJax-related error?

dometto:
Yes I can see why this is confusing. The issue is there's always a "severe" console error when `custom.css` is not present: the app tries to load it, the file is not present, and we then get redirected to the `/create` route, triggering a MIME mismatch for `custom.css`. Would the test be clearer if we substract that error from the console log, and then check that it's empty?

benjaminwil:
Thank you for explaining that. I wonder if there's an issue that tracks that `custom.css` error issue? Seems like a fun one.

>  Would the test be clearer if we substract that error from the console log, and then check that it's empty?

Hm, I am not sure that that would be much easier to read, and it means that if something changes with the logged console errors occurs in the future then this test will break and need to be fixed.

I think the best test in this would be one where we the console errors *not* to include `mathjax.js` or some sub-string from the error message that matches the known issue.

test/integration/test_js_errors.rb
```diff
@@ -0,0 +1,46 @@
...
+    Precious::App.set(:gollum_path, @path)
+    Precious::App.set(:wiki_options, {mathjax: true})
+    Capybara.app = Precious::App
+    Capybara.default_driver = :selenium_chrome
+    Capybara.server = :webrick
+  end
+  
+  test 'expected asset errors' do
+    test_routes = ['/', '/create/Foobar.rst']
+    test_routes.each do |route|
```
benjaminwil, line 31: If this test breaks on one of the two routes, it will take more time to debug why this test is broken.

Perhaps we should add a distinct test block for each route that we care about?

test/integration/test_js_errors.rb
```diff
@@ -0,0 +1,46 @@
...
+      visit route
+      console_logs(page) do |logs|
+        expected_errors.each do |error|
+          assert logs.find {|log| log.message.match?(error) }
+        end
+        assert logs.size == expected_errors.size # No other errors
```
benjaminwil, line 37: Oh okay, this is actually a pretty okay assertion overall!

benjaminwil:
While I still like the idea in my comment, I re-reviewed your tests, and I think the assertion you have now is good.

dometto:
Thanks for the feedback, I tried to make the test a little more expressive still.

dometto:
I've also requested an open source account for https://saucelabs.com/ so hopefully we can get it [running on travis](https://docs.travis-ci.com/user/gui-and-headless-browsers/)!

dometto:
I got this to work with a simple [CI task](https://github.com/gollum/gollum/runs/6951199686?check_suite_focus=true). To run the capybara tests, use the new rake task:

* `bundle exec rake test:capybara`

The default test task (`bundle exec rake test`) will skip the capybara tests.

PR #1717, opened 2021-05-03
Use ranked search

jklippel:
In a project using gollum for documentation it came up that the search was not as helpful as it could be.
One has to search for "relevant" results in the result list.

The request came up to improve the search by implementing a rank.

The "main" change of this is to gollum-lib. A separate pull request was created in the gollum-lib Repository.

To actually use the rank of gollum-lib, adaptions must be made to the search.mustache template to actually display the rank and to the search.rb view to sort the result based on the rank.

### DIFF
--- a/lib/gollum/templates/search.mustache
+++ b/lib/gollum/templates/search.mustache
@@ -13,7 +13,7 @@
   <div class="Box-header border-bottom p-0"></div>
   {{#results}}
     <li class="Box-row Box-row--gray">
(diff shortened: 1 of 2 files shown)


jklippel:
gollum-lib Pull-Request: https://github.com/gollum/gollum-lib/pull/400

bartkamphorst:
Closing this as per https://github.com/gollum/gollum-lib/pull/400.

Issue #1450, opened 2020-02-01
Feature request: simplify layout customization


jstolarek:
I would like to customize my wiki layout in two ways:

1. Move sidebar from right to left
2. Widen the body of the page.

With Gollum 4.x I have found a relatively easy way to do it with `custom.css`. I moved the sidebar to the left with the following css:

```css
.has-rightbar #wiki-sidebar {
  float: left;
}

#wiki-body {
  float: right;
  clear: right;
}
```

and widened the page body with:

```css
.has-sidebar #wiki-body {
  width: 77%;
}

#wiki-wrapper {
    max-width: 90%;
}

#wiki-sidebar {
  width: 17%;
}
```

I just tested my wiki with the 5.x development branch and the above css is a recipe for a major disaster. After some experimenting it seems there is no simple way to perform the above customizations in an easy way. I am not complaining here about the lack of backward compatibility - this is understandable given a significant evolution from the 4.x branch. However, it is sad that there is no easy way to customize wiki appearance. Hence my feature request: **provide a way to easily customize basic layout settings like sidebar location, colours, text width, bullet-list styles, etc.** Perhaps allow custom mustache and scss files?

bartkamphorst:
> provide a way to easily customize basic layout settings like sidebar location, colours, text width, bullet-list styles, etc. Perhaps allow custom mustache and scss files?

Both of these options are already supported. You can create your own mustache templates and point to them using `--template-dir [PATH]`. Custom CSS is supported precisely through `custom.css` and `--css`. 

I think your CSS example is simply pointing to HTML elements that have been renamed or restructured; I'll have a look to see if I can come up with a snippet that does what you want it to do.

jstolarek:
> Both of these options are already supported.

Indeed, it is possible to point to a directory with custom mustache files, but I believe you cannot override just a single mustache file. You need to override all of them.

bartkamphorst:
In 5.x, to change the sidebar from right to left, use the following in your `config.rb`.
```ruby
wiki_options = {sidebar: :left}
Precious::App.set(:wiki_options, wiki_options)
```
This functionality existed in 4.x but was broken in 5.x (see https://github.com/gollum/gollum/pull/1477) so you will have to `git pull` the latest changes on the 5.x branch.

As to your other request, most if not all the things you mentioned can be changed in `custom.css`. For example, here is a snippet to change the widths of the page and the sidebar, and the change the bullet list style to squares:

```css
.container-lg {
  max-width: 90%;
[...]

PR #1719, opened 2021-05-14
Allow for overriding only specific Mustache templates/partials.

beporter:
Relates to #1450.

* This PR adds the ability to override Mustache template and partial files without having to copy the entire `lib/gollum/templates/` directory. 
* This PR introduces a change in behavior for the `--template-dir` configuration directive in that the default `Precious::App.settings.mustache[:templates]` path now **never** changes, and is instead **supplemented** with an additional directory that contains override files.

## Screenshot

gollum-template-dir-cascade

## Testing locally

* Clone and check out this branch.
* Run `bin/gollum $YOUR_LOCAL_WIKI_PATH --template-dir=./test/examples/template_cascade`
* Open http://localhost:4567/Home and note the big red sections on the page.

## Issues

* [x] Currently the single provided test passes when run in isolation (`ruby -I"lib:lib:test:." test/test_template_cascade.rb`)
  * ~~However, the test fails when run as a part of the entire suite (`bundle exec rake test`). I suspect this is due to the `Page` View maintaining state between individual test file execution but haven't figured out how to reset it explicitly.~~
  * ~~A debugging line is currently included (that should be removed before merge) to expose the template selection process.~~
* [x] There likely needs to be more tests added once the above testing issue can be solved.


beporter:
I made documentation changes. Not feeling 100% sure about the wording I chose. The "you" probably doesn't belong in there. 🤔

benjaminwil (commented):
Thanks!

> ...the test fails when run as a part of the entire suite (`bundle exec rake test`). I suspect this is due to the `Page` View maintaining state between individual test file execution but haven't figured out how to reset it explicitly.

I spent some time trying to debug this as well. I haven't figured it out yet.

I semi-verified what you're saying about the page view state being different in the test file versus the test suite. I put a `binding.irb` breakpoint inside of the `TemplateCascade#first_path_available` method:

```
      def first_path_available(name)
        priority = "#{template_priority_path}/#{name}.#{template_extension}"
        default = "#{template_path}/#{name}.#{template_extension}"
        
        binding.irb if priority.include? "page.mustache"
        
        File.exists?(priority) ? priority : default
      end
```

And this breakpoint is hit only when I run the file individually.

But I don't yet understand how this state can be retained between teardowns. 😢

benjaminwil:
I just spent a bit more time trying to track down the issue with the test suite. What I've learned:

- It doesn't matter what order I attempt to run the test suite in – the test seems to fail.
- That said, I am not entirely sure whether the tests are actually running in random order.

   It's possible there's a deeper issue with our test setup, as the `context`/`setup`/`teardown` methods are bespoke things in `test/helper.rb` written many years ago.
   
   It's possible they no longer place nice with Minitest 5, and no one has noticed yet.
- Using standard Minitest classes and methods seems to side-step the issue, in the meantime.

I was able to make the suite pass by replacing your test file with the following:

```ruby
# ~*~ encoding: utf-8 ~*~
require File.expand_path(File.join(File.dirname(__FILE__), 'helper'))

class TestTemplateCascade < Minitest::Unit::TestCase
  include Rack::Test::Methods

  def setup
    @path = cloned_testpath('examples/lotr.git')
    Precious::App.set(:gollum_path, @path)
    Precious::App.set(:wiki_options, nil)
    Precious::App.set(:wiki_options, {template_dir: testpath('examples/template_cascade')})
    @wiki = Gollum::Wiki.new(@path)
  end

  def teardown
    FileUtils.rm_rf(@path)
  end

  def app
    Precious::App
  end

  def test_overridden_page_template_is_used
    get '/Home'

    assert last_response.body.include?('PAGE_OVERRIDE')
  end

  def test_overridden_navbar_template_is_used
    get '/Home'

    assert last_response.body.include?('NAVBAR_OVERRIDE')
  end
end
```

README.md
```diff
@@ -125,7 +125,7 @@ Gollum comes with the following command line options:
 | --no-display-metadata | none  | Do not render metadata tables in pages. |
 | --user-icons      | [MODE]    | Tell Gollum to use specific user icons for history view. Can be set to `gravatar`, `identicon` or `none`. Default: `none`. |
 | --mathjax-config  | [FILE]    | Specify path to a custom MathJax configuration. If not specified, uses the `mathjax.config.js` file from repository root. |
-| --template-dir    | [PATH]    | Specify custom mustache template directory. |
+| --template-dir    | [PATH]    | Specify custom mustache template directory. Only templates you wish to override need to exist in this path. |
```
benjaminwil, line 128: I don't think including "you" is too bad.

If you are still unhappy with it, I might suggest:

```
Only overrides templates that exist in this directory.
```

beporter:
Thanks for the further attempts. I have not abandoned this! (Thought I'd have time to address the feedback this past weekend and that didn't quite turn out.) I'm hesitant to set a self deadline with no confidence in being able to meet it, but my intent is to steal some time this week at work since this overlaps a bit. Thanks for your patience and assistance!

benjaminwil:
No pressure! 👍

lib/gollum/views/template_cascade.rb
```diff
@@ -0,0 +1,36 @@
...
+        priority = "#{template_priority_path}/#{name}.#{template_extension}"
+        default = "#{template_path}/#{name}.#{template_extension}"
+puts exist: File.exists?(priority), priority: priority, default: default
+        File.exists?(priority) ? priority : default
+      end
+
+      # Following methods should track lib/mustache/settings.rb from Mustache project.
+      def template_file
+        @template_file || first_path_available(template_name)
+      end
```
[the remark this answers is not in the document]
beporter: Good comment! To be honest I was following the pattern used in the Mustache lib itself [here](https://github.com/mustache/mustache/blob/07018763/lib/mustache/settings.rb#L134). The method we're overriding looks like this natively:

```ruby
def template_file
  @template_file || "#{path}/#{template_name}.#{template_extension}"
end
```

It probably isn't likely in production, but especially in development when the Rack app is long-running and watching/reloading file changes on the fly, I can see instances where the `first_path_available(...)` result could _change_ during runtime and we'd _want_ to recompute which base file to use. Am I misunderstanding the optimization here though?

test/examples/template_cascade/navbar.mustache
```diff
@@ -0,0 +1,105 @@
+<nav class="TableObject actions pt-4 px-2 px-lg-0 overflow-x-scroll">
+  <div style="background-color: red;">NAVBAR_OVERRIDE</div>
+
```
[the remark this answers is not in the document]
beporter: Agreed! I was being lazy and copy/pasting the existing ones as a start. I'll pare them down, but still want to make sure to include "default" partials to ensure the cascade works.

beporter:
Made most of the changes discussed. Ready for additional feedback and thoughts.

beporter:
Also added a test to explicitly ensure the overrides do **not** stick around when `--template-dir` is unset.

beporter:
Only Travis failure was jruby, but it looks transient-- couldn't pull from rubygems.org at the time of the build due to a bad SSL cert:

https://travis-ci.org/github/gollum/gollum/jobs/772756565#L263

I don't have access to trigger the build to run again but I'd expect it to pass.

benjaminwil:
I've seen the same error occur over and over on another PR. It makes me think it's not transient. Hopefully we can diagnose it shortly.

Sorry for not responding sooner; I've been moving to a new home!

lib/gollum/views/template_cascade.rb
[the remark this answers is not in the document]
benjaminwil: No I think you're right. Thanks for explaining!

benjaminwil (approved):
I think this is good.

Would love a second opinion! (cc @dometto @bartkamphorst)

lib/gollum/views/template_cascade.rb
```diff
@@ -0,0 +1,35 @@
...
+        @@template_priority_path
+      end
+
+      def template_priority_path=(path)
+        @@template_priority_path = File.expand_path(path)
+        @template = nil
+      end
+
+      def first_path_available(name)
+        priority = "#{template_priority_path}/#{name}.#{template_extension}"
```
dometto, line 14: Please use `::File.join()` here instead of the hardcoded slashes. I think in practice slashes work on Windows, but better to make sure! Same on the next line.

lib/gollum/views/template_cascade.rb
```diff
@@ -0,0 +1,35 @@
...
+        default = "#{template_path}/#{name}.#{template_extension}"
+        File.exists?(priority) ? priority : default
+      end
+
+      # Following methods should track lib/mustache/settings.rb from Mustache project.
```
dometto, line 19: I can find the `#template_file` method in https://github.com/mustache/mustache/blob/master/lib/mustache/settings.rb but not the `#partial` method. Is that in some other file?

dometto: Asking because without a reference to the Mustache method, what happens in the `#partial` method looks a bit obscure. Good to know what we should be mirroring, exactly!

dometto (commented):
Thanks for working on this @beporter and @benjaminwil!

beporter:
Still haven't forgotten about this, for the record. Just busy. 🤦‍♂️

lib/gollum/views/template_cascade.rb
beporter: No problem. This is how the parent Mustache class does it so I had a choice of whose patterns to respect.

lib/gollum/views/template_cascade.rb
beporter: Yeah that's a good point, and complicated a bit by the fact that the Mustache lib itself has two (and a half) of them:

- https://github.com/mustache/mustache/blob/070187/lib/mustache/context.rb#L26
- https://github.com/mustache/mustache/blob/070187/lib/mustache.rb#L177,L198 <--

The second link is the one getting overridden in this PR. This is confusing because in Mustache, the `settings.rb` file still just operates on the main `Mustache` class so how the _code_ is able to access the method is slightly disconnected from where the _file_ lives.

Updated the comments in the PR to point to the relevant methods more directly. 👍

beporter:
Rebased on the latest `master` and feedback addressed.

dometto (approved):

dometto:
Great feature, thanks for putting in the work @beporter!

If you would be so kind, could you perhaps write a few words about the new behavior [in the wiki](https://github.com/gollum/gollum/wiki/Custom-Styling-and-Views#customizing-views)?

The gollum gem is overdue for an update, so I'll cut a new gem release soon.

bartkamphorst:
Thanks @beporter !

beporter:
Hey I'm pretty much just grateful to encounter a GitHub project that isn't abandoned where the feature request sits unanswered for six years. 😂

I will definitely write something up for the wiki as soon as I can. 👍

PR #1734, opened 2021-06-26
Update test.yaml

dometto:
This should replicate the old Travis behavior, where PR's will give an alert when the tests for that branch don't pass yet.


benjaminwil:
Ah yes, this is what we were missing!

benjaminwil (approved):

PR #1735, opened 2021-07-07
Fix an IME rendering issue

yy0931:
I fixed a problem that occurs when inputting Japanese (and probably Chinese etc.). It is caused by a `padding` setting applied to `.form-control, .form-select` in primer.css overriding the ace editor's style.
I also replaced `File` with `::File` in lib/gollum.rb because they throwed errors during `bundle exec rake precompile`.

before:
before

after:
after


benjaminwil (approved):
Great! Thanks for including the before/after example.

dometto:
Thanks @yy0931! Looks good.

PR #1739, opened 2021-07-12
unescape page names in the preview tab

yy0931:
I fixed a problem that non-ascii characters in page names are not rendered correctly in the preview tab of the "Edit" page. The change is to add `decodeURIComponent()` to `location.pathname`.

You can also see the same problem using a page name "C#".

before


bartkamphorst:
Looks good to me. 👍 @dometto ?

dometto:
Agreed! Thanks again @yy0931

PR #1759, opened 2021-08-28
Fix broken History button when viewing historical deleted file

NikitaIvanovV:
This PR fixes a bug that made viewing history of deleted files impossible. It usually occured when in Latest Changes you clicked on one of the pages and then clicked the History button: if the page was not existent at that time, clicking on the History button redirected to the main page instead of showing commit history of the page.


ViChyavIn:
To clarify, I made two /history/ routes because I failed to make a proper regex to handle two cases when SHA is provided and when it's not. (Actually, I [made such a regex](https://regex101.com/r/ImS0LJ/2/) but when I used it I got this error: `Mustermann::CompileError: regular expression should not contain $`).

lib/gollum/app.rb
```diff
@@ -654,9 +667,9 @@ def update_wiki_page(wiki, page, content, commit, name = nil, format = nil)
       wiki.update_page(page, name, format, content.to_s, commit)
     end
 
-    def wiki_page(path, version = nil)
+    def wiki_page(path, version = nil, wiki = nil)
       pathname = (Pathname.new('/') + path).cleanpath
-      wiki = wiki_new
+      wiki = wiki_new if wiki.nil?
```
dometto, line 672: Why not just always do `wiki = wiki_new`? I see that you've added `wiki = nil` as a default parameter to this method, but if I see correctly that's not really doing much, as on https://github.com/gollum/gollum/pull/1759/files#diff-65eee83dd06f55c49a91cd71e716ae5e5ef09b353bf7472d4079ffdc175c395aR443-R444 you basically just pass on `wiki_new` as the parameter.

dometto (commented):
Thanks for submitting the fix and sorry for the slow reply! I was a bit confused by the two routes at first, but on second thought it looks pretty readable. Just one question in the comments. Also, is there an issue in the issue tracker that this solves? If so, please reference it in the PR title or body so the issue will auto-close :)

dometto:
Took the liberty to resolve the trivial merge conflict that had somehow cropped up.

More important question: it would be great to have a test for the new route/behavior!

lib/gollum/app.rb
NikitaIvanovV: On line 443 I already get a wiki instance using `wiki_new` in order to get a commit object on line 444. But `show_history` method needs the same wiki instance too, so I thought it was not very optimal to create another wiki instance inside the function if it was already created outside of it and can be easily passed to it. If it actually doesn't matter I can always do `wiki = wiki_new`.

NikitaIvanovV:
> Also, is there an issue in the issue tracker that this solves?
Not that I'm aware of. I stumbled upon this bug on wiki I maintain so I thought it'd be helpful to fix it in the original repo too.

NikitaIvanovV:
Added tests! To some reason adding another commit to revert.git broke `edits page header footer and sidebar` test so I had to tweak it too.

lib/gollum/templates/navbar.mustache
```diff
@@ -71,6 +71,7 @@
           </button>
           <a
             class="btn BtnGroup-item btn-sm hide-sm hide-md"
+            id="minibutton-history"
             href="{{edit_path}}/{{escaped_url_path}}"
             id="minibutton-edit-page"
```
benjaminwil, line 76: I believe it is invalid for HTML elements to have two `id`s defined. So if they are both actively being used (in JavaScript) we will either have to consolidate the functionality to one ID or use classes instead.

NikitaIvanovV: I searched for `minibutton-history` in the project files and it was only found in navbar.mustache, so it is not user in any JS code. I don't remember why I added that id, that must be my mistake. Will remove in the future commit!

benjaminwil (approved):

lib/gollum/app.rb
dometto: Thanks for the explanation!

dometto (approved):

dometto:
Thanks for the contribution and you patience @NikitaIvanovV!

PR #1773, opened 2021-10-22
Fix mathjax on edit and create pages

### LINKED ISSUE #1772: Mathjax not properly loaded on new pages and page edit
**Describe the bug**
Mathjax is not loaded when creating new pages. Custom mathjax config is not loaded on page edit.

**To Reproduce**
Try above.

**Expected behavior**
Mathjax should behave same on page view, page create and page edit.

**Environment**
```
Gollum 5.2.3
Using:
rugged 1.1.1
gollum-rugged_adapter 1.1.1
gollum-lib 5.1.3
Other renderering gems:
none
```

fhchl:
Fixes #1772

I don't have much ruby experience, but I think this patch works as expected. Apart from those that err on master, all test are running.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -365,6 +365,7 @@ module Precious
         @name = wikip.name
         @ext  = wikip.ext
         @path = wikip.path
+        @mathjax = wikip.wiki.mathjax
         @template_page = load_template(@path) if settings.wiki_options[:template_page]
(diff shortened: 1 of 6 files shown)


dometto:
You're right, it didn't fail on `master` until I recently merged a change to the Readme -- which makes it pretty clear that it must be a problem with the latest JRuby.

PR #1777, opened 2021-10-28
Reduce size of container image

rgov:
This change reduces the size of the Docker image from 446 MB to 126 MB (-71%). This can be tested using the public [`rgov/gollum`](https://hub.docker.com/repository/docker/rgov/gollum) image.

The new image also makes it possible to pass arguments to the `gollum` process (e.g., `docker run gollumorg/gollum --config=...`). Previously the image only passed `--mathjax`. (**Note:** After applying this change, users will have to explicitly pass `--mathjax`.)

The smaller image image is based on Alpine Linux instead of Debian Bullseye. It also uses a separate build stage for compiling and installing the gem dependencies, which are then copied over to the final image.

I tested the new image by browsing an existing wiki, creating a new page, editing it, previewing the edits, and uploading a file.

Note that `libicu` and `openssl` are both build-time dependencies that do not appear to be needed at runtime so they are no longer in the final image.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -34,5 +34,5 @@ jobs:
         run: echo ${{ steps.docker_build.outputs.digest }}
       - name: docker state
         run: docker image ls
-      - name: Run tests
+      - name: Run gollum as test
         run: docker run -e CI=true -w /app ${{ env.CI_IMAGE }} --versions
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -35,4 +35,4 @@ jobs:
       - name: docker state
         run: docker image ls
       - name: Run gollum as test
-        run: echo "exit" | docker run -e CI=true ${{ env.CI_IMAGE }} --irb
+        run: docker run -e CI=true ${{ env.CI_IMAGE }} --irb
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -35,4 +35,4 @@ jobs:
       - name: docker state
         run: docker image ls
       - name: Run gollum as test
-        run: docker run -e CI=true ${{ env.CI_IMAGE }} --h1-title
+        run: echo "exit" | docker run -a stdin -e CI=true ${{ env.CI_IMAGE }} --irb
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -35,6 +35,4 @@ jobs:
       - name: docker state
         run: docker image ls
       - name: Run gollum as test
-        run: docker run -e CI=true ${{ env.CI_IMAGE }} --versions
-      - name: Check git repo present
-        run: docker run -e CI=true --entrypoint git ${{ env.CI_IMAGE }} -C /wiki rev-parse
+        run: docker run -e CI=true ${{ env.CI_IMAGE }} --h1-title
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -34,7 +34,7 @@ jobs:
         run: echo ${{ steps.docker_build.outputs.digest }}
       - name: docker state
         run: docker image ls
-      - name: Check git repo present
-        run: docker run -e CI=true --entrypoint git ${{ env.CI_IMAGE }} -C /wiki rev-parse
       - name: Run gollum as test
         run: docker run -e CI=true ${{ env.CI_IMAGE }} --versions
+      - name: Check git repo present
+        run: docker run -e CI=true --entrypoint git ${{ env.CI_IMAGE }} -C /wiki rev-parse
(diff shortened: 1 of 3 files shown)


dometto:
I was trying to add in the deprecation warning based on @rgov's suggestion. This triggered the tests to run, with two failures. I reverted my changes, and the Docker build error persists. @rgov do you happen to know how to fix this? I will take of the JRuby failure.

rgov:
It is failing trying to run:

    docker run -e CI=true -w /app --entrypoint bundle gollum-ci-img exec rake

I am not familiar with what the intention of this step is, but I would guess that the CI step of running `bundle exec rake` just doesn't apply anymore.

The first stage contains things needed to build Gollum/its dependencies, and the final stage (which becomes the published image) contains only the installed software. So the Gemfile or Rakefile or whatever it is looking for may no longer be there.

Note that during the build we run `bundle exec rake install`, which means that if building the Docker image succeeds, then this command must have also succeeded, so I'm not sure what the value is in running it again.

Perhaps the `CI` environment variable causes it to behave differently? If so we can build two variants of the image customized by a build argument.

srbaker:
I think I've confirmed that this is caused by #1779 and should not hold up release.  I'm sure a smaller docker image is surely desireable for the kinds of people who like to use docker for things.

ahaerpfer:
I have another caveat here:

The biggest part of the image size reduction is due to the two stage build in the `Dockerfile`.  However, this PR additionally switches from a Debian base image to Alpine.  This indeed reduces the image size even a bit further but has the side effect of switching out glibc for musl.

I'd instead suggest to do the two stage build but stick with the Debian base image.  This will result in an image size of 177 MB (see [ahaerpfer/gollum](https://hub.docker.com/repository/docker/ahaerpfer/gollum/tags?page=1&ordering=last_updated). … vs. 126 MB for Alpine/two-stage and 465 MB original image size.

(I'd also *not* remove the `--mathjax` option because that would change default behaviour.)

rgov:
What concern do you have with switching from glibc to musl? Some software may be sensitive to minor variations between libc implementations but Gollum does not appear to be. Ruby is well-tested on Alpine.

ahaerpfer:
No real concerns. I have very limited experience with the Ruby runtime and can't actually judge how well-tested Ruby is on Alpine.  However, I have sometimes encountered issues with musl for libraries that are not pure Ruby/Python/… but where parts are e.g. written in C and need to be compiled.  That's why I usually try to stick with a glibc base image … to be future-proof and at the expense of slightly larger images.

On the other hand, when you are running many containers that all use the same Debian base image, even this difference in size does not really matter as the respective image layers are only stored once.

dometto:
Gollum does use gems like `nokogiri` (which relies on `libxml2` I think), other gems with C-extensions, and optionally uses external tools like `pandoc`. I have no experience with Alpine/`musl` so I have no idea how to judge the "risk" involved in this. If anyone has a clear idea of the best practices in this regard (@rgov?) I would be very happy for more information.

rgov:
From the [Docker documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#from):

>  We recommend the Alpine image as it is tightly controlled and small in size (currently under 6 MB), while still being a full Linux distribution.

Consequently, all major official Docker images support it—Postgres, Node, Ruby, Redis, Nginx, Apache, etc. Therefore it has gotten a lot of testing over the years and is perfectly stable.

The [functional differences](https://wiki.musl-libc.org/functional-differences-from-glibc.html) between glibc and musl are minor and judging by the [hundreds of packages on Alpine](https://pkgs.alpinelinux.org/packages), they don't seem to really matter much. Gollum, being a relatively simple piece of software, is not affected.

An Alpine-based image is not somehow unstable or experimental. A Debian-based image is mostly beneficial to a user who wants to make low-level changes to the container (adding custom package repositories or whatever) *and* is more familiar with Debian. Tweaking the Gollum config or installing additional gems would be the same either way.

dometto:
Thanks for the detailed explanation! I'm happy to go with Alpine, then, unless someone else feels very strongly.

ahaerpfer:
Fine for me, too.  No strong feelings here. And for anyone more happy with a Debian-based image (like me) it shouldn't be too hard to build it themselves. The main advantage in both cases is the huge reduction in image size. So feel free to close the other merge request (#1783).

rmbrad:
I think this is a good update, only thing I would change is adding `CMD ["--mathjax"]` to the final stage in the Dockerfile to maintain the current default behavior when not specifically overridden in the `docker run` command.

dometto:
Sorry for the long wait and thanks again for the fix!

dometto:
@rgov apologies if my attempt to fix the build steps swarmed you with test updates (I was assuming only I was receiving them, but just realised that may not be the case).

In any case, I have:

1. added `--mathjax` by default and included a deprecation warning
2. fixed the 'test' step in the workflow, so that we now don't run `bundle exec rake` as a test, but instead `gollum --irb` (which loads the repo, so that's also tested)

I'm going to assume this is ready to merge now, but if anyone has any feedback please let me know.

@magnusviri after this is merged feel free to further improve the Docker setup in the ways you were thinking of.

magnusviri:
@dometto, yes, I'll test this and rework my PR after the merge. I'll go ahead and close my PR.

PR #1783, opened 2021-12-10
Use multi-stage build in Dockerfile to reduce image size

ahaerpfer:
(Also allow for additional command options when running the image.)

This is an alternative to gollum/gollum#1777 and does *not* switch to an Alpine base image.  See the discussion there for details.


dometto:
Closing this as discussed in #1777, thanks for the feedback there and helping to improve our Docker setup, @ahaerpfer

PR #1784, opened 2021-12-17
Added docker environment variables for cmdline options, author, and email. Resolves #702

### LINKED ISSUE #702: Last edited by
I have a very basic question sorry for asking. When one edits a Gollum wiki, the page always says "last edited by anonymous" how do I set users so that I can know which user has edited that page?

magnusviri:
This PR adds docker support for several things.

- docker-compose file (easier to start a container with this)
- added environment variable for git author
- added environment variable for git email
- added environment variable for gollum cmdline options, e.g. "--h1-title --allow-uploads"

This was all adapted from https://github.com/pjeby/gollum-portable except this PR changes the existing setup very little.

To use the environment variables rename .env_example to .env and edit the variables. The docker-compose.yml file can also be edited but that is less desirable then editing .env.

I believe I've dotted all the i's and crossed all the t's. Please let me know if there's more I can do to get this PR accepted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Gemfile.lock
 !.sprockets*
 !lib/gollum/public/gollum/stylesheets/_styles.css
 !.github*
+.env
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,4 @@ RUN bundle exec rake install
 VOLUME /wiki
 WORKDIR /wiki
 COPY docker-run.rb /docker-run.rb
-ENTRYPOINT ["/docker-run.rb"]
+ENTRYPOINT "/docker-run.rb" $GOLLUM_OPTIONS
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,5 +25,5 @@ RUN bundle exec rake install
 
 VOLUME /wiki
 WORKDIR /wiki
-COPY docker-run.sh /docker-run.sh
-ENTRYPOINT ["/docker-run.sh"]
+COPY docker-run.rb /docker-run.rb
+ENTRYPOINT ["/docker-run.rb"]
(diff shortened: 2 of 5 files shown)


magnusviri:
I just noticed one thing I changed, which might not be desirable. The old docker-run.sh had '--mathjax' as a default option. Because I moved the options to an environment variable, this is no longer a default option in the docker container. However, it can still be enabled by editing the .env file and changing 

> GOLLUM_OPTIONS=""

to

> GOLLUM_OPTIONS="--mathjax"

magnusviri:
And I just noticed I missed one thing with the GOLLUM_OPTIONS and I added it. So it should all be good.

dometto:
Sorry for the slow response @magnusviri. Until recently we already had two PRs with changes to the Docker setup open. I have just closed one, which leaves this and #1777. In the discussion of #1777, we have decided to move to an Alpine based image, because of the much smaller image size. Perhaps the best way to go is for me to merge #1777 first, and then you can rebase this PR against the latest version of the Dockerfile?

Some other pointers:

* I think using an environment variable to pass on `--options` is a bit unconventional. I think it would be more desirable to allow the user to set such options as extra command line arguments to the `docker run` command. That means using `"$@"` in the Dockerfile. See #1777 for an example of this.
* Our current plan is to keep the default `--mathjax` for now but throw a deprecation message, and remove `--mathjax` as a default option in the next minor-level release. We'll implement this in #1777
* Currently you have `anon@anon.com` as a default email address. That's as good as any, but are there any reasons to go with a different name/email (such as `gollum@localhost` or some such)?

dometto:
By the way, a better way to change the default username/email address might be to just set the global git config variables (running the `git config` command from the Dockerfile). I think `rugged` will just use those settings if they are available, but please give that a test. :)

magnusviri:
I'm closing this PR and will rework it and re-submit it after #1777 is merged.

PR #1787, opened 2021-12-22
Take account of possible https referer in upload route. Resolves #1771

### LINKED ISSUE #1771: Files are uploaded to the wrong path with HTTPS reverse proxy
I have essentially the same issue as #1557 but the reporter closed it.

When I upload a file by dragging it onto the edit area, the following snippet gets inserted:

    [[/uploads/path/to/page/attachment.png]]

However this is not where the file ended up, and so it renders as `Image not found`.

If I browse the "Overview" structure I find that it has uploaded to `uploads/https:/servername/path/to/page/attachment.png` or in one instance `uploads/https:/servername/gollum/edit/path/to/page/attachment.png`. These seem pretty clearly wrong?

Perhaps it is related to this logic here:

https://github.com/gollum/gollum/blob/75173890727255940d260847edcf7b9cf793f961/lib/gollum/app.rb#L242-L251

```
Gollum 5.1.2
Running on: x86_64-linux with Ruby version 2.7.2
Using:
rugged 0.99.0
gollum-rugged_adapter 1.0
gollum-lib 5.0.6
Other renderering gems:
none
```

dometto:


dometto:
@rgov explains the underlying issue in #1771:


> The issue involves having a frontend web server that serves HTTPS while Gollum serves HTTP:

> dir = request.referer.sub(request.base_url, '')

> Because request.base_url starts with http:// and request.referer starts with https://, the substitution does not work.

benjaminwil (approved):

PR #1788, opened 2021-12-22
Release preparation: update readme, rakefile and changelog

dometto:
Added a Dockerhub badge to the Readme to make more clear what the official Docker image is.

I think we should restructure the Installation instructions in the Readme to reflect that there are multiple ways to install/run Gollum: Docker, `apt`, `gem install`, etc.

We should also have a Release workflow to automate pushing a Docker image tagged for specific releases. Right now we only have one `latest` image based on `master`, which may in some cases be unstable.

I've tried to quickly write up something on Docker usage [in the wiki](https://github.com/gollum/gollum/wiki/Gollum-via-Docker), but all this could use much more help. Advice welcome!


benjaminwil (approved):

dometto:
I've updated the instructions in the wiki: https://github.com/gollum/gollum/blob/master/README.md

The [DockerHub info page](https://hub.docker.com/r/gollumwiki/gollum/) now also links to the relevant info.

I'll rewrite the Readme's installation section to include Docker, and make it a bit more concise.

dometto:
I've added some more goodies to this PR. To help with the new automatic GitHub releases action, I've added `LATEST_CHANGES.md`. This should describe the changes to the latest version only, which will then be displayed in the GitHub release.

In the `Rakefile`, I have added a task that:

1. Checks whether `LATEST_CHANGES.md` and `HISTORY.md` are present and non-empty
2. Checks whether the current version of Gollum (defined in `gollum.rb`) is already described in `HISTORY.md`
3. Checks whether the version in `LATEST_CHANGES.md` corresponds to the current version of Gollum
4. If these sanity checks pass, it prepends the contents of `LATEST_CHANGES.md` to `HISTORY.md`, thus bringing the changelog up to date

This new task is automatically invoked when running `rake release`, so it has the added benefit of forcing us to keep the changelog up to date. :)

I've configured the GitHub release action to create a draft release for now, so we can see how it works the first time before turning on full automagic mode.

dometto:
Rebased on `master` to get the proper test build matrix.

benjaminwil (approved):
This all seems great.

PR #1789, opened 2021-12-22
Added docker environment variables for author and email

### LINKED ISSUE #702: Last edited by

magnusviri:
Resolves #702.

This PR adds environment variable for git author for git email. It adds support for author and email by editing the git config.


magnusviri:
I just added a change to the docker-compose.yml file so it uses the image on hub.docker.com instead of building an image from the current directory.

docker-run.sh
```diff
@@ -5,6 +5,14 @@ if [ ! -d .git ]; then
     git init
 fi
 
+# Set git user.name and user.email
```
dometto, line 8: Nice and clean way of handling this now!

.gitignore
```diff
@@ -9,3 +9,5 @@ Gemfile.lock
 !.sprockets*
 !lib/gollum/public/gollum/stylesheets/_styles.css
 !.github*
+!.env_example
+wiki
```
dometto, line 13: Think this should not be here, right? We're not really expecting the user's wiki to be inside the gollum directory, I think.

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3"
+services:
+  gollum:
+    image: "gollumwiki/gollum"
+    ports:
+      - "4567:4567"
+    volumes:
+      - "./wiki:/wiki"
```
dometto, line 8: If I understand well this assumes the wiki will be in `.wiki/` rather than `./`. No strong feelings either way, but it should be documented. Also, just to make sure, this behavior will be specific to using Docker with the `compose` file, right? So if you just pull the `gollumwiki/gollum` image and `docker run` it, it will still expect the wiki to be in `./`?

.env_example
```diff
@@ -0,0 +1,2 @@
+GOLLUM_AUTHOR_USERNAME="Anonymous"
```
dometto, line 1: Maybe add comments to this file to explain its usage? Is this just for use with `docker compose`, or does a 'normal' `docker run` also make use of the `.env` file?

.gitignore
magnusviri: I have never worked with automated docker image building, so I'm not used to how this works. Your comments got me to think about it more and I have a better understanding.

It only makes sense to include the docker-compose.yml and .env_example files in the repository for building the docker image from the current directory. The purpose of that is to test everything.

That would mean the wiki directory would have to be located somewhere else, like in a folder named wiki. Because this is only for testing, not deploying, one could argue to not include it in the project at all.

If you would like to include the 2 files for testing purposes only, let me know and I'll tailor it for that purpose only.

Otherwise, I should just remove docker-compose.yml, .env_example, and the changes to .gitignore.

.env_example
magnusviri: .env is only used with docker compose.

docker-compose.yml
magnusviri: See my comment about testing. To test, it builds from the current directory and it wont pull from gollumwiki/gollum. My original commit was building. But then I got confused and I changed it to pull from gollumwiki/gollum. But that actually doesn't make sense.

benjaminwil (commented):
Hello! Thanks for your recent contributions.

After reviewing this pull request, I have some questions. I don't think that most end users of Gollum should need to clone this repository. So they would not benefit from the `docker-compose.yml` you've added here.

After reviewing the source code for some other Ruby web apps that provide Dockerized versions of themselves, I noticed that they do not provide `docker-compose.yml` at the root of the project. Take [Huginn](https://github.com/huginn/huginn) for example.

It is more conventional for web apps like Gollum and Huginn to provide documentation with a typical `docker-compose.yml` and any required `.env` in documentation somewhere. [Huginn has a Docker section in their README](https://github.com/huginn/huginn#docker).

I'm not opposed to the idea that this repository could have its own `docker-compose.yml` and some kind of env file provided for developers. But I don't think that is the goal of your PR here.

Hopefully what I'm saying makes sense. And I'm of course happy to discuss this more!

dometto:
Thanks for doing the reseach @benjaminwil. What about including an example `docker-compose.yml` and the accompanying `.env` file in the `contrib` directory ([here](https://github.com/gollum/gollum/tree/master/contrib)) and explaining it's usage in the wiki?

magnusviri:
Yes, I agree with you both. I've removed the docker-compose.yml, .env_example, and the changes I made to .gitignore.

benjaminwil:
Thank you again for this contribution. It's a great way to handle `git config` within Docker. And we should definitely merge your `docker-compose.yml` and `.env-example` files intro the `contrib/` directory as @dometto suggests.

I will merge this PR as is when our CI suite passes.

benjaminwil (approved):

benjaminwil:
I've resolved our CI issues on `master`, so I'm going ahead and merging.

We can handle the other changes discussed on this pull request in future change requests.

PR #1793, opened 2021-12-23
Add arm64 Docker build

magnusviri:
This adds an arm64 build for Docker.

This PR only has 1 commit.

I feel kind of sheepish submitting this pull request because I don't know if it works, especially the tags. I don't really know how to test this though.

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -31,8 +31,9 @@ jobs:
           context: ./
           file: ./Dockerfile
           builder: ${{ steps.buildx.outputs.name }}
+          platforms: linux/amd64,linux/arm64
           push: true
-          tags: ${{ env.DEPLOY_NAME }}
+          tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
       - name: Image digest
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -1,7 +1,7 @@
 name: Deploy docker
 on: [push]
 env:
-  DEPLOY_NAME: gollumwiki/gollum:latest
+  DEPLOY_NAME: magnusviri/gollum:latest
 jobs:
   build:
     runs-on: ubuntu-latest


dometto:
Thanks! Some questions:

* Will this automatically take care of using an `arm` version of the base Alpine image, or does that require changes to the Dockerfile?
* What is `steps.meta.outputs.tags` supposed to refer too?

Perhaps you could try to test the deploy step and the resulting image on your own fork of gollum? I think all you'd need to do is activate GitHub Actions, and add the required GitHub secrets (e.g. your Dockerhub credentials).

magnusviri:
I've used github actions to push to other services (pypi.com), so I'm familiar with the ideas. But I have never pushed to hub.docker.com, so I don't know how that specific service works. And I don't know the answers your questions.

I looked at another project that builds an x86 and an arm image and I compared their github workflow with yours and I found 2 changes that looked relevant. That's what is in my pull request. I didn't know you didn't have much docker experience and so I was hoping you'd know if it was correct. But I now see that the original workflow file was added by the person known as darless.

I will attempt to test this. Pushing directly from github requires a paid account or I can apply to join their open source community. I've applied and hopefully will get accepted. If I get accepted I'll test it and answer your questions. If I don't get accepted I'll post back to this thread to see if I can test this some other way.

benjaminwil:
If we can verify this is working, I think this is great! I am assuming that the change  `DEPLOY_NAME: magnusviri/gollum:latest` will be reverted after a successful test?

magnusviri:
Yes, `DEPLOY_NAME: magnusviri/gollum:latest` will be reverted. I am still waiting to hear if I can get a free account with Docker hub.

dometto:
Hi @magnusviri! Any update on this? I've juts made some changes to `docker-deploy.yml` in order to also trigger a deploy when a new tagged version of gollum is released. So there is now a merge conflict for this PR, but nothing that should be too difficult. :)

.github/workflows/docker-deploy.yml
```diff
@@ -31,9 +31,10 @@ jobs:
           context: ./
           file: ./Dockerfile
           builder: ${{ steps.buildx.outputs.name }}
+          platforms: linux/amd64,linux/arm64
           push: true
-          tags: ${{ env.DEPLOY_NAME }}
+          tags: ${{ steps.meta.outputs.tags }}
```
dometto, line 36: Not sure I understand this change. What is the value of `steps.meta.output.tag`?

In my recent update to this file, I've made the value of `env.DEPLOY_NAME` conditional (it's set to `master` on a `push` event, and to e.g. `v5.3` for a `release` event. So this line will take some tweaking. I'm not sure if we can have the same tag for two different images (`arm` and `amd`)? If not, we should make sure `tags` is set to a sane value combing the `master`/version tag, and a tag indicating the plafform.

magnusviri: I just looked and I did this wrong. 'steps.meta.outputs.tags' depends on a previous step which I didn't add, docker/metadata-action. If we add docker/metadata-action then the tags and labels are much handled automatically and can have many different values. I'm actually not sure if it is actually required or if it's worth it. This is [the documentation](https://github.com/docker/metadata-action#tags-input). Because I've never pushed an image to hub.docker.com, I'm still unsure how this will all work or what tags are needed.

I don't know how you feel about pushing to docker hub and seeing if it works. I don't know if there is a test docker hub to try this on. If it were me, I'd just push this to docker hub and figure out the right way, but Docker hasn't looked at my open source application yet. They sent an email basically saying they'll get to it eventually.

dometto: For me it would be fine to test it on Dockerhub! Also looks like a good idea to use the `docker/metadata-action` action. Do you want to try to integrate that in this PR? If so, the goal would be to push versioned tags (e.g. `v5.2.3`) to Dockerhub on GitHub `release` events, and use the `master` tag when something is pushed to the GitHub `master` branch (which I now implemented without the `metadata-action`, see [here](https://github.com/gollum/gollum/commit/0c36cdf5ba080f0c5e8e925d1a446fda2fbcfc8b)). If that would cost you too much time, just let us know!

magnusviri:
I found out that I can push docker images to the GitHub Container Registry (ghcr). I can build the image and push it to ghcr but when I pull it and run the container has errors. Already my changes are significant enough that I'm going to close this pull request now and open a new one when I get everything working completely.

Issue #1779, opened 2021-11-06
JRuby: headers and anchors not displayed properly


dometto:
[Two tests are failing on JRuby](https://github.com/gollum/gollum/runs/4125783787?check_suite_focus=true). After some research, I figured out that the culprit is the latest `nokogiri` version for JRuby. Long story short, it appears to be outputting some invalid XHTML in the [toc filter](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/toc.rb#L127), when adding anchors to header elements. This appears to be a bug in nokogiri, which I've reported [here](https://github.com/sparklemotion/nokogiri/issues/2355).

As a workaround, affected users can revert back to `nokogiri` ~`v1.11.7`~ `v.1.12.4` or earlier.

flavorjones:
Please note that this change in behavior was introduced in Nokogiri v1.12.5, so reverting to v1.12.4 or earlier should be sufficient to work around it for now.

dometto:
This should be fixed in the next release of nokogiri. Will leave this open for now in case users look for it.

dometto:
Thanks to @flavorjones by the way!

daniol:
Downgrading just nokogiri is not enough. After downgrading with:

```
gem uninstall nokogiri
gem install nokogiri -v 1.12.4
```

I ran in the following errors when starting gollum:

```
LoadError: no such file to load -- nokogiri
  require at org/jruby/RubyKernel.java:1017
  require at C:/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:160
   <main> at C:/jruby/lib/ruby/gems/shared/gems/loofah-2.12.0/lib/loofah.rb:4
  require at org/jruby/RubyKernel.java:1017
  require at C:/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
   <main> at C:/jruby/lib/ruby/gems/shared/gems/gollum-lib-5.1.3-java/lib/gollum-lib.rb:18
  require at org/jruby/RubyKernel.java:1017
  require at C:/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
   <main> at C:/jruby/lib/ruby/gems/shared/gems/gollum-5.2.3/lib/gollum/app.rb:5
  require at org/jruby/RubyKernel.java:1017
  require at C:/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:85
[...]

Issue #1795, opened 2021-12-29
Drop Ruby 2.4 support


benjaminwil:
Ruby 2.4.x [has reached end of life](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/).

I would like to propose we stop supporting it and drop it from our CI matrix.

I am curious if there is a specific reason we still support it? After a search, I didn't find any specific rationale.

dometto:
All good to drop it! No specific reason why we are still testing it, other than a vague idea that some backward compatibility is a good thing. :) But 2.4 is very old now.

benjaminwil:
I've dropped Ruby 2.4 from our GitHub Actions workflow version matrix.

There's no explicit mention of supported Ruby versions in the README, nor mentions of Ruby versions anywhere else. So I think we can just close this issue now. :+1:

PR #1796, opened 2021-12-29
Improve test suite and CI run performance

benjaminwil:
This pull request makes our GitHub Actions test suite workflow run faster. I took this opportunity to also simplify the workflow YAML file.

I started doing this work because of JRuby CI failures. (See #1779.) But then noticed some low-hanging fruit that would improve our test suite runs and maintenance in general.

### DIFF
--- a/test/test_app.rb
+++ b/test/test_app.rb
@@ -39,31 +39,29 @@ context "Frontend" do
     assert_match /<pre><code>one\ntwo\nthree\nfour\n<\/code><\/pre>\n/m, last_response.body
   end
 
-  def nfd utf8
-    TwitterCldr::Normalization.normalize(utf8, using: :nfd)
-  end
-  
   test 'mathjax assets are served' do
     get '/gollum/assets/mathjax/MathJax.js'
     assert last_response.ok?
   end
 
   test "UTF-8 headers href preserved" do
-    page = 'utfh1'
-    text = nfd('한글')
+    page_content = <<~TEXT
+      ## 한글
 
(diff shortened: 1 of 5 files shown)


benjaminwil:
If there is not a good reason to continue support Ruby 2.4, we can also drop 2.4 from CI runs as part of this PR. See #1795.

test/test_app.rb
```diff
@@ -39,31 +39,29 @@
...
-    page = 'utfh1'
-    text = nfd('한글')
+    page_content = <<~TEXT
+      ## 한글
 
-    # don't use h1 or it will be promoted to replace file name
-    # which doesn't generate a normal header link
-    @wiki.write_page(page, :markdown, '## ' + text,
-                     { :name => 'user1', :email => 'user1' });
+      Test page "utfh1" content.
```
dometto, line 51: Much more readable, thanks!

test/test_app.rb
```diff
@@ -39,31 +39,29 @@
     assert_match /<pre><code>one\ntwo\nthree\nfour\n<\/code><\/pre>\n/m, last_response.body
   end
 
-  def nfd utf8
```
dometto, line 42: I think this means we can remove `twitter_cldr` as a development dependency from `gollum.gemspec`, too.

dometto:
See comments, but looks great! Fine to also remove 2.4 testing for further simplification

test/test_app.rb
benjaminwil: Oh wow! Thanks for noticing that.

PR #1797, opened 2022-01-01
Start localizing the views

benjaminwil:
This pull request localizes many of Gollum's views so they can start being translated into other languages. (Although users would not be easily be able to change languages yet.)

The remaining views require some additional work to be done in our `I18n` wrapper for Mustache templates, so I'll add those in a future PR.

The `test_app` file caught some of my mistakes as I swapped out hard-coded English for translation strings. However, I will need to manually test all of the views to ensure I haven't missed any minor glitches that could come from this change.

## Manual tests

I've manually tested these changes successfully:

- [x] Compare views
- [x] Error views
- [x] History views
- [x] Overview views
- [x] Search partials
- [x] Layout

But the following views still have issues:

- [x] Latest changes views: Pagination "previous" and "next" labels do not appear at all.


lib/gollum/templates/commit.mustache
```diff
@@ -3,8 +3,14 @@
     <h1 class="header-title text-center text-md-left py-4">
       {{message}}
     </h1>
-    {{author}} commited {{authored_date}}
-    <span class="px-2 float-right">commit <code>{{version}}</code></span>
```
dometto, line 7: Should `commit` here be `{{t.commit}}`?

dometto (commented):
Great work! Let us know if we can help with localization efforts somehow.

lib/gollum/templates/commit.mustache
benjaminwil: Yes! Thank you.

dometto (approved):

benjaminwil:
I plan to manually test these changes later this week.

After testing, I will merge if no issues arise.

dometto:
> I plan to manually test these changes later this week. After testing, I will merge if no issues arise.

Please do! I may cut a new release before the end of the week, so maybe that's good qua timing: if the localization efforts end up introducing some bugs, we have before the next release to fix them. :)

benjaminwil:
After rebasing against the latest changes on `master`, I've encountered new errors in the CI runs using JRuby and Ruby 3+.

At first glance, they don't seem related to my changes. But I need to look more deeply into this.

benjaminwil:
Some updates:

- I dropped the commit where I localize the commit views. I did this in order to put off resolving a git conflict that arose when #1653 was merged recently.
- I found one issue in my manual tests. I've added it as a to-do item in the PR description.

benjaminwil:
I rebased against `master` and added a single commit. The single commit resolves an issue where I wasn't resetting the `I18n.load_path` after testing the locale helper methods with test-specific translation YAML files, which caused order-dependent test failures and errors.

benjaminwil:
I've added one more commit that lets us use translations strings outside the scope of a view class. This is helpful for view partials like `pagination.mustache`, which is rendered via many other partials across Gollum's frontend. See the commit and its commit message for more information.

benjaminwil:
@dometto This doesn't localize all of the views yet, 
but I think I should merge this now.

I've manually tested this does not break any pages, that are localized, but it would be great to merge this and release it in case users find that's not true.

If you disagree feel free to revert.

PR #1798, opened 2022-01-11
Remove explicit nokogori version requirement. Resolves #1779

### LINKED ISSUE #1779: JRuby: headers and anchors not displayed properly

dometto:
No longer needed after nokogiri's recent release, see https://github.com/gollum/gollum/issues/1779

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [jruby-9.2.18]
+        ruby: [jruby-9.3.2.0]
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


dometto:
Needed to upgrade JRuby on CI to the latest version because the latest `nokogiri` requires ruby `>= 2.6`. I've also updated the version requirement in the `gemspec`. Not sure if we should do a bugfix release?

dometto:
Just noticed we actually have a whole bunch of unreleased bugfixes, so a new release is worth it anyway! However, I think [this change](https://github.com/gollum/gollum/commit/75173890727255940d260847edcf7b9cf793f961) means we should make it a minor rather than patch level release?

PR #1799, opened 2022-01-11
Support versioned Docker builds

dometto:
This PR

1. Also builds/deploys the Docker image on new releases, not just on pushes
2. Conditionally sets the DockerHub tag to either `master` in case of a push or the tag name (e.g. `v5.2.3`) in case of a release

I have stopped using the label `latest` because many seem to recommend against using that label, since it is confusing: https://vsupalov.com/docker-latest-tag/

Specifically, in our case, it could be understood to refer to the lastest *stable* version. But `master` is sometimes not stable (for instance when we add features that also require changes to gollum-lib). I think using the tag `master` for the Docker image is more descriptive, but open to other suggestions! (Could also be e.g. `nightly-2021-01-11`, etc.)

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -1,12 +1,20 @@
 name: Deploy docker
-on: [push]
-env:
-  DEPLOY_NAME: gollumwiki/gollum:latest
+on:
+  push:
+    branches:
+      - master
+  release:
+      types: [published]
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


dometto:
Protip: you can `brew install actionlint` (or equivalent) and run `actionlint` inside your gollum git repo root to automatically get a validity check for the workflow `.yml` files. See https://github.com/rhysd/actionlint

benjaminwil:
Amazing!

PR #1801, opened 2022-01-13
Set encoding explicitly in bin/gollum

### LINKED ISSUE #1699: Gollum reports encoding error on '--'
**Describe the bug**
If a markdown file contains a double dash '--' it can not be loaded and reports an encoding error.
File is UTF-8. Removing the double dash allows rendering of the page. Writing the double dash as `\-\-` also works.

**To Reproduce**
```bash
mkdir test
cd test
git init
cat > test.md << EOF
# Test

ab --ab
EOF
git add test.md
git commit -m "test"
gollum
```
go to `localhost:4567/test.md

**Expected behavior**
Should show rendered markdown page.

**Environment Info**
```
$ gollum --versions
Gollum 5.2.1
Running on: java with Ruby version 2.5.7
Using:
rjgit 5.7.0.2
gollum-rjgit_adapter 0.6
gollum-lib 5.1
Other renderering gems:
none
```

dometto:
Resolves #1699

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/usr/bin/env -S ruby -Eutf-8
 
 $:.unshift File.join(File.dirname(__FILE__), *%w[.. lib])
 


bin/gollum
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/usr/bin/env -S ruby -Eutf-8
```
benjaminwil, line 1: Wow! Wouldn't have thought of that.

Issue #1306, opened 2018-05-13
Image corruption with --alow-uploads


NHALX:
All images uploaded through the web interface are being corrupted. 

- Text files upload ok.
- Image display is ok for files uploaded and committed through other means (shell access).

Versions: 
Gollum 4.1.2, git 2.14.1, slackware 14.2, ruby 2.2.10p489 (2018-03-28 revision 63023) [x86_64-linux]

bartkamphorst:
Cannot reproduce this on `5.x` gollum.

bartkamphorst:
@NHALX You mentioned your git version, but not the git adapter you were using. Did you use gollum's default grit adapter?

NHALX:
It was most likely the default. It's been a while, but I don't recall doing any special configuration.

bartkamphorst:
Actually, I've just now come across this in `5.x`.

dometto:
Any difference between uploading via the button or drag-and-drop?

dometto:
> Actually, I've just now come across this in 5.x.

With grit or rugged, though?

bartkamphorst:
> With grit or rugged, though?

RJGit (on Windows).

bartkamphorst:
Upload relies on `committer.add_to_index` to commit uploaded files to the repository, but this method attempts to normalize the files by calling the following `normalize` method:
```ruby
def normalize(data)
  data.gsub(/\r/, '')
end
```
This works fine for text files, but my hunch is that this code is eating up some bytes from the binary files that are uploaded. And the kicker is that any errors are suppressed (from `committer.rb`):
```ruby
rescue ArgumentError => err
  # Swallow errors that arise from data being binary
  raise err unless err.message.include?('invalid byte sequence')
end
```

bartkamphorst:
The `ptools` gem uses this code to guess whether a file is a binary file or not:
```ruby
  def self.binary?(file, percentage = 0.30)
    return false if image?(file)
    bytes = File.stat(file).blksize
    bytes = 4096 if bytes > 4096
    s = (File.read(file, bytes) || "")
    s = s.encode('US-ASCII', :undef => :replace).split(//)
    ((s.size - s.grep(" ".."~").size) / s.size.to_f) > percentage
end
```
(where `image?(file)` just does a check based on file extension.)

@dometto Perhaps we can use the same heuristic for determining whether or not to pass `data` to `normalize`?

dometto:
[...]

PR #1802, opened 2022-01-20
Restore the normalize check that appears to be inadvertantly removed

manofstick:
Originally there was a problem #1306 for image uploads. I installed the latest version of gollum and noticed that this was still a problem, although....

@bartkamphorst had provided a fix #1353

But it appears that this was inadvertently destroyed by @dometto in #1369 (I don't think he intended to remove this, because he didn't remove the variable `normalize` which was created, but rather when he consolidated the code to `write_file` he just accidentally removed the option that was inline, as he'd created a new one with large scope, but didn't populate it with the `normalize` variable)

So this is a very trivial PR, it's just adding the normalize variable into the options.

So compare the commits: https://github.com/gollum/gollum/commit/ca9702088f2c5cec86d99f9c4fe019dccc2bd637 and https://github.com/gollum/gollum/commit/b40a344c495c4317b2aadc32cdf7a5358c3d5ef7

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -274,7 +274,7 @@ module Precious
           options.merge! author
         end
 
-        normalize = Gollum::Page.valid_extension?(fullname)
+        options[:normalize] = Gollum::Page.valid_extension?(fullname)
 
         begin
           wiki.write_file(reponame, contents, options)


manofstick:
@dometto I'll have a look, but I'm not particularly familiar with Ruby - the last time before Gollum I played with it was in pre-Rails days, a couple of decades ago! So happy to crank together some one-liners, but we'll see if I can make a good and useful test... Hmmm...

dometto:
Took the liberty of adding the test myself. Thanks again for the fix @manofstick!

PR #1805, opened 2022-02-21
Switch from TestUnit to Minitest

benjaminwil:
This pull request changes the test suite to use Minitest instead of the current test framework, TestUnit. Some of the changes I made to make this switch successful are a bit obscure. If you view the changes commit-by-commit I hope that my commit messages explain the issues clearly.

I think that there are a couple of benefits that come with these changes:

**Minitest is a complete testing framework, where as TestUnit is a unit testing framework.**

Without making substantial changes to our existing test suite, I've been able to switch over to a different test framework that provides a larger featureset. For example, it has first-class support for Capybara acceptance testing, which is something we've been interested in adding anyway.

**Without additional configuration, Minitest runs our tests in a random order.**

It seems as though TestUnit was running our tests in a deterministic order. Switching to Minitest here forced me to make some changes that were clearly order-dependent test failures. I think having our tests run in a random order can help us improve our test suite long-term.

### DIFF
--- a/bin/gollum-migrate-tags
+++ b/bin/gollum-migrate-tags
@@ -11,8 +11,11 @@ migrate_options = {
   :hyphenate => true
 }
 
-def setting(const)
-  Object.const_defined?(const.upcase) && Object.const_get(const.upcase)
+def setting(variable_name)
+  class_variable_name = :"@@#{variable_name.to_s}"
+
+  Object.class_variable_defined?(class_variable_name) &&
+    Object.class_variable_get(class_variable_name)
 end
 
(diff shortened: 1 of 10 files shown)


benjaminwil:
Now running this against CI, I discovered one other order-dependent test failure in JRuby only. This has now been fixed.

dometto:
Think we should go ahead and merge this, as you report it also fixes #1817 :)

benjaminwil:
All right, here goes !! Apologies to anyone who may have to rebase after this.

PR #1806, opened 2022-02-25
raspberry pi compatibility

cooperspencer:
I added platforms to the `build and push` task and added libc6-compat to the docker container.

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -45,5 +45,6 @@ jobs:
           tags: ${{ env.DEPLOY_NAME }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+          platforms: linux/amd64, linux/arm64, linux/arm/v7, linux/arm/v6
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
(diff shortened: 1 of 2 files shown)


benjaminwil (approved):

benjaminwil:
Thank you!

dometto:
Sadly it looks like that the Docker Deploy action is failing. The nokogiri build (as part of the `bundle install` step) is failing on `arm`: https://github.com/gollum/gollum/runs/6248371141?check_suite_focus=true

@cooperspencer have you successfully built gollum/nokogiri on `arm` locally? If you have no idea what's causing this, I can investigate myself. :)

cooperspencer:
@dometto I just rebuild the docker container on my pi and it runs fine. So I am a bit clueless right now tbh...

benjaminwil:
I noticed this earlier and was planning to try doing a manual build myself. If I can't get it working I might revert this PR.

benjaminwil:
I was also able to build the container successfully for platform `linux/arm64`, which makes me think there is a package missing from the GitHub Actions environment. I'm just not sure how to identify what package that might be?

cooperspencer:
If I read the error message correct, the problem is with arm/v6 and v7. If we remove those architectures from the pipelines it should be fine.

benjaminwil:
Thank you!

PR #1811, opened 2022-03-20
Remove fields from gemspec that should not be set

mishina2228:
RubyGems says not to set the value to `date`, `specification_version` and `rubygems_version`.
https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L1677-L1682
https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L750-L755
https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L536-L541

This PR removes those fields from gemspec.
And does the following:
- Fix gemspec task not to set date field.
- Remove unused codes related with rubyforge_project from gemspec task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,5 @@
 require 'rubygems'
 require 'rake'
-require 'date'
 
 #############################################################################
 #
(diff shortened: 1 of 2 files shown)


dometto:
Thanks!

PR #1824, opened 2022-05-08
Start using Yarn to manage JavaScript dependencies

benjaminwil:
This pull request takes the minimum steps required to start managing JavaScript dependencies using a JavaScript package manager. In short:

- We can stop committing third-party JS libraries to the codebase. For more information, read the longer description below, which comes from the first commit message.
- Gollum end users can continue using the provided, compiled Sprockets as normal. These will continue to be committed to the codebase.

--- 

Gollum has included JavaScript assets by committing the source code to the `lib/gollum/public/gollum/javascript/` directory and including them for compilation in the Sprockets asset manifest file at `lib/gollum/public/gollum/javascript/app.js`. This has been a
reasonable way to deal with third-party JavaScript, but there are a few downsides:

  - It's a burden to find out if JavaScript dependencies have been updated and require updating in Gollum.
  - It doesn't give us good visibility into the JavaScript dependencies required by our dependencies.
  - It forces us to commit external code to our repository, which can make our developer tools more difficult to configure or use. For example: when I search for key words in the repository using Ripgrep, I often get "garbage" results from minified JavaScript.\*

\* Note: We still commit the static, compiled assets that Sprockets compiles when we run `bundle exec rake precompile`. We just don't have to commit source code from external projects.

Managing JavaScript dependencies via a JS package manager can resolve all of these issues.

This commit allows us to manage JavaScript dependencies using the Yarn package manager for Node JS modules[1]. I chose Yarn over NPM for one reason: Yarn is the JavaScript package manager that Rails uses by default. So many Ruby developers will already be familiar with Yarn.

To demonstrate how this can change how we manage JS assets in Gollum, I've configured Yarn and started to manage the `mousetrap` dependency with it. I chose `mousetrap` to start because it's a smaller, mostly uncomplicated dependency. I was easily able to manually test that `mousetrap` is still working after re-compiling the assets.

Hopefully this gives anyone reading enough context to jump in and start moving our third-party JS assets out of the codebase.

[1]: https://yarnpkg.com/


benjaminwil:
Ah yes, I should make this clearer in my PR description.

There are no workflow changes as far as the end user is concerned. In the changeset, when I run `bundle exec rake precompile`, the `mousetrap` dependency is still compiled and included in the Sprockets assets, the same as it always has been.

The changes only effect developers who change assets and need to recompile them. They will need to have Node JS and `yarn` installed.

Perhaps I should even add another commit in which running `bundle exec rake precompile` warns the users, or even errors out, if `yarn` is not installed or `yarn install` hasn't been run recently.

benjaminwil:
Updated the description.

dometto:
I think this functionality is great and the implementation looks clean to me. I like the idea of just doing `mousetrap` in this PR, and then gradually disentangling the other dependencies. So from my end this looks ready to merge!

dometto:
Ah, sorry for creating the merge conflict with the latest release! Since it's only the static assets I hope it's easy to fix. If you could merge this that would be great, I'll get to work on implementing support for mermaid (#1118).

benjaminwil:
@dometto I just added one new commit that changes the `precompile` task. It now exits early, before precompilation, if the developer's machine cannot successfully run `yarn install`.

But I've resolved the merge conflicts. If you're okay with this change to the precompilation task, could you merge?

PR #1826, opened 2022-05-24
Update octicons

dometto:
See https://github.com/gollum/gollum-lib/pull/420

- [x] Compile static assets


dometto:
This PR now utilizes the newly released `gollum-lib` `v6.0`. That release drops the dependency on `octicons` to avoid the dependency loop between `gollum` and `gollum-lib`. I've updated `octicons` in two steps:

* First to 17.0, which required renaming the `clippy` icon to `paste`, and `file-directory` to `file-directory-fill`
* Then to `~> 19.0` (latest), which according to the changelog does not introduce any breaking changes relevant to us.

benjaminwil (approved):
✨

PR #1831, opened 2022-06-08
Refactor logic for determining per page upload location. Resolves #1830

### LINKED ISSUE #1830: Files uploaded with '--allow-uploads page' and '--base-path' are stored in the wrong directory
**Describe the bug**
I have a gollum instance within a Docker environment that is started with the following parameters:
```
gollum --bare  --allow-uploads page --base-path myBasePath
```
When I drag&drop a file `MyFile.pdf` in the editor for the wiki page `/Projects/SomeStuff.md` it creates a link to 
`[[MyFile.pdf | /uploads/Projects/SomeStuff/MyFile.pdf]]`.
However, the file is actually saved under
`/uploads/myBasePath/gollum/edit/Projects/SomeStuff/MyFile.pdf`

The URL of the editor page is thereby:
`https://example.org/myBasePath/gollum/edit/Projects/SomeStuff.md`

**Expected behavior**
I expected the file to be saved under the path given in the generated link.

My best guess is that the base-path parameter impacts the generation of the correct path, because the actual upload-path contains `myBasePath/gollum/edit/` much like the URL of the editor page.

**Environment Info**
```
Gollum 5.3.0
Running on: x86_64-linux-musl with Ruby version 2.7.6
Using:
rugged 1.1.1
gollum-rugged_adapter 1.1.2
gollum-lib 5.2
Other renderering gems:
org-ruby 0.9.12

```

Dockerfile
```
FROM ruby:2.7-alpine AS builder

RUN apk add \
    build-base \
    cmake \
    git \
    icu-dev \
    openssl-dev
                                                                                                                                                                                                                                          
RUN gem install github-linguist
RUN gem install gollum
RUN gem install org-ruby  # optional

ENV GOLLUMARGS=""

WORKDIR /wiki
EXPOSE 4567

CMD gollum $GOLLUMARGS
```
Actual docker command:
```
docker build -t gollum .

docker run --restart=unless-stopped -d -p 127.0.0.1:4567:4567  -e GOLLUMARGS="--bare  --allow-uploads page --base-path myBasePath" -v /tmp/testWiki:/wiki --name testWiki gollum

```

Cmd used to create the git directory
```
git init --bare 
```

dometto:
This PR fixes #1830 and attempts to make the logic for generating the correct upload location for the "per page" upload mode more robust.

We generate the upload location from `request.referer` in the `upload_file` route. #1830 was caused because of a mistake in the regular expression trying to match the `base_url` part in the referer: we were not expecting it to start with a slash.

Investigation also revealed another, similar problem: we were also not handling referers containing `/gollum/foo` at the start of the URL. Again, our regex was expecting a slash at the beginning, whereas that leading slash was already removed earlier in the method.

These two problems were not caught by the tests because we were not testing these two cases:

1. The case in which the referer contains `/gollum/foo`
2. The case of a wiki with base path enabled

Instead, we were only testing referers of the form `localhost:4567/Home.md`.

I have therefore:

* Removed the above described errors from the matching regexes
* Moved the whole logic into a `Precious::Helpers` method to facilitate easier testing
* Added tests for this helper method
* Added a test in `test_app.rb` where we explicitly test uploading with a referer containing `gollum/edit` (the case of a drag and drop upload)

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -241,22 +241,9 @@ module Precious
           tempfile = params[:file][:tempfile]
         end
         halt 500 unless tempfile.is_a? Tempfile
+        
+        dir = wiki.per_page_uploads ? find_per_page_upload_subdir(request.referer, request.host_with_port, wiki.base_path) : 'uploads'
 
-        if wiki.per_page_uploads
-          dir = request.referer.match(/^https?:\/\/#{request.host_with_port}\/(.*)/)[1]          
-          # remove base path if it is set
-          dir.sub!(/^#{wiki.base_path}/, '') if wiki.base_path
-          # remove base_url and gollum/* subpath if necessary
-          dir.sub!(/^\/gollum\/[-\w]+\//, '')
-          # remove file extension
-          dir.sub!(/#{::File.extname(dir)}$/, '')
-          # revert escaped whitespaces
-          dir.gsub!(/%20/, ' ')
-          dir = ::File.join('uploads', dir)
-        else
-          # store all uploads together
-          dir = 'uploads'
-        end
         halt 500 if dir.include?('..')
         halt 500 unless Pathname(dir).relative?
 
--- a/test/test_app_helpers.rb
+++ b/test/test_app_helpers.rb
@@ -1,0 +1,32 @@
+require File.expand_path(File.join(File.dirname(__FILE__), "helper"))
+
+context 'Precious::Helpers' do
+  include Precious::Helpers
+  
+  test 'remove trailing and leading slashes' do
+    ['/wiki', '/wiki/', 'wiki/', '//wiki//'].each do |param|
+      assert_equal 'wiki', remove_leading_and_trailing_slashes(param)
+    end
+    assert_equal 'wi/ki', remove_leading_and_trailing_slashes('/wi/ki/')
+    assert_equal '', remove_leading_and_trailing_slashes('/')
+  end
+  
+  test 'per page upload location helper' do
+    # https referer with and without base path
+    host_with_port = 'localhost:4567'
+    assert_equal 'uploads/Home', find_per_page_upload_subdir('https://localhost:4567/Home.md', host_with_port, nil)
+    assert_equal 'uploads/Home', find_per_page_upload_subdir('https://localhost:4567/wiki/Home.md', host_with_port, '/wiki')
+    
+    # http referer with and without base path
+    assert_equal 'uploads/Home', find_per_page_upload_subdir('http://localhost:4567/Home.md', host_with_port,  nil)
+    assert_equal 'uploads/Home', find_per_page_upload_subdir('http://localhost:4567/wiki/Home.md', host_with_port, '/wiki')
+    
+    # edit page referer with and without base path
+    assert_equal 'uploads/foo/Home', find_per_page_upload_subdir('http://localhost:4567/gollum/edit/foo/Home.md', host_with_port, nil)
+    assert_equal 'uploads/foo/Home', find_per_page_upload_subdir('http://localhost:4567/wiki/gollum/edit/foo/Home.md', host_with_port, '/wiki')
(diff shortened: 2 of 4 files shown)


dometto:
A review of this would be great because of the detailed nature of the logic. We've already introduced bugs into this logic a bit too often. :)

Questions:

1. Are these tests sufficient or should we also add a test testing the upload logic with base path enabled in `test_app.rb`?
2. After releasing `5.3.0`, we've already merged a PR adding support for default keybinding, and a PR adding `yarn` to our build setup. I'm not sure if that still qualifies as a bugfix release. Should I backport this fix and release a `5.3.1`? Or just release a `5.4`?

lib/gollum/app.rb
```diff
@@ -241,22 +241,9 @@ class App < Sinatra::Base
           tempfile = params[:file][:tempfile]
         end
         halt 500 unless tempfile.is_a? Tempfile
+        
+        dir = wiki.per_page_uploads ? find_per_page_upload_subdir(request.referer, request.host_with_port, wiki.base_path) : 'uploads'
 
-        if wiki.per_page_uploads
-          dir = request.referer.match(/^https?:\/\/#{request.host_with_port}\/(.*)/)[1]          
-          # remove base path if it is set
-          dir.sub!(/^#{wiki.base_path}/, '') if wiki.base_path
```
dometto, line 248: The regex for this `sub!` would not match because 1. `base_path` usually starts with a slash and 2. on https://github.com/gollum/gollum/pull/1831/files#diff-65eee83dd06f55c49a91cd71e716ae5e5ef09b353bf7472d4079ffdc175c395aL246 we were selecting the part of the referer after that slash

lib/gollum/app.rb
```diff
@@ -241,22 +241,9 @@ class App < Sinatra::Base
...
-          # remove base_url and gollum/* subpath if necessary
-          dir.sub!(/^\/gollum\/[-\w]+\//, '')
```
dometto, line 250: Here we are expecting `dir` to start with a slash, but again, the above logic selected part of the string after the slash.

test/test_app_helpers.rb
```diff
@@ -0,0 +1,32 @@
+require File.expand_path(File.join(File.dirname(__FILE__), "helper"))
+
+context 'Precious::Helpers' do
+  include Precious::Helpers
+  
+  test 'remove trailing and leading slashes' do
```
dometto, line 6: This test could also be removed, I just used it to make sure I was getting the regexes right :)

benjaminwil:
> Are these tests sufficient or should we also add a test testing the upload logic with base path enabled in test_app.rb?

Is it a lot of effort to add this? I think more test coverage around this feature would be better.

> After releasing 5.3.0, we've already merged a PR adding support for default keybinding, and a PR adding yarn to our build setup. I'm not sure if that still qualifies as a bugfix release. Should I backport this fix and release a 5.3.1? Or just release a 5.4?

After reading the linked issues, it doesn't seem like this issue is a regression. So maybe a 5.4 release is fine? I don't have a strong opinion either way.

test/test_app_helpers.rb
bartkamphorst: I don't mind leaving this in; up to you.

bartkamphorst (approved):
Looks good, thanks for taking up this issue.

bartkamphorst:
> Is it a lot of effort to add this? I think more test coverage around this feature would be better.

I second this.

dometto:
>> Are these tests sufficient or should we also add a test testing the upload logic with base path enabled in test_app.rb?

> Is it a lot of effort to add this? I think more test coverage around this feature would be better.

It was trivial to add this because I could add basically the same test to the "Frontend with basepath" context. Only downside is a bit of code duplication, but agreed that in this case it makes sense to have better coverage!

Thanks for the review to both of you, I'll do a release soon since it's a bug that many people might run into.

PR #1834, opened 2022-06-19
Add support for Mermaid. Resolves #1118.

### LINKED ISSUE #1118: mermaid support

dometto:
This PR adds support for the MermaidJS diagram rendering library. Both GitLab and GitHub now support Mermaid, so this is a feature that would increase compatibility.

### Approach

1. This uses the new [language handler](https://github.com/gollum/gollum-lib/pull/410) functionality from `gollum-lib` to handle codeblocks with the `mermaid` language.. These blocks are simply wrapped in `<div class="mermaid"> </div>`
2. Mermaid is conditionally loaded if the new `--mermaid` flag is active
3. Mermaid is initialized with the `securityLevel: sandbox` parameter, which renders diagrams in iframes. This is to prevent security issues which may otherwise result from mermaid interpreting the code in the codeblock. (With other security settings, Mermaid even allows calling JS from diagrams.)

### Usage:

With `--mermaid` active, just start a three-backtick codeblock with `mermaid` as a language.

```
    ```mermaid
         sequenceDiagram
              participant Alice
              participant John
              links Alice: {"Dashboard": "https://dashboard.contoso.com/alice", "Wiki": "https://wiki.contoso.com/alice"}
              links John: {"Dashboard": "https://dashboard.contoso.com/john", "Wiki": "https://wiki.contoso.com/john"}
              Alice->>John: Hello John, how are you?
             John-->>Alice: Great!
             Alice-)John: See you later!
    ```
```

(Apologies for the strange indentation, this is because GitHub now supports ```mermaid blocks as well :))

### Result

Something like this:

<img width="1088" alt="Screenshot 2022-06-19 at 12 20 06" src="https://user-images.githubusercontent.com/147111/174479679-e25ffc5c-f698-4286-9d8e-32d4677c3eac.png">

### Issues and questions

1. Both GitLab and GitHub render diagrams in iframes for security reasons, but it appears they use their own implementation for this, as Mermaid's `securityLevel: 'sandbox'` is still [a relatively new feature](https://github.com/mermaid-js/mermaid#security-and-safe-diagrams). Still, for gollum, I think we should just go for the inbuilt feature. The question is whether we consider this safe enough to activate by default, or whether we want to enable Mermaid only when the appropriate flag is set (as I've done in this PR for now)?

2. I've discovered [an annoying bug in the sandbox functionality](https://github.com/mermaid-js/mermaid/issues/3153). Not sure if this is a showstopper.

3. I've not considered things such as theming and user configurability yet, but think this should be achievable from `config.js`. For MathJax we have a separate config file, but I am wondering whether we should not fold all of that into `config.js` anyway.

NB: this PR does not yet contain compiled static assets.


lib/gollum/assets.rb
```diff
@@ -16,7 +17,7 @@ def self.sprockets(dir = File.dirname(File.expand_path(__FILE__)))
       env.append_path ::File.join(dir, 'public/gollum/images')
       env.append_path ::File.join(dir, 'public/gollum/fonts')
 
-      env.js_compressor  = :uglify unless Precious::App.development?
+      env.js_compressor  = ::Uglifier.new(harmony: true) unless Precious::App.development?
```
dometto, line 20: Ran into https://github.com/lautis/uglifier/issues/127 when trying to add mermaid as an asset

lib/gollum/public/gollum/javascript/mermaid.js
```diff
@@ -0,0 +1,2 @@
+// Load mermaid from installed Node module
```
dometto, line 1: Created this separate JS file instead of adding to `app.js` so we can require the mermaid library conditionally (so we don't load it when mermaid isn't active anyway)

benjaminwil (approved):

lib/gollum/app.rb
```diff
@@ -31,6 +31,8 @@
 Gollum::set_git_timeout(120)
 Gollum::set_git_max_filesize(190 * 10**6)
 
+Gollum::Filter::Code.language_handlers[/mermaid/] = Proc.new { |lang, code| "<div class=\"mermaid\">\n#{code}\n</div>" }
```
bartkamphorst, line 34: Very cool, this.

bartkamphorst (approved):

dometto:
I'm sadly not getting any response on https://github.com/mermaid-js/mermaid/issues/3153 . @benjaminwil if you have a moment could you let me know if I'm overlooking anything obvious that could prevent this issue? Otherwise I'll try asking around on StackOverflow.

benjaminwil:
I read through the documentation and don't think there's anything obvious you're overlooking. The docs do say that the "sandbox" security level is a beta feature. But that's all I could find.

Maybe using the "strict" security level is reasonable enough for now. We can sanitize the rest of the document the same way we do for Markdown files perhaps?

dometto:
Good suggestion @benjaminwil! I've set the default security level to strict, and made this (and other mermaid settings) overrideable via `custom.js`. The user can just modify the `mermaid_conf` variable. Obviously this will need to be documented. Please let me know what you think of this approach. If it makes sense, we could think about implementing this way of overriding settings for mathjax as well (instead of requiring the user to commit a separate `mathjax.conf.js` file).

Further sanitizing the original mermaid code in the document is tricky since this risks breaking mermaid syntax, so I don't know what else we could do to improve security. An alternative would be to go with sandbox as a default, and make this overrideable via `custom.js`. What do you think is acceptable, @benjaminwil @bartkamphorst?

dometto:
Running into some new JS compression issues with the latest version of mermaid. Basically, it looks like we should stop using `uglifier` and move to `terser`: https://github.com/lautis/uglifier/issues/174#issuecomment-1054463562

Annoyingly, `terser` won't work with the standard JRuby JS runtime, `therubyrhino`. An alternative is available, although it doesn't seem widely used: [dienashorner](https://github.com/kares/dienashorner). But I can only get this to work when requiring it in `Gemfile`, so when running `gollum` using `bundle exec`.

However, given #1847, I am wondering whether the thing to do isn't just to give up on precompilation support on JRuby, and treat this only as a development feature. Using static assets, everything is fine on JRuby. Thoughts welcome!

benjaminwil:
> Further sanitizing the original mermaid code in the document is tricky since this risks breaking mermaid syntax

@dometto Maybe I'm misunderstanding something, but won't whatever sanitizes the parent document (the Markdown file) inadvertently sanitize the Mermaid diagram anyway? It looks unlikely to me, though, that that would be a real issue.

I think making the security level configurable makes sense. But I might disagree with defaulting to "sandbox" until we have resolved the weird double-iframe issue.

benjaminwil:
> However, given #1847, I am wondering whether the thing to do isn't just to give up on precompilation support on JRuby, and treat this only as a development feature. Using static assets, everything is fine on JRuby. Thoughts welcome!

I am in favour of treating `--no-static` as a development feature. I think it is reasonable to ship static JavaScript and CSS as this is satisfactory for most users.

We already provide another way for users to add additional JavaScript and CSS at runtime (`--assets`, `--js`, `--css`), so users can continue to add additional assets should they please.

If we find that users are unhappy being able to use `--no-static` for some other reason--it would be good to find out what that reason is, as maybe there is another way to solve their problem that's not `--no-static`.

dometto:
> won't whatever sanitizes the parent document (the Markdown file) inadvertently sanitize the Mermaid diagram anyway? 

That's correct! I was thinking that still, we cannot exclude there being some strings that a) won't be sanitized on our side by loofah b) will be rendered as executable script somehow by Mermaid internally. I think that's more or less why the `sandbox` feature exist. If that turns out to be possible, of course that's an exploit that should be fixed by Mermaid ASAP, and we can fix it by updating. So I agree the risk is probably acceptable, since `--mermaid` will be false by default.

lib/gollum/templates/layout.mustache
```diff
@@ -56,8 +58,26 @@
...
+  };
+  </script>
+  {{/mermaid}}
+  
   {{#js}}<script type="text/javascript" src="{{custom_js}}"></script>{{/js}}
 
+  {{#mermaid}}
+  {{#sprockets_javascript_tag}}mermaid{{/sprockets_javascript_tag}}
+  <script>
+    console.log(mermaid_conf)
```
bartkamphorst, line 76: leftover debug statement?

dometto: Yes, I left this in because I was still trying to get the tests to succeed. :) Thanks for marking it!

PR #1836, opened 2022-07-17
Refactor RSS view renderer

### LINKED ISSUE #1815: Only first line of commit message should be in RSS feed item title, the rest should go to description
**Describe the bug**

Perhaps, not a bug but a feature request. The RSS feed seems to get an entire commit message, and it is put into RSS item `<title>` tag, while the `<description>` tag only gets meta information such as `Commited by`, `Commit ID` and `Affected files`.

This works okay while editing via web-interface, because the changes description input field only has one line, but if one works with Gollum repository on his local machine, there he can compose multiline commit messages, sometimes very long and detailed ones, and all that content will go to the `<title>`, messing up the looks of the feed in one's RSS client.

**To reproduce**

1. Commit some changes with a multiline commit message, for example:
```
Some changes

- one thing
- another thing
```
2. Push it to remote (*or wherever Gollum serves the repository contents from*).

**Expected behavior**

RSS feed item gets only first line (*perhaps with an empty line as separator*) of the commit message to its `<title>` tag, while the rest goes to `<description>` tag.

**Environment Info**

```
$ gollum --version
Gollum 5.2.3

$ git --version
git version 2.25.1

$ lsb_release -a
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
```

benjaminwil:
Resolves #1815

In issue https://github.com/gollum/gollum/issues/1815, it was reported that Gollum RSS feeds attempt to put entire commit messages into a feed item's `<title>` element.

Given that commit messages can be many paragraphs long, this is not an acceptable way to render a feed item title.

If a commit has many lines or paragraphs, we now put those inside the `<description>` element as of this commit.

While I was editing this view, I decided it would be a good time to increase test coverage, as this view renderer was not under test at all. I also fixed a typo ("Commited" should say "Committed"). There are implicit dependencies on other gems that provide the
`Gollum::Git::Commit`, `Gollum::Git::Actor`, and `Rugged::Commit` classes here, so mocking out simple versions of their interfaces seemed like the path of least resistance to setting up controllable tests.

I also removed some now-duplicated assertions from `test/test_app.rb`.

### DIFF
--- a/lib/gollum/views/rss.rb
+++ b/lib/gollum/views/rss.rb
@@ -1,43 +1,84 @@
 require 'rss'
 
 class RSSView
-  
   include Precious::Views::AppHelpers
   include Precious::Views::RouteHelpers
-  
+
   attr_reader :base_url
-  
+
   def initialize(base_url, wiki_title, url, changes)
     @base_url = base_url
     @wiki_title = wiki_title
     @url = url
     @changes = changes
   end
-  
+
   def render
-    latest_changes = "#{@url}#{latest_changes_path}"
     RSS::Maker.make('2.0') do |maker|
       maker.channel.author = 'Gollum Wiki'
-      maker.channel.updated = @changes.first.authored_date
-      maker.channel.title = "#{@wiki_title} Latest Changes"
       maker.channel.description = "Latest Changes in #{@wiki_title}"
-      maker.channel.link = latest_changes
+      maker.channel.link = latest_changes_url
+      maker.channel.title = "#{@wiki_title} Latest Changes"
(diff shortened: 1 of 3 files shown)


dometto (approved):
Thanks! Besides solving #1815, this is a nice refactor.

(I wasn't sure whether github was linking this PR with #1815 for autoclose purposes, so I explicitly added "Resolves" in the message body, but maybe this was superfluous -- I don't see any visual feedback that autoclose should work now.)

PR #1849, opened 2022-08-21
Update sprockets, move to sassc-embedded

dometto:
See [#1435](https://github.com/gollum/gollum/issues/1435#issuecomment-1221537121)


benjaminwil (approved):

dometto:
We could also choose to use `sassc-embedded` on both MRI and JRuby for consistency?

bartkamphorst:
> We could also choose to use `sassc-embedded` on both MRI and JRuby for consistency?

Another benefit of this approach is that we would be using the primary implementation of Sass.

dometto:
Using `sassc-embedded-shim-ruby` (that is: with dart sass) I was able to move the `primer` dependency to `yarn`. With `sassc` this fails due to differences between `sassc` and `dart-sass`. I guess that's an argument for moving to `sassc-embedded-shim-ruby` entirely.

I am a bit unsure about the following: https://github.com/ntkme/sassc-embedded-shim-ruby#behavioral-differences-from-sassc-ruby

I don't know much about SASS, but the ':style => :compress` option mentioned there sounds like minification support? However, when I rune `bundle exec rake precompile` on this branch, the resulting application CSS looks minified just fine to me. @benjaminwil can you weigh in on whether the mentioned behavioral differences provide any reason against migrating to `sassc-embedded-shim-ruby`?

benjaminwil:
@dometto I'm not sure I know much more about SASS than you do, but reading that list of behavioural differences doesn't phase me for the (small) amount of non-library CSS bundled with Gollum. It's possible that the precompiled CSS will have trivial differences re: line comments, but I don't think we will be breaking any stylesheets.

Looks like [`:compressed` does the right thing](https://github.com/ntkme/sassc-embedded-shim-ruby/blob/v1.54.0/test/output_style_test.rb#L103-L108) as far as our needs for providing production assets.

lib/gollum/app.rb
```diff
@@ -25,6 +24,10 @@
 require 'gollum/views/rss.rb'
 require 'gollum/views/template_cascade'
 
+['sassc', 'sassc-embedded'].each do |gem|
+  require gem if Gem::Specification.find {|spec| spec.name == gem}
```
dometto, line 28: If not running via `bundle exec`, these gems may not be available unless installed manually, so only require them conditionally.

PR #1873, opened 2022-11-13
Ensure search terms in URLs are escaped

benjaminwil:
Resolves #1873. 

An issue was reported (in discussion #1872) that search terms with special characters (specifically `#`) were resulting in URLs that did not lead to any search results where there should be search results.

I was able to reproduce this issue in a feature test and fix it in the view class.

### DIFF
--- a/lib/gollum/templates/search.mustache
+++ b/lib/gollum/templates/search.mustache
@@ -18,7 +18,7 @@
   <ul>
   <div class="Box-header border-bottom p-0"></div>
   {{#results}}
-    <li class="Box-row Box-row--gray">
+    <li class="Box-row Box-row--gray search-result">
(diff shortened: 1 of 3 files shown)


benjaminwil:
Seemingly unrelated tests are failing. Once I've resolved that I think this is safe to merge.

dometto:
Fix and spec look good, thanks for picking this up so quickly!

> Seemingly unrelated tests are failing. Once I've resolved that I think this is safe to merge.

How annoying that the integration tests are failing again.. Could be a different version of Selenium or the webdriver, which is behavinig differently than our tests locally.

dometto:
Think this may need a rebase to get rid of the `mathjax.config.js` failures.

benjaminwil:
All is well 👍

PR #1877, opened 2022-11-14
Don't double escape url for the last modified info link

### LINKED ISSUE #1851: Double URL escape in "last modified" link
1. Create a page with a space in the name, e.g., `foo bar.md`
2. View the page and click the "When was this page last modified?" link.

It makes a request to

```
http://localhost:4567/gollum/last_commit_info?path=foo%2520bar.md
                                                      ^^^^^ double escaped
```

The API does not return the expected response because it tries to access a file `foo%20bar.md` instead of `foo bar.md`.

Gollum 5.3.0, Ruby 2.7.2.

dometto:
Resolves #1851

Since we're passing the path to our ajax route as `data` using JQuery, it is already auto-encoded. We hence don't need to use `escaped_url_path`, but can simply use `path`. (Note: since we're using mustache's `{{ }}` to print `{{path}}`, this is aready html-escaped.)

### DIFF
--- a/lib/gollum/templates/wiki_content.mustache
+++ b/lib/gollum/templates/wiki_content.mustache
@@ -47,7 +47,7 @@
 	<div id="footer" class="pt-4">
 	  {{^historical}}
 		{{^preview}}
-		  <p id="last-edit"><div class="dotted-spinner hidden"></div> <a id="page-info-toggle" data-pagepath="{{escaped_url_path}}">When was this page last modified?</a></p>
+		  <p id="last-edit"><div class="dotted-spinner hidden"></div> <a id="page-info-toggle" data-pagepath="{{path}}">When was this page last modified?</a></p>
 		  {{#allow_editing}}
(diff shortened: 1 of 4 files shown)


dometto:
@benjaminwil this PR should also fix the latest integration test errors related to the JavaScript error logs (it looks like a new error was being thrown about the absent`matjax.config.js`, which I've added to the expecations).

dometto:
Capybara test vs the Docker build currently failing because of the way I set up the new test: it uses`@wiki.write_page`, but this has no effect on the wiki that that is used by the Dockerized app. We could either:

1. Separate the current integration tests and have a special suite of tests (acceptance?) to run against the Docker build.
2. Try to dynamically set `@wiki` in the tests to the repository that the Dockerized app is using and modify that.

2 seems slightly hackish but may be preferable, since it would allow us to run all our standard tests against the Docker build, thus hopefully catching more potential issues with the latter.

PR #1878, opened 2022-11-15
Update MathJax by adding it to Yarn

dometto:
This PR:

* Removes the MathJax source from our repo
* Adds the latest MathJax as a yarn dependency
* Updates the default MathJax configuration
  * I have also enabled safe mode, which so far was turned off by default

This is not as clean as I'd like it to be. Since MathJax loads modules on the fly, I had to add three files as separate sprockets manifests:

* `ui/safe.js`
* `svg/svg.js`
* `svg/tex.js`

...and tell MathJax (in the `layout.mustache` template) where to find these using our sprockets helper tags. This means that potentially, we'll have to add more such files if MathJax changes in the future.

I also still had to manually add the MathJax webfont to the repo.

As a result of loading MathJax via yarn/sprockets, I also had to switch to requiring the full set of TeX extensions at load time (currently I think these are loaded on the fly). This may have an adverse impact on page load times (but I have not measured this). Given all this, I wonder if switching to yarn is worth it for this dependency?

I also ran into an Uglifier error when precompiling, forcing me to switch to Terser for JS compression. But that seems to work well.

(Todo: make sure the webfonts are not removed when precompiling assets.)


bartkamphorst:
Honestly, this feels like it's not going to be a better solution than updating and bundling mathjax manually.

dometto:
Agreed. Will close this and open a new PR that simply updates MathJax without bundling it with yarn.

PR #1887, opened 2022-11-27
Render Mermaid in Preview

polgarc:
Fix Mermaid block rendering in the Preview tab.

Before;
image

After:
image

Note the Mermaid documentation says that [calling `camermaid.init` is deprecated](https://github.com/mermaid-js/mermaid/blob/develop/docs/config/usage.md#calling-mermaidinit):
> This type of integration is deprecated. Instead the preferred way of handling more complex integration is to use the mermaidAPI instead

However mermaidAPI seems to only have the [render](https://mermaid-js.github.io/mermaid/#/Setup?id=render) function, which requires an id. This means we'd have to look up all `.mermaid` divs, generate ids, then call `render` in a loop. However this is exactly what [mermaid.init()](https://github.com/mermaid-js/mermaid/blob/3b79ee35e0382edfec78a617c77d1c578bfc5fd1/packages/mermaid/src/mermaid.ts#L136) does, so I thought it's just much simpler to call `mermaid.init()`.


dometto:
Thanks for the report and fix! The Mermaid docs are very unclear, but I think `mermaid.init` itself is not deprecated, just the use of it for configuring/adjusting the mermaid configuration. Since we've already configured mermaid and just want to retrigger the rendering of diagrams, this should be no problem for us, I think.

@polgarc do you think you could add a kind of Capybara test to make sure no JavaScript errors occur on the preview page, like [here](https://github.com/gollum/gollum/blob/master/test/integration/test_js_errors.rb#L36)?

dometto:
Will add the test myself. Thanks @polgarc!

PR #1897, opened 2022-11-29
Bundle ace with yarn

dometto:
* Removes ACE source from the repo
* Adds latest version of ACE via `yarn`
* Adds capybara test ensuring there are no JS errors when switching editor modes and keybinding


benjaminwil (approved):
🎊 Amazing. 518 files! Wow.

dometto:
Yes I was also very surprised by how easy this proved to be! Thanks for setting up the `yarn` workflow!

Issue #1880, opened 2022-11-20
ACE editor breaks on non-ascii characters


vfrc2:
**Describe the bug**
ACE editor breaks on non-ascii characters. On input it misaligned rendered characters and cursor.
It may happens due to two different fonts used in ace editor:  
```css
#gollum-editor-body-ace {
    overflow: hidden;
    font-family: Consolas, "Liberation Mono", Courier, monospace;
    font-size: 1em;
    padding: 0;
}
...
.ace_editor {
    position: relative;
    overflow: hidden;
    font: 12px/normal 'Monaco', 'Menlo', 'Ubuntu Mono', 'Consolas', 'source-code-pro', monospace;
    direction: ltr;
    text-align: left;
    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
}
```

I fix this by providing custom css via `--css` option:
```css
#gollum-editor-body-ace {
    font-family: Monaco, Menlo, "Ubuntu Mono", Consolas, source-code-pro, monospace !important;
}
```

**To Reproduce**
1. Edit or create page
2. Start type 20+ non-asci characters

**Expected behavior**
Expect to input character renders on right positions

**Screenshots**

**Environment Info**
Run on official docker image.  

bash-5.1# gollum --versions
Gollum 5.3.0
Running on: aarch64-linux-musl with Ruby version 2.7.6
Using:
rugged 1.1.1
gollum-rugged_adapter 1.1.2
gollum-lib 5.1.3
Markdown rendering gem: kramdown
Other renderering gems:
RedCloth 4.3.2
org-ruby 0.9.12
creole 0.5.0
asciidoctor 2.0.17
wikicloth 0.8.3

dometto:
Not able to reproduce this:

<img width="1097" alt="Screenshot 2022-11-27 at 19 33 50" src="https://user-images.githubusercontent.com/147111/204153294-f110c238-9273-45b4-8364-fbce51d767b8.png">

Is it really any non-ascii string that breaks for you @vfrc2? Or is this browser dependent? I'm on Firefox on OS X.

vfrc2:
I'm think so... I tried only on chrome. I found it strange, what fonts are different.
I will try on another browser, and comment back.

> 
I'm think so... I tried only on chrome. I found it strange, what fonts are different.
I will try on another browser, and comment back.

vfrc2:
[...]

PR #1898, opened 2022-11-30
Fix/bad cursor fix

vfrc2:
Fix issue #1880
~~Move editor font to Ace config~~
Update public assets

Consider add node_modules to git ignore
```
diff --git a/.gitignore b/.gitignore
index 92a0a600..8cb7b0f2 100644
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Gemfile.lock
 !.sprockets*
 !lib/gollum/public/gollum/stylesheets/_styles.css
 !.github*
+/node_modules
\ No newline at end of file
```


lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -126,6 +126,10 @@
       }
         
       editor.setTheme("ace/theme/tomorrow");
+      editor.setOptions({
+        fontFamaly: 'Consolas, "Liberation Mono", Courier, monospace',
```
dometto, line 130: Typo: `fontFamaly`

vfrc2: you right, but it break fix... ))) need to dig more

dometto: Thanks for looking in to it!

vfrc2:
Add Monaco font to font-family, it solves the problem. Seems it caused by bad default font in Chrome on MacOS.

dometto:
> Consider add node_modules to git ignore

Good idea, will do this in #1903!

Thanks for the fix, looks good to me!

PR #1900, opened 2022-12-04
Katex support. Resolves #1404

### LINKED ISSUE #1404: KaTeX for Mathematical Typesetting
Currently mathematical typsetting is provided by mathjax, which potentially has a small bug due to the discontinuation of the CDN #1390.

However, for a few reasons, mathjax is possibly less preferred than KaTeX for typesetting. In particular, the KaTeX project has more github stars and contributors than mathjax. Further, KaTeX demonstrates that they are faster than mathjax.

Currently gollum's mathematical typsetting is hardcoded to be provided by mathjax, e.g. with the flag `--mathjax`. 

Is it possible to have `5.x` also support KaTeX provided typesetting?

dometto:
- [x] Change `--math` option to allow MathJax (default) or KaTeX
- [x] Make KaTeX setup configurable
- [x] Decide what to do about `math.conf`, `config.js` `mathjax.conf` (see discussion)
- [x] Render KaTeX in Preview tab
- [x] Update docs


Rakefile
```diff
@@ -63,6 +65,15 @@ def replace_header(head, header_name)
   head.sub!(/(\.#{header_name}\s*= ').*'/) { "#{$1}#{send(header_name)}'"}
 end
 
+def bundle_katex_fonts(assets_path)
```
dometto, line 68: I wasn't able to succesfully the fonts using sprockets: adding `*.woff2` to the sprockets manifest would result in the fonts being saved to `lib/gollum/public/assets/katex/dist/fonts` as expected, but since they would be managed the would have filesnames of the form: `KaTeX_Fontname-SPROCKETSHASH.woff2`. However, the KaTeX CSS is unware of our sprockets setup and would still look for the 'normal' filename `KaTeX_Fontname.woff2`, which did not exist.

The solution here is to just copy the fonts manually at the end of the precompilation task, thus preserving the original filenames.

An alternative could be to override the font definitions in our own style files, but that would also be a bit of a hack, and possible a more difficult one to maintain.

dometto: Also note that I'm only copying the `.woff2` font files to save space. Given [woff2 support](https://caniuse.com/woff2) these days I think this should be ok?

dometto:
With regard to allowing users to customize the `KaTeX` config: currently for `MathJax` we are allowing users to load `mathjax.conf.js` to override the default config. I see two options:

1. Rename `mathjax.conf.js` to `math.conf.js` and allow users to add custom config for either MathJax or KaTeX (depending on what they're usiing) there.
1. Or more radically: remove support for `mathjax.conf.js` and instead allow users to customize MathJax and KaTeX from inside of `custom.js`.

The 2nd option would require users to run gollum with the `--js` flag in addition to `--math`, but would simplify the setup in our `layout` template quite a bit, and would stop the occurrence of 404's when looking for `mathjax.conf.js`, when the user just uses the default config.

dometto:
> Or more radically: remove support for mathjax.conf.js and instead allow users to customize MathJax and KaTeX from inside of custom.js.

I have gone with this option for now. Does this make sense for a 6.0 release or is it too radical/unnecessary?

> Change --math option to allow MathJax (default) or KaTeX

I have implemented this, but alternative we could also just have two switches: `--mathjax` and `--katex` (if the use supplies both then the app will go with the last one given). Which is preferable?

dometto:
Also pinging @ProgramFan for feedback, since I know you have much experience with math in gollum. If you have any suggestions about how math and katex should be implemented from the next release on, please let us know.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -440,10 +440,16 @@ $(document).ready(function() {
             sectionHeading.scrollIntoView();
           }
         }
+        previewContent = $('.tabnav-div#preview-content')[0];
         if (window.MathJax != null) {
-          MathJax.startup.defaultReady();
```
dometto, line 444: Caught a bug with this implementation of the mathjax rendering in the preview page: if you switched to the preview pane a second time math wouldn't render!

I copied this implementation from the [mathjax demo](https://www.mathjax.org/#demo) and the way it refreshes constantly.

Rakefile
benjaminwil: > Also note that I'm only copying the `.woff2` font files to save space. Given [woff2 support](https://caniuse.com/woff2) these days I think this should be ok?

Yes, at work I only need to support browsers that support `woff2` these days. If the corporate overlords are okay with that, maybe we can be too!

benjaminwil: Regarding fingerprinted/digested assets: that is too bad about the fonts. But fonts don't change that often, so maybe your solution is okay long term.

If we run into this a lot, we may be able to find or create a Sprockets extension that gives us the ability to un-fingerprint certain types of assets—like fonts.

benjaminwil:
I like all of the decisions you've made so far.

I am not opposed to having two switches, `--mathjax` and `--katex` but would also be happy with a single `--math` switch that implicitly uses the `--js` switch as well.

Rakefile
```diff
@@ -3,6 +3,8 @@ require 'rake'
 require 'date'
 require 'tempfile'
 
+NODE_MODULES = File.join(File.dirname(__FILE__), 'node_modules')
```
bartkamphorst, line 6: `::File`.

Rakefile
bartkamphorst: > The solution here is to just copy the fonts manually at the end of the precompilation task, thus preserving the original filenames.

Could we rename them instead of copying over them? Seems computationally cheaper, though it may not matter that much. Just a thought.

bin/gollum
```diff
@@ -121,10 +121,13 @@ MSG
     wiki_options[:allow_uploads]    = true
     wiki_options[:per_page_uploads] = true if mode == :page
   end
-  opts.on('--mathjax', 'Enable MathJax (renders mathematical equations).',
-    'By default, uses the \'TeX-AMS-MML_HTMLorMML\' config with the \'autoload-all\' extension.') do
-    wiki_options[:mathjax] = true
-    wiki_options[:mathjax_config] = 'mathjax.config.js'
+  opts.on('--math [RENDERER]', [:mathjax, :katex], 'Enable rendering of mathematical equations.',
+    'Valid renderers: mathjax, katex. Default: mathjax.') do |mode|
```
bartkamphorst, line 125: Have we discussed enabling `katex` as the default option? Given that it is easier to bundle and to update, this may be preferable for the majority of users?

benjaminwil: That sounds great. We just may want to make it very explicit that MathJax and KaTeX are not fully compatible with each other (to my limited knowledge, at least). So if someone is already using MathJax they may want to continue using MathJax.

bartkamphorst: Yes, especially since Github and Gitlab also diverge in this respect. So for the time being I think we want to offer both. But given how `--mathjax` and the corresponding config setting will remain functional for `6.x`, I think those existing users shouldn't be bothered if we default the new `--math` flag to `katex`, right?

benjaminwil: Right!

Rakefile
dometto: > If we run into this a lot, we may be able to find or create a Sprockets extension that gives us the ability to un-fingerprint certain types of assets—like fonts.

Good idea to keep in the back of our minds if we're not satisfied with this in the long run!

> Could we rename them instead of copying over them?

We could but then we'd be breaking the source installation of the library inside `node_modules`. If you then run gollum with `--development-assets` (which serves the assets from `node_modules`) it would break. :)

Rakefile
dometto: Thanks!

bin/gollum
dometto: Great idea! An even more radical solution would be to activate `katex` out of the box, so without the need for a switch, and then let users override it with `--mathjax`? I'm not sure if there are downsides to enabling math support out of the box (potentially page load times, I suppose).

dometto:
> would also be happy with a single --math switch that implicitly uses the --js switch as well.

That's a very nice idea! Here's a little worry, not sure how important it is: I just discovered by accident that `--math` already functions in `gollum` as a way of activating `--mathjax`! I think this is probably a feature (or a bug?) of the `OptParser` library, because we don't define `--math` ourselves.

Now, this is not a big deal in itself, but: it means there might be people using the `--math` switch. And *if* we now make `--math` enable `--js` implicitly, is there a possible security concern? I mean: someone may have a `config.js` committed to their repo, but not be using `--js`. Then after upgrading gollum, `--js` would be sneakily enabled by the `--math` switch they're already using.

Just let me know if this is too paranoid. 😆 Maybe the solution is just to clearly mark in the release notes that `--math` now enables `--js`.

benjaminwil:
> Just let me know if this is too paranoid. 😆 Maybe the solution is just to clearly mark in the release notes that `--math` now enables `--js`.

That sounds reasonable to me!

If this is all in a major release bump, too, I think it's more acceptable to end users if things break unexpectedly—as long as there is information in the changelog to help them with the upgrade.

dometto:
This is now ready for review. In the end, I decided for the following, for simplicity's sake

* `mathjax.config.js` is renamed to the generic `math.config.js`
* `--math` enables support for `math.config.js` both when using katex, and when using mathjax
* When using `--mathjax`, a message is printing exhorting the user to rename the file to `math.config.js`
* Use of `--mathjax` is discouraged (it prints a deprecation warning, use of `--math mathjax` is encouraged)

dometto:
Also, I have removed the automatic addition of the `--mathjax` flag in `docker-run.sh`, which we had already deprecated.

lib/gollum/templates/layout.mustache
```diff
@@ -90,7 +102,23 @@
...
+  <script type="text/javascript" src="{{math_config}}"></script>
+  {{/math_config}}
+
+  {{#mathjax}}
+  <script async src="{{mathjax_js}}"></script>
+  {{/mathjax}}
+  {{#katex}}
+  <script type="text/javascript">
+    document.addEventListener("DOMContentLoaded", function() {
+        renderMathInElement(document.body, katex_conf);
```
dometto, line 118: We could also try passing in the wiki content div instead of `document.body`.

bartkamphorst: Makes sense to me, as all markup language (and katex blocks) should be contained in `wiki-content`.

dometto:
Rebased on `master` and updated the documentation here:

After releasing 6.0, we should update the section on mathematics support in the wiki's [main page](https://github.com/gollum/gollum/wiki#mathematics).

It's been a while, but I think this is ready to merge!

dometto:
Argh, I've accidentally merged unsquashed the commits from this PR into `master`. I was planning to squash-and-merge, but after checking something in the code one more time the button state had jumped back to rebase-and-merge. 😢 I can clean up `master` of course, but that would require a force push. Maybe best to leave things like this?

dometto:
@bartkamphorst @benjaminwil let me know if you prefer me to squash the commits on `master` and force-push, otherwise I'll leave it like this. Apologies again!

bartkamphorst:
I actually think it's fine if you clean up the commit history of `master` and force push in this instance. It's not like other people are actively committing to `master` at the moment, so it should be fine. You can use `--force-with-lease` [flag](https://salferrarello.com/never-git-push-force/) to be on the safe side. @benjaminwil ?

dometto:
I'm following @bartkamphorst's suggestion to cleanup `master`.

PR #1901, opened 2022-12-06
Make “copy to clipboard” button Permissions API-aware

### LINKED ISSUE #1885: Add Capybara tests for codeblock "copy to clipboard" feature
We recently merged #1754, which provides a "copy to clipboard" button for code blocks when viewing a page. We would like to add tests to ensure this feature continues working. I volunteered to do this. I've just created this issue to hold me accountable to doing the work.

benjaminwil:
Fixes #1885.

This pull request makes the "copy to clipboard" button on wiki pages usable with modern browsers that implement the Permissions API for the "clipboard-write" permission as it exists today. And it generally improves the "copy to clipboard" functionality of code blocks on wiki pages:

- ~Users who don't use pointing devices can use the "copy to clipboard" using the keyboard only.~
  Note: Before this pull request, the feature was not keyboard-accessible, and it still isn't after this pull request. There is discussion about this in the pull request comments.
- The button only uses Primer-provided CSS.


benjaminwil:
CI is exiting before the tests run... I am not sure why yet. Will check it out later this week.

dometto:
Looks like the `psych` gem is failing to install: `yaml.h not found`. A new version of psych (`5.0`) has just been released, so that's probably causing it.

benjaminwil:
Ah, so locally I had to ensure `libyaml` is installed. I can try to add a commit to #1903 later today to remedy this.

dometto:
Sounds good!

benjaminwil:
I believe Selenium is failing because we are not accepting the clipboard permissions prompt that appears in Chromium-based browsers.

This failure is helpful, as it has made me realize that we need to handle the case where the the user has not provided clipboard permissions to Gollum from their browser.

I am a little bit confused, though, why my new test case passes and only the JS errors Capybara test fails. So I must still be missing something.

test/integration/test_page.rb
```diff
@@ -0,0 +1,101 @@
...
+    Capybara.use_default_driver
+  end
+
+  test "'copy to clipboard' functionality on code blocks" do
+    create_page title: "Page with code block", content: <<~TEXT
+      This page includes a code block. In this test, we will use the provided
+      'copy to clipboard' button to ensure that it works properly.
+
+      ```
+      a juicy fish,
```
dometto, line 30: Excellent example :)

dometto:
> I am a little bit confused, though, why my new test case passes and only the JS errors Capybara test fails. So I must still be missing something.

That is quite odd. 😕

Maybe [this](https://github.com/DevExpress/testcafe/issues/2668#issuecomment-412901579) is helpful to address the permissions problem?

benjaminwil:
Just making a note that I'll be expanding the scope of this PR to include a change we talked about in #1937:

> I really would favor a solution though that uses Primer utilities, rather than styling this button ourselves. 
>
> [Here](https://primer.style/css/utilities/marketing-layout) it is explained that the marketing layout utilities are built on top of the [Primer core utilities](https://primer.style/css/utilities/layout#position) so presumably we could position the button using only core utility classes?

bartkamphorst:
@benjaminwil What's the current status of this PR?

benjaminwil:
- It needs to be rebased against the main branch.
- The `wip` commit needs to be dropped or my issue with it needs to be resolved.
- My issue in the `wip` commit is that I cannot figure out how to drive Chrome with the [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API)  enabled or disabled.

I was actually hoping to pick this up again and get it ready for review this weekend! So very timely of you to comment on it now.

bartkamphorst:
That's great. Thanks for the update!

benjaminwil:
@bartkamphorst @dometto This is finally ready to go. 😌

bartkamphorst (changes requested):
The code looks clean, but on Firefox 115.0.2 I'm getting the following error when I click on the copy to clipboard button:
`TypeError: 'clipboard-write' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.` Any ideas?

**EDIT:** The issues persists in 116.0.1. The content is copied, however.

Also I think the fade-in time is a little bit long for my taste, makes the UI feel a bit sluggish.

benjaminwil:
@bartkamphorst Thanks for catching that error in Firefox.

This opened a small can of worms for me, and I'm just adding this comment to explain that can of worms so (hopefully) no one has to waste time digging through the MDN, Firefox bug reports, and ClipboardJS GitHub issues the next time we touch this feature.

---

It turns out that Firefox's support for the clipboard via the Permissions API is nonexistent, but we can still use `Clipboard.writeText()` as long as the user hasn't explicitly turned off clipboard access in their Firefox browser settings. For those users (like me, actually!), I don't know how to offer a good experience yet.

I also tried transitioning to use the [ClipboardJS](https://clipboardjs.com/), but I ran into [a different issue](https://github.com/zenorocha/clipboard.js/issues/407) that stems from ClipboardJS using `document.execCommand()` in Firefox, which [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/document/execCommand) anyway, to copy the text.

Since `Clipboard.writeText()` does actually work in Firefox, we can just return early from the permission checking function if we're confident the current browser is Firefox. This is definitely an unpleasant hack, but I don't see a clean way around it.

benjaminwil:
@bartkamphorst 

> Also I think the fade-in time is a little bit long for my taste, makes the UI feel a bit sluggish.

I agree. Just fixed up the `Style 'copy to clipboard' button with Primer` commit to use different animation classes and removed the delayed hiding of the button.

benjaminwil:
Will also add that I manually tested this in Chrome, Firefox, and Safari without issue.

bartkamphorst (commented):
Thanks @benjaminwil ! Sorry this was such a pain to resolve. I do wonder if we could do without an explicit check for `userAgent == Firefox` but see my remark about that in the review. 

Could you explain how the button is now navigable with the keyboard? When I "tab through" the items, I go from the `Search` input, to `Upload` to `Rename` and then to some image in the page, but never to  `Overview`, `History`, `Edit` or `New` or to any of the `copy to clipboard` buttons.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -143,23 +143,51 @@ function flashNotice(type, notice, button_label, button_function, button_type) {
...
+  // `Clipboard.writeText()` we can return early for browsers that report as
+  // Firefox.
+  if (navigator.userAgent.indexOf("Firefox") > 0) {
+    return false;
+  }
+
+  if (navigator.permissions && navigator.permissions.query) {
+    return await navigator.permissions.query({name: "clipboard-write"})
+      .then((permissionStatus) => permissionStatus.state === "denied")
+      .catch((error) => console.error(error));
```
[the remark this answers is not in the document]
dometto: My view would be that that's not necessary: if users are expecting copy/paste functionality on websites and not getting it, that won't be experienced as a problem with gollum specifically.

dometto (commented):
I have no comments in addition to those of @bartkamphorst. Thanks for the improvement and the extensive tests, @benjaminwil!

dometto:
The fix for keyboard navigation could definitely be in a different PR, but otherwise it would be great to have the improved code buttons in the next release!

And please don't worry about holding up that release: I had an unexpectedly very busy last four months or so, so I didn't have to work on the new release anyway, until very recently!

bartkamphorst:
> Weirdly, keyboard navigation is behaving much differently in Chrome versus Firefox for all menu items. Maybe this is worth looking into separately.

I agree. And this behavior is the same on master (on Firefox) so it's not introduced by your changes. 

My only concern for this PR is whether we should be checking explicitly for whether `userAgent` is Firefox or whether we can check if (for example) `Clipboard.writeText()` is functional.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -143,23 +143,51 @@ function flashNotice(type, notice, button_label, button_function, button_type) {
...
 
+// Check whether the user has allowed their browser to access their system
+// clipboard.
+async function isClipboardAccessDenied() {
+  // Firefox doesn't support the "clipboard-write" permission but still allows
+  // websites to use the `Clipboard.writeText()` method unless the user has
+  // explicitly turned off Firefox clipboard access globally. Since we use
+  // `Clipboard.writeText()` we can return early for browsers that report as
+  // Firefox.
+  if (navigator.userAgent.indexOf("Firefox") > 0) {
```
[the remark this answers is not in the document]
dometto: FWIW, in Firefox:

* `typeof(navigator.clipboard.writeText)` appears to always return `function`
* `navigator.clipboard.writeText("clipboard test")` works depending on whether the user has recently interacted with the site ([transient user activation](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation)) 
* if the above requirement is not met, attempting to write to the clipboard will throw: `Promise { <state>: "rejected", <reason>: DOMException } Uncaught (in promise) DOMException: Clipboard write was blocked due to lack of user activation.`
* regardless of whether the transient user activation requirement is met, `navigator.permissions.query({name: "clipboard-write"})` will always throw an exception and return `undefined`: `Uncaught (in promise) TypeError: 'clipboard-write' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.`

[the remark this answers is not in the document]
dometto: So instead of testing for the presence of 'Firefox' in the UA, we could perhaps test whether the permission check returns `undefined` -- and in that case assume that access to the clipboard is not denied. But given the complexities, I'm also happy to just go with checking the UA (provided checking for "Firefox" is enough to cover also other Gecko-based browsers that have the same behavior with regard to permissions).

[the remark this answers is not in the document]
benjaminwil: The reasons I did it this way initially is because the Firefox browser really is the exception here. But I've found another way to do this, inspired by @dometto's notes above. Let me know if you can think of any improvements:

```diff
 // Check whether the user has allowed their browser to access their system
 // clipboard.
 async function isClipboardAccessDenied() {
-  // Firefox doesn't support the "clipboard-write" permission but still allows
-  // websites to use the `Clipboard.writeText()` method unless the user has
-  // explicitly turned off Firefox clipboard access globally. Since we use
-  // `Clipboard.writeText()` we can return early for browsers that report as
-  // Firefox.
-  if (navigator.userAgent.indexOf("Firefox") > 0) {
-    return false;
-  }
-
   if (navigator.permissions && navigator.permissions.query) {
     return await navigator.permissions.query({name: "clipboard-write"})
       .then((permissionStatus) => permissionStatus.state === "denied")
-      .catch((error) => console.error(error));
+      .catch((error) => {
+
+        // If the error is about a browser not supporting the 'clipboard-write'
+        // PermissionDescriptor member, then we can safely return false, as the
+        // browser should not require clipboard permissions for copy-to-clipboard
+        // functionality. Of the major browsers, only Firefox should encounter
+        // an error as of this writing.
+        //
+        // Here's Firefox's current error message for reference:
+        //
+        // TypeError: 'clipboard-write' (value of 'name' member of
+        //   PermissionDescriptor) is not a valid value for enumeration
+        //   PermissionName.
+        //
+        if (error.message.includes("clipboard-write")) {
+          console.warn(error);
+          return false;
+        } else {
+          console.error(error);
+        }
+      });
   }
 }
```

If the error message is about "clipboard-write" specifically, we emit a warning instead of an error, because the "copy to clipboard" button will _probably_ work. Otherwise we emit an error and return `undefined` from that function. Which will also let the "copy to clipboard" button _probably_ work.

benjaminwil:
@dometto @bartkamphorst Thanks for your reviews so far. I've updated the PR description and title to reflect the reduced scope of functionality it provides. Let me know if you have further comments.

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -143,23 +143,62 @@ function flashNotice(type, notice, button_label, button_function, button_type) {
   if (type == 'success') { setTimeout(function() {$('#gollum-flash').fadeOut();}, 5000); }
 }
 
+// Check whether the user has allowed their browser to access their system
+// clipboard.
+async function isClipboardAccessDenied() {
+  if (navigator.permissions && navigator.permissions.query) {
```
bartkamphorst, line 149: It would seem that nothing is returned when this `if` statement fails (e.g., if `navigator.permissions` does not exist). Is this the desired behavior or do we need an additional `else` clause?

bartkamphorst: For me, the function name `isClipboardAccessDenied` signals a boolean response type. Either access is denied or not. Calling this function, I would not expect to receive `undefined`. @dometto ?

dometto: I don't think `undefined` is too strange a result in case the permissions aren't set, but I can see how the method name could be a bit confusing. What about changing it to something more neutral, e.g. `checkClipboardAccessDenied`?

dometto:
Thanks for working on the new approach to clipboard permissions @benjaminwil! 🎆 Happy we can still release this with 6.0, which is now really, really close. :)

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: Could we rename it `checkClipboardAccess` instead and flip the return statement? The comment also says "Check whether the user has allowed their browser to access their system".

bartkamphorst: Otherwise I'm happy with `checkClipboardAccessDenied` and I'll shut up. Thanks for your patience @benjaminwil !

benjaminwil: Thanks for helping me resolve! I just pushed a fix to this commit with the new name.

benjaminwil:
Thanks again for the reviews. I've pushed up changes and recompiled the static assets. Should be ready to go. 👍

dometto (approved):

bartkamphorst (approved):
Looks good, thanks for all your work on this!

PR #1903, opened 2022-12-06
Improve Actions workflows, fix tests on CI

dometto:


dometto:
This is to avoid the double running of tests when a new PR is opened, which apparently triggers both the `push` and `pull_requests` events. Fix is (hopefully) to run tests only on pushes to the `master` branch specifically.

As a bonus, add `node_modules` to `.gitignore`.

Seems to be working now (after failed attempt in #1902), but the [tests are currently failing for different reasons](https://github.com/gollum/gollum/pull/1901#issuecomment-1339172943).

benjaminwil:
Looks like I won't be able to push to your branch after all.

But if you cherry-pick this commit: https://github.com/gollum/gollum/pull/1901/commits/d431cc73e2440ef553ac77a9f21058fa76e9f347, we are down to just JRuby and Selenium failures. ✨

dometto:
There's a JRuby bug with the latest version of `psych`: https://github.com/ruby/psych/issues/598

dometto:
@headius I see you've recently contributed a fix to the `_native_parse` method. Any idea why this might still be broken in the `gem` release (but apparently not on the `psych` CI)?

headius:
The psych bug is a release problem; the 5.0 gem was shipped with a pre-5.0 jar extension. I have updated the related issue.

headius:
Psych 5.0.1 has been released, fixing the `_native_parse` issue.

benjaminwil:
@headius Thanks for letting us know.  🙏 Have a great week.

PR #1905, opened 2022-12-08
Resize searchbar to match navigation buttons.

bartkamphorst:
I just noticed that gollum's search bar was not neatly aligned with the navigation buttons.

Screenshot 2022-12-08 at 21 48 04

This PR fixes this by giving the search input the class "input-sm".

Screenshot 2022-12-08 at 21 49 00


dometto:
Nicely caught!

PR #1907, opened 2022-12-09
Allow setting notes in commit_options from session data

### LINKED ISSUE #1766: Question: getting the IP address of the Client which performed the change on the page in the Hook
**Describe the bug**
We would like to use the gollum hook functionality to update the "Creator" and "Editor" of an article with the respective internal ip address or hostname. However i was unable to retrieve the Information from the context where the Hook is running.
Is there any chance to get the information?

**To Reproduce**
1. configure:
```
Gollum::Hook.register(:post_commit, :hook_id) do |committer, sha1|
     # some magic which retrieves the ip address of the user which created 
end
```
2. Create a new article
3. The article is created with the user account "<hostname>"

**Expected behavior**
The ip address of the creator / editor should be available for the hook

**Screenshots**

**Environment Info**
```
# gollum --version
Gollum 5.2.1
```

dometto:
Resolves #1766

* Renames the `#commit_message` method to the more accurate `#commit_options` (since it already returns author data in addition to the message)
* Adds a `:notes` key to the  hash returned by `#commit_options`, the value of which is sourced from the session
  * the relevant session key (`gollum.notes`) can be set by a piece of middleware, analogous to the way this works for author info
  
Possible use cases are e.g. getting the client IP and using this info in the `:post_commit` Hook. In the future, we could provide a feature to actually store the contents of the `:notes` field in a [git note](https://git-scm.com/docs/git-notes). For now, this only passes the info on to the `Committer` object. https://github.com/gollum/gollum-lib/pull/431 ensures the note is stored in the `Committer` object, where it can be accessed from the Hook.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -298,7 +298,7 @@ module Precious
           rename                    = "#{target_dir}/#{target_name}"
         end
 
-        committer = Gollum::Committer.new(wiki, commit_message)
+        committer = Gollum::Committer.new(wiki, commit_options)
         commit    = { :committer => committer }
 
         success = wiki.rename_page(page, rename, commit)
@@ -332,7 +332,7 @@ module Precious
           halt 412, {etag: page.sha, text_data: page.text_data}.to_json
         end
 
-        committer = Gollum::Committer.new(wiki, commit_message)
+        committer = Gollum::Committer.new(wiki, commit_options)
         commit    = { :committer => committer }
 
         update_wiki_page(wiki, page, params[:content], commit, page.name, params[:format])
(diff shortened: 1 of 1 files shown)


lib/gollum/app.rb
```diff
@@ -683,12 +683,12 @@ def wiki_new
     #     :email     - The String email address.
     # message is sourced from the incoming request parameters
     # author details are sourced from the session, to be populated by rack middleware ahead of us
-    def commit_message
+    def commit_options
       msg               = (params[:message].nil? or params[:message].empty?) ? "[no message]" : params[:message]
-      commit_message    = { :message => msg }
+      commit_options    = { message: msg, notes: session['gollum.notes'] }
```
bartkamphorst, line 688: What happens when `session['gollum.notes']` is not set? Will `notes` be `nil` or will this raise a KeyError of sorts?

bartkamphorst, line 688: Should it be `notes` or `note`?

dometto: Just tested, it's `nil` as intended 👍 .

dometto: Agreed that `note` is better!

PR #1908, opened 2022-12-16
Make docker-run.sh work if you're in a bare git repository

paulv:
Hi,

Things go sideways if you're trying to use a bare repository and a docker container and I believe this change will fix it.

Creating an empty .git directory is another workaround.

### DIFF
--- a/docker-run.sh
+++ b/docker-run.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Initialize the wiki
-if [ ! -d .git ]; then
+if [ ! -d .git ] && [ "$(git rev-parse  --is-bare-repository 2>/dev/null)" != "true" ]; then
     git init
 fi
 


benjaminwil:
Thanks for the pull request!

I will manually test this change against a bare repository and a non-bare repository before approving.

benjaminwil:
I manually tested this and it works as intended! Thank you.

After building the docker image, I was able to run it against a regular clone and a bare clone of the same wiki without issue.

Bare clone:

```sh
git clone --bare ~/my-wiki ~/bare-test
docker run --rm -p 4567:4567 ~/bare-test:/wiki my-new-docker-image
```

Clone:

```sh
git clone ~/my-wiki ~/test
docker run --rm -p 4567:4567 ~/test:/wiki my-new-docker-image
```

guillaume-d:
@benjaminwil Maybe this should be tagged with a 6.0 milestone for clarity, to avoid others wondering why it is not yet in the official Docker image (`docker run --rm --entrypoint cat docker.io/gollumwiki/gollum:v5.3.1 /docker-run.sh | grep bare` returns nothing)...

...or is https://github.com/gollum/gollum/blob/master/LATEST_CHANGES.md only meant for big things?

benjaminwil:
Thanks @guillaume-d.

I haven't done any release management for Gollum yet. @bartkamphorst, @dometto: The change being referenced here doesn't seem like it should be with-held for a 6.0 release. Seems like more of a `5.3.2` patch to me. But I don't want to step on anyone's toes. I see [the docs about releases](https://github.com/gollum/gollum/blob/master/CONTRIBUTING.md#releasing-the-gem), but am unsure if I have permission to do this, or do anything on the `gollumwiki` Docker account.

(As an aside: I'd be happy to do release chores in the future. Just unsure about the specifics of the process.)

dometto:
I wasn't intentionally withholding this from the `5.x` image. Images are pushed to Dockerhub automatically after a push to `master`, which happened also after the PR that introduced the fix for this issue: https://github.com/gollum/gollum/actions/runs/3728260498/jobs/6323105289

So it really *should* be present in the latest Docker image. If it's really not there, we need to investigate!

@benjaminwil would be great to explain the build/release steps to you so there's someone else who can take care of this. I'll write you an email, or just update the release docs with a bit more detail.

benjaminwil:
Hi @guillaume-d, just sat down to check out your issue:

> Maybe this should be tagged with a 6.0 milestone for clarity, to avoid others wondering why it is not yet in the official Docker image

For clarity, I'll say here that it _is_ in the official Docker image, but not part of the `v5.3.1` tagged version. If you need this feature today, you'll have to use the `master tagged version—until we cut a new release. (Likely `v5.3.2`?)

Since this is essentially a bug fix, I don't know that it requires more explicit documentation then this comment. Hope that makes sense.

```sh
docker run --rm --entrypoint \cat docker.io/gollumwiki/gollum:master /docker-run.sh | grep bare

PR #1910, opened 2022-12-20
Remove uglifier. Resolves #1909

### LINKED ISSUE #1909: Move uglifier to development dependency for faster package build
**Describe the bug**
I think uglifier should be a development dependency and not a runtime dependency as it is right now.
As far as I can tell it is not required to run gollum, the only use is here -> https://github.com/gollum/gollum/blob/869292ed1e5018102729db1b1d46815807761e34/Rakefile#L251

- https://github.com/gollum/gollum/blob/master/gollum.gemspec#L35
- https://apidock.com/ruby/Gem/Specification/add_dependency

Why does this matter right now the FreeBSD ports requires that which then in requires 
nodejs to be built which is a big port which takes long.
`gollum -> uglifier -> execjs -> nodejs` dependency tree 
(Quickly checked on arch linux and its the same there i suspect it is packaged like this in most places)

**Expected behavior**
If it is move out of the runtime dependency the gollum package could be built faster (less dependencies to run).

dometto:
As described in #1909, we only use the `uglifier` gem when generating the packaged assets. This PR moves it to the development dependencies.

Additionally, I have added the ability to specify a custom location for the `node_modules` directory when using the `--development-assets` option.

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -204,7 +204,7 @@ MSG
   opts.separator ''
   opts.separator '  Development:'
 
-  opts.on('--development-assets [PATH]', 'For development purposes only. Use local node modules instead of the packaged assets. Path defaults to ./node_modules.') do |path|
+  opts.on('--development-assets [PATH]', 'For development purposes only. Use assets in PATH (default: `node_modules` in the gollum project root) instead of the packaged assets.') do |path|
     if path
       module Precious
         module Assets
--- a/bin/gollum
+++ b/bin/gollum
@@ -199,19 +199,11 @@ MSG
     puts results.empty? ? 'none' : results
     exit 0
   end
-
-
+  
   opts.separator ''
   opts.separator '  Development:'
 
-  opts.on('--development-assets [PATH]', 'For development purposes only. Use assets in PATH (default: `node_modules` in the gollum project root) instead of the packaged assets.') do |path|
-    if path
-      module Precious
-        module Assets
-        end
-      end
-      ::Precious::Assets.const_set('NODE_MODULES', path)
-    end
+  opts.on('--development-assets', 'For development purposes only. Use local instead of packaged assets.') do
     wiki_options[:static] = false
   end
 end
--- a/bin/gollum
+++ b/bin/gollum
@@ -204,7 +204,14 @@ MSG
   opts.separator ''
   opts.separator '  Development:'
 
-  opts.on('--development-assets', 'For development purposes only. Use local stylesheets and javascript instead of the packaged assets.') do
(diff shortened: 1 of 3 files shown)


lib/gollum/assets.rb
```diff
@@ -4,20 +4,21 @@ module Precious
...
+        ::File.join(dir, '../../node_modules')
 
       env.append_path ::File.join(dir, 'public/gollum/javascript')
       env.append_path ::File.join(dir, 'public/gollum/stylesheets/')
 
       env.append_path ::File.join(dir, 'public/gollum/images')
       env.append_path ::File.join(dir, 'public/gollum/fonts')
 
-      env.js_compressor  = Precious::Assets::JS_COMPRESSOR unless Precious::App.development?
+      env.js_compressor  = ::Precious::Assets::JS_COMPRESSOR if defined?(::Precious::Assets::JS_COMPRESSOR)
```
dometto, line 21: A behavioral change here: whether `uglifier` will be used only depends on whether `Precious::Assets::JS_COMPRESSOR` is defined, and not on the `RACK_ENV`. So a user can enable uglifier in any `RACK_ENV`, but by default it is turned off in any mode. We enable it ourselves in the `precompile` task in our `Rakefile`.

bin/gollum
```diff
@@ -204,7 +204,14 @@ MSG
...
 
-  opts.on('--development-assets', 'For development purposes only. Use local stylesheets and javascript instead of the packaged assets.') do
+  opts.on('--development-assets [PATH]', 'For development purposes only. Use assets in PATH (default: `node_modules` in the gollum project root) instead of the packaged assets.') do |path|
+    if path
+      module Precious
+        module Assets
+        end
+      end
+      ::Precious::Assets.const_set('NODE_MODULES', path)
+    end
```
benjaminwil, line 214: I wonder if there's a way we can provide this flag to developers without advertising a flag to end users.

I would personally be happy to use an "undocumented" environment variable instead. i.e.:

```
GOLLUM_DEV_ASSETS=path/to/assets bundle exec gollum ~/my-wiki
```

dometto: I also like this idea. We could manually filter out the option [here](https://github.com/gollum/gollum/blob/master/bin/gollum#L174)?

benjaminwil: Great idea!

dometto: This would have become a bit hackish as there is no clean way of turning  the`opts` object into an array of option-like objects -- we can only get an array of the lines of the help text. This means we'd have to filter out each line in a description separately. So I opted to leave it in the help text, as it's clearly marked as a development option. Since it is marked as such, we needn't give full documentation for the option. So I've just implemented your idea of the `GOLLUM_DEV_ASSETS` environment variable without documenting it.

PR #1919, opened 2022-12-22
Upgrade mermaid to 9.3

dometto:
Upgrades mermaid to 9.3. This requires using `terser` instead of `uglifier` (`uglifier` hasn't been updated for three years and has poor support for ES6).

On `sprockets` 4.0 I was seeing errors attempting to precompile mermaid. Turns out the problem was there was a clash between our file `mermaid.js`, and the file `mermaid.js` in the `node_modules/mermaid/dist` directory. Presumably `sprockets` 4.0 resolves filenames in a different way. I've solved the issue by renaming our file to `gollum.mermaid.js`.


bartkamphorst:
Could we make the change to Terser into a separate PR? That seems conceptually cleaner to me. Now this (possibly significant) change of library is hidden behind a mermaid update PR.

dometto:
> Could we make the change to Terser into a separate PR?

Done: https://github.com/gollum/gollum/pull/1920

Will rebase this PR when #1920 is merged.

bartkamphorst:
Thanks!

PR #1920, opened 2022-12-22
Move from Uglifier to Terser

dometto:
`uglifier` hasn't been updated for three years and has poor support for ES6. Precompiling newer versions of e.g. mermaid with uglifier leads to errors. This PR switches to the more modern `terser` gem for JS minification.


bartkamphorst (approved):
Straightforward enough!

PR #1915, opened 2022-12-21
Clean up unused styling (CSS/SCSS)

bartkamphorst:
There is still quite a lot of SCSS in gollum that is no longer used, or at least no longer needed, since introducing Primer. This PR attempts some cleaning up.

It also upgrades Primer CSS manually from v.15 to v.20.8.0.

PR #1914, opened 2022-12-21
Move primer to yarn

dometto:
Update primer to the latest version (20.8.0) via yarn.

PR #1881, opened 2022-11-21
Add rounded corners to tables. Resolves #1693.

bartkamphorst:
This PR resolves #1693 by adding rounded corners to gollum's tables. 
Screenshot 2022-11-21 at 20 58 07

PR #1937, opened 2023-01-12
Move primer to yarn

dometto:
Take 2. :)

It tuns out the 'product' component is necessary to style things like tables in the markdown body. However, this does not work neatly unless we import the primer components *after* importing our own `template.scss`. In turn, this caused primer to override some of our own styling from `template.scss`. I have addressed this by identifying which parts of `template.scss` are superfluous, which need to be included before primer, and which need to be included after primer.

- [x] Move our own `.markdown_body` styling to a separate file
- [x] The spinner with 'When was this page last modified?' is no longer there
- [x] Rebuild assets


lib/gollum/public/gollum/stylesheets/app.scss
```diff
@@ -1,8 +1,102 @@
...
+
+//  Simply importing @primer/css/index.scss also imports the unneeded marketing styles, so import the following components manually:
+//= require @primer/css/color-modes
+//= require @primer/css/core
+//= require @primer/css/product
+
+@import '_base';
+
+body {
+    font: 13.34px $font-default;
```
dometto, line 17: @bartkamphorst @benjaminwil: I've moved this here from `template.scss` in order to reproduce the current styling as it exists on `master`. However, we could also leave this out and go with the (new) primer defaults. The latter also looks nice to me; I have no strong preference. Thoughts?

lib/gollum/public/gollum/stylesheets/app.scss
```diff
@@ -1,8 +1,102 @@
...
+    font-size: small;
+    line-height: 1.4;
+    counter-reset: h1;
+}
+
+.markdown-body {
+
+    img {
+        max-width: 100%;
```
dometto, line 26: Not sure if this is needed

lib/gollum/public/gollum/stylesheets/app.scss
```diff
@@ -1,8 +1,102 @@
...
+
+        > tt, > code {
+            margin: 0;
+            padding: 0;
+            white-space: pre;
+            border: none;
+            background: transparent;
+        }
+
+        button.btn-octicon {
```
dometto, line 48: This is the 'copy to clipboard button', see comment below.

lib/gollum/public/gollum/stylesheets/app.scss
```diff
@@ -1,8 +1,102 @@
...
+@import '_base';
+
+body {
+    font: 13.34px $font-default;
+    font-size: small;
+    line-height: 1.4;
+    counter-reset: h1;
+}
+
+.markdown-body {
```
dometto, line 23: All of this needs to override the primer defaults

lib/gollum/public/gollum/stylesheets/app.scss
```diff
@@ -1,8 +1,102 @@
...
+    }
+
+    pre:hover {
+        button {
+            animation: fade-in 200ms both;
+        }
+    }
+}
+
+/* Table of contents */
```
dometto, line 63: Similarly, this is needed to override the primer defaults

lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -164,10 +164,7 @@ function enable_code_blocks() {
     button.classList.add(
       "btn", 
       "btn-octicon", 
-      "border",
```
dometto, line 167: The copy to clipboard button was broken after updating primer. It looks like the `top-2` and `right-2` classes don't exist in the latest release of primer. For that reason, I'm now styling this button in `app.scss`.

dometto:
@benjaminwil @bartkamphorst I believe this solves all the styling issues @bartkamphorst discovered after having (prematurely) merged #1914. The problem was that some of our own styling from `template.scss` was being overridden by primer. Annoyingly, however, importing `template.scss` *after* primer also didn't work (presumably because some `primer` definitions were being used in `template`). So in this PR, I have:

1. Radically cut some styling from `template.scss` that it looks like primer is already providing. I have tested as best as I could, but it would be great if you could give this branch a spin and see if you see anything strange.
2. Moved some of the styling from `template.scss` that we don't want primer to override into `app.scss`, after primer is included.

If this approach seems sane, I'll move the new styling in `app.scss` to a separate new file (something like `wiki_content.scss`).

lib/gollum/public/gollum/stylesheets/template.scss.erb
```diff
@@ -199,311 +173,6 @@ img {
...
-    > span {
-      display: block;
-      overflow: hidden;
-      margin: 13px auto 0;
-      text-align: right;
-    }
-  }
-
-  code, pre, tt {
-    font-family: $font-console;
```
dometto, line 415: Note: removing this means we now use the same font for normal text as for text in code blocks. I think this looks fine, but I'm happy to restore this rule!

lib/gollum/public/gollum/stylesheets/app.scss
benjaminwil: No real preference, although I think using the Primer defaults as much as possible benefits all of us.

bartkamphorst:
Overall it seems to look okay. Two things I noticed straight away:

1) The styling of the tables no longer has the rounded corners from #1881 .
Screenshot 2023-01-14 at 17 40 39

2) The spinner with 'When was this page last modified?' is no longer there.

lib/gollum/public/gollum/stylesheets/template.scss.erb
bartkamphorst: Github seems to be using monospace, so I would suggest we do so as well.

lib/gollum/public/gollum/javascript/gollum.js.erb
bartkamphorst: I'm not sure that diagnosis is correct. The screenshots below are from `master` with primer 20.8.

Screenshot 2023-01-15 at 14 51 33
Screenshot 2023-01-15 at 14 51 49

dometto: Thanks for pointing this out. I just looked into it and it turns our `.right-2` is only defined in the primer `marketing` component:

```
$ grep -r "\.right-2" node_modules/@primer/css | awk -F":" '{print $1}'
node_modules/@primer/css/dist/primer.css
node_modules/@primer/css/dist/marketing-utilities.css
node_modules/@primer/css/dist/marketing.css
node_modules/@primer/css/dist/stats/primer.json
node_modules/@primer/css/dist/stats/marketing.json
node_modules/@primer/css/dist/stats/marketing-utilities.json
```

I am not sure whether it's worth importing the entire marketing style just for the `right-2` styling. :)

dometto:
> This seems to be straightforwardly caused by the table styling being removed from template.scss.erb.

Thanks for catching! Re-added that styling.

> The spinner with 'When was this page last modified?' is no longer there.

Again thanks for catching. Not sure yet why.

I've added the spinner issue to the todo for this PR. Apart from that issue, I think this is good to go. We may of course still discover slight differences in the styling in the future, but if they need to be addressed at all, it should be straightforward to fix them by looking at the changes in this PR.

lib/gollum/public/gollum/stylesheets/wiki_content.scss
```diff
@@ -0,0 +1,157 @@
...
+
+        > tt, > code {
+            margin: 0;
+            padding: 0;
+            white-space: pre;
+            border: none;
+            background: transparent;
+        }
+
+        button.btn-octicon {
```
dometto, line 77: @benjaminwil since we're not using the Primer [marketing component](https://primer.style/css/marketing/) I had to style the copy-to-clipboard button here. Does that interfere with the work you're doing that you just mentioned in #1942? If so I can remove this, or even add the marketing component back in, if you feel that would be useful

benjaminwil: I don't think it interferes!

bartkamphorst:
The problem with the spinner can be straightforwardly resolved by adding back the `//= require spinner` directive in `app.scss`. 😄

lib/gollum/public/gollum/stylesheets/wiki_content.scss
bartkamphorst: I really would favor a solution though that uses Primer utilities, rather than styling this button ourselves. 

[Here](https://primer.style/css/utilities/marketing-layout) it is explained that the marketing layout utilities are built on top of the [Primer core utilities](https://primer.style/css/utilities/layout#position) so presumably we could position the button using only core utility classes?

benjaminwil: That would be my preference to, and absolutely how I will approach refactoring this button in #1901.

dometto:
Ok, I'm merging this. This was more complicated than I hoped, and as I said there may be tiny style issues that I've missed. But I think it should be much easier to fix such issues now (also after #1915). Thanks for the feedback @benjaminwil and @bartkamphorst!

lib/gollum/public/gollum/stylesheets/wiki_content.scss
dometto: As a heads up, it seems that (very surprisingly) only `right-0` (and not `right-1` etc.) is defined in the core utilities. That's what led me to style the button myself. Or maybe I overlooked something.

Anyway, I'll merge this PR with the self-styled button, since @benjaminwil will take care of it in #1901! Many thanks.

PR #1941, opened 2023-01-16
Enable downloading page sources with ?raw.

### LINKED ISSUE #1906: Feature request: Download raw Markdown source for a page
I would like an API for grabbing the raw, unrendered Markdown source for a page, like the "Raw" button on GitHub:

https://raw.githubusercontent.com/gollum/gollum/master/CONTRIBUTING.md

I understand I can access the underlying file on disk, or use the Edit button and copy the contents of the page.

This would make it easy to access a page from a script, or to export it to make a plain text attachment on an email, etc.

tstein:
resolves #1906

Folks on that issue sounded okay with this implementation, and imo it's better to break literally no one than to expand the set of paths that clash with user content.

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -590,10 +590,25 @@ module Precious
     end
 
     get '/*' do
-      show_page_or_file(params[:splat].first)
+      fullpath = params[:splat].first
+      if params.has_key?("raw")
+        show_raw_page(fullpath)
+      else
+        show_page_or_file(fullpath)
+      end
     end
 
     private
+
+    def redirect_to(redirect_path, fullpath, query_params)
+        redirect to("#{encodeURIComponent(redirect_path)}?redirected_from=#{encodeURIComponent(fullpath)}#{query_params}")
+    end
+
+    def page_does_not_exist()
+      @message = "The requested page does not exist."
+      status 404
+      return mustache :error
+    end
 
     def show_history(wikip)
       @name      = wikip.fullname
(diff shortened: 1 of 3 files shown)


lib/gollum/app.rb
```diff
@@ -590,7 +590,7 @@ class App < Sinatra::Base
     end
 
     get '/*' do
```
bartkamphorst, line 592: I don't have all the details at the moment, but would it be more straightforward to check if `params.has_key?("raw")` here in this route, and, if so, call a new method `show_raw_contents(params[:splat].first)`?

tstein: It makes sense to me to use `show_file` because `raw` essentially means "retrieve that wiki page and skip rendering" - basically just "give me the file". If you accept that, `show_page_or_file` looks like the perfect place to use the existing call to `show_file` and also to only give `?raw` an effect when you know you're serving an actual wiki page (and not a redirect to another page or a creation view).

If you don't buy that, I'll defer to you, of course. :)

lib/gollum/app.rb
```diff
@@ -613,9 +613,9 @@ def show_history(wikip)
       end
     end
 
-    def show_page_or_file(fullpath)
+    def show_page_or_file(fullpath, raw)
```
bartkamphorst, line 616: Would it be sensible to default `raw` to `false` in this method signature?

benjaminwil: Based on the way it's implemented, it should always be provided a value. So it's technically okay. But I think there's a benefit to providing a default value and ensuring readers see what the "happy path" (read as: non-raw) looks like.

I wouldn't block merge, but I think that this implementation could be simpler by extracting raw-specific concerns to a separate method:

```
def show_page_or_file(fullpath, raw: false)
  return show_raw_page_or_file(fullpath) if raw

  # The already-committed implementation of `#show_page_or_file`.
end

def show_raw_page_or_file(fullpath)
  # Only the parts of the if statement in this diff where the presence
  # of `raw` matters.
end
```

This results in more lines of code, but less room for error.

lib/gollum/app.rb
```diff
@@ -590,7 +590,7 @@ class App < Sinatra::Base
     end
 
     get '/*' do
-      show_page_or_file(params[:splat].first)
+      show_page_or_file(params[:splat].first, params.has_key?("raw"))
```
benjaminwil, line 593: If you decided to try out my suggested method signature, this would also need to be changed to match:

```suggestion
      show_page_or_file(params[:splat].first, raw: params.has_key?("raw"))
```

tstein: Done.

lib/gollum/app.rb
tstein: Added the default. Just for fun, I did your suggested version on a different branch: https://github.com/tstein/gollum/commit/02fbf7ad67e4a792a74af6aed8b6953539318219

I also did it the way you suggested - IMO the duplication necessitates some very small helpers, and overall I think it's not worth the extra volume. lmk what you prefer, it'll only take a moment to swap that commit in for this one. :)

lib/gollum/app.rb
tstein: Done, and see the other thread - I took a stab at the other approach in another branch, linked there.

lib/gollum/app.rb
bartkamphorst: I like the idea of extracting a `page_not_found` helper (from the alternative branch), I think that improves the readability of the logic overall. More generally, I'm still on board with @tstein 's logic here and using `show_file` to treat a page as a file when `raw` is true.

tstein: Works for me. This PR is now for https://github.com/tstein/gollum/commit/02fbf7ad67e4a792a74af6aed8b6953539318219.

dometto:
Thanks for working on this @tstein! Could I ask you to add a small line to `LATEST_CHANGES.md` highlighting your contribution? We should update the PR template to request people to do so. :)

tstein:
Sure thing - done.

bartkamphorst (approved):

bartkamphorst:
Okay to merge, @benjaminwil ?

benjaminwil:
Yes! Thanks @tstein

bartkamphorst:
Thanks @tstein , we appreciate your contribution, as well as your patience and willingness to take a slightly modified approach. If there are any other issues you'd like to take a stab at, please let us know and we'd be happy to discuss options and help you on your way.

tstein:
:+1: Thanks for the review and for your work on gollum, I'm a big fan. :)

PR #1944, opened 2023-01-18
Use to_xml defaults from gollum-lib

dometto:
The `Page` view was setting its own xml conversion options rather than using the identical settings from `gollum-lib`.

I came across this while trying to fix a nokogiri error in JRuby. See https://github.com/sparklemotion/nokogiri/issues/2761 for context.


bartkamphorst:
Good catch!

PR #1954, opened 2023-03-13
Fix mustache

dometto:
The `mustache-sinatra` gem just received a bump release, which now utillizes the latest `mustache`. Unfortunately, this has suddenly exposed `gollum` to breaking changes in the latest `mustache`. I found two:

## Broken call to View#escape

Error: `NoMethodError private method escape called for <<Precious::Views::Create: ...`.

**Diagnosis**:  in the latest `mustache`, views are expected to have an `escape` instead of an `escapeHTML` method. We were using the `escapeHTML` method provided by the `Rack::Utils` modules. By including that module, we also include a private `#escape` method, which mustache is now attempting to call.

**Fix**: instead of including `Rack::Utils`, just use the default `Mustache#escape`, which uses `CGI.escapeHTML`. Overriding the `#escape` method is only necessary for providing custom (non-HTML) escaping.

## Broken template cascades

Our logic for template cascading fails, causing a test failure for using custom partials inside a custom template (`"test_overridden_navbar_partial_is_used"` in `test_template_cascade.rb`).

**Diagnosis**: we provided a custom *class* method `Layout.partial`, but the newest version of mustache ended up calling an *instance* method `Page#partial`, i.e. the mustache default. So our overriding logic was simply not being called.

**Fix**: providing the necessary instance method `#partial` which provides the necessary overriding logic. I have still implemented the helper methods as class methods.

@bartkamphorst I would appreciate a review of this fix, since I'm not really familiar with the template cascade functionality.

This fix would only be included in the next major release of gollum. Should we backport a fix for `5.x`, pinning `mustache` to an older version?

### DIFF
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -15,6 +15,7 @@ require 'gollum'
 require 'gollum/assets'
 require 'gollum/views/helpers'
 require 'gollum/views/helpers/locale_helpers'
+require 'gollum/views/template_cascade'
 require 'gollum/views/layout'
 require 'gollum/views/editable'
 require 'gollum/views/has_page'
@@ -22,7 +23,6 @@ require 'gollum/views/has_user_icons'
 require 'gollum/views/has_math'
 require 'gollum/views/pagination'
 require 'gollum/views/rss.rb'
-require 'gollum/views/template_cascade'
 
 ['sassc', 'sassc-embedded'].each do |gem|
   require gem if Gem::Specification.find {|spec| spec.name == gem}
@@ -118,7 +118,6 @@ module Precious
       @default_keybinding = settings.wiki_options.fetch(:default_keybinding, 'default')
 
       if settings.wiki_options[:template_dir]
-        Precious::Views::Layout.extend Precious::Views::TemplateCascade
         Precious::Views::Layout.template_priority_path = settings.wiki_options[:template_dir]
       end
 
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -13,9 +13,12 @@ require 'pathname'
 
 require 'gollum'
 require 'gollum/assets'
+
+# Requirements for loading the layout view:
 require 'gollum/views/helpers'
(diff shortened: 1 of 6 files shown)


dometto:
> This fix would only be included in the next major release of gollum. Should we backport a fix for 5.x, pinning mustache to an older version?

If this fix is approved, I think we should backport it, because we would then also have a version of gollum compatible with ruby 3.2 :+1:

lib/gollum/app.rb
```diff
@@ -15,14 +15,14 @@
...
 require 'gollum/views/helpers/locale_helpers'
+require 'gollum/views/template_cascade'
 require 'gollum/views/layout'
 require 'gollum/views/editable'
 require 'gollum/views/has_page'
 require 'gollum/views/has_user_icons'
 require 'gollum/views/has_math'
 require 'gollum/views/pagination'
 require 'gollum/views/rss.rb'
-require 'gollum/views/template_cascade'
```
benjaminwil, line 25: Is the order of these requires important, or is it just an incidental change?

If it's important, it might be worthwhile to add an inline comment explaining why, so the future versions of us don't get caught by the order-related issues again.

benjaminwil (approved):

benjaminwil:
I worked on the unit tests for the template cascade feature. I think they describe the feature pretty well if that's helpful.

This may be a good opportunity to add a basic Capybara test for them, too. 👍

lib/gollum/app.rb
dometto: Yes, order is important now because our layout now includes `TemplateCascade` directly, instead of the layout being extended dynamically later on in this file. I'll add a note!

dometto:
> This may be a good opportunity to add a basic Capybara test for them, too. 👍

That's a good idea, but I don't see a good way of getting this to work on CI yet, since we're running the tests against a separate gollum instance there. That situation isn't ideal but I don't see a quick way of doing better. Since the normal template_cascade tests picked up the errors here, and since you worked on the tests, I feel this can be merged. :)

Thanks for your feedback!

Issue #1959, opened 2023-03-27
matrix and magnet links won't work with AsciiDoc


tastytea:
[Matrix specifies an URI scheme](https://github.com/matrix-org/matrix-spec-proposals/blob/old_master/proposals/2312-matrix-uri.md). A link to a user account looks like this, for example: `matrix:u/foo:example.org`.

**Describe the bug**
Matrix links work with asciidoctor like this: `link:matrix:u/foo:example.org[foo]`. However, they don't work with asciidoctor in gollum. There is no hyperlink generated at all. asciidoctor versions are the same.

The same problem exists with magnet links, like `link:magnet:?xt=urn:sha1:YNCKHTQCWBTRNJIV4WNAE52SJUQCZO5C[bar]`.

**To Reproduce**
1. Create asciidoc wiki page with `link:matrix:u/foo:example.org[foo]`
2. look at it in the browser

**Expected behavior**
`<a href="matrix:u/foo:example.org">foo</a>` in the generated HTML.

**Environment Info**
```
Gollum 5.3.1
Running on: x86_64-linux with Ruby version 2.7.7
Using:
rugged 1.6.2
gollum-rugged_adapter 2.1.0
gollum-lib 5.2.3
Markdown rendering gem: kramdown
Other renderering gems:
asciidoctor 2.0.18
```

benjaminwil:
Hey there, thanks for reporting this issue.

I was able to reproduce your issue:

- Gollum does not render Magnet and and Matrix prefixed URLs.
- Using `Asciidoctor#convert` directly _does_ render those links.

I was able to determine that `gollum-lib` uses the gem `loofah` to sanitize URLs added to wiki pages. `loofah` includes a list of acceptable protocols: `::Loofah::HTML5::SafeList::ACCEPTABLE_PROTOCOLS`. [Link to current source](https://github.com/flavorjones/loofah/blob/5f7efdb02ed092f4dd73a88c451c66cdd12baaca/lib/loofah/html5/safelist.rb#L980-L1009). It seems to me that `loofah` is missing a lot of modern and weirdo protocols: not only `magnet:` and `matrix:`, but also `gemini:` and others.

I haven't done any research yet into what `loofah` chooses which protocols are "acceptable" versus "not acceptable," so I am unsure whether submitting a patch upstream to `loofah` would be fruitful or not. But it may be worth learning whether that's an option or not.

[...]

PR #1963, opened 2023-04-16
Add browser tests for Asciidoctor rendering

benjaminwil:
This pull request reproduces the issue reported in #1959, although it doesn't resolve it. When we resolve the issue, we can un-skip the end of the test to ensure that this is true.

Unrelatedly, I found a stale `FIXME` comment and removed it while I was editing the Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,21 +2,6 @@ source 'https://rubygems.org'
 
 gem 'warbler', platforms: :jruby
 
-# FIXME:
-#
-# There's an issue in 1.12.5 that causes XHTML elements to be generated badly,
-# causing Gollum's test suite to fail.[1] The issue has been fixed upstream,
-# but we're still waiting for a new Nokogiri point release.
-#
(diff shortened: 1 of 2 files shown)


bartkamphorst:
This may be a point for discussion. I actually don't think we should be adding `asciidoctor` to the `Gemfile` or adding tests for all possible markups, as we are relying on `github/markup` to ensure that the correct renderer is called. Of course, if we're doing something to the input to or output from `asciidoctor` that is causes issues, we want to resolve that, but otherwise we want to remain agnostic. Any thoughts on this, @dometto  ?

dometto:
I think we have tests for specific markups in `gollum-lib`, testing e.g. codeblocks and such with asciidoc and orgmode (with those gems added to the `Gemfile` only for testing/development). If we want to add tests for links with certain protocols, I think it makes sense to do it in `gollum-lib`, since the cause of the undesired behavior lies there.

benjaminwil:
That all makes sense to me. I will close this pull request. 👍

PR #1968, opened 2023-04-29
Style Note and Warn macros and inject requested Octicon icons for Flash and Icon macros.

bartkamphorst:
This PR accompanies https://github.com/gollum/gollum-lib/pull/441, which eliminates the octicons gem from gollum-lib. This PR ensures that
* `Warn` and `Note` macros are styled with a default Octicon using a `::before` pseudo-element;
* Icon and Flash macros are styled with an Octicon of the user's choice. 

The latter works through jQuery html injection and a new `octicon` route, which uses the existing `OcticonHelpers` to lookup and prepare the appropriate octicon svg. 

Screenshot 2023-04-29 at 19 28 24

If a requested icon name is not available, a broken img link is inserted instead (with a helpful `title`), to signal to the user that something went wrong.

Screenshot 2023-04-29 at 19 27 29

Open issues:
- [x] Alignment of the `::before` icons is a little wonky. 

Screenshot 2023-04-29 at 19 27 22

@benjaminwil Any suggestions on how to improve the alignment with the text?

### DIFF
--- a/lib/gollum/public/gollum/stylesheets/template.scss.erb
+++ b/lib/gollum/public/gollum/stylesheets/template.scss.erb
@@ -181,7 +181,7 @@ nav.actions {
   display: inline-block;
   position: relative;
   padding-right: 0.5em;
-  align-items: center;
+  vertical-align: middle;
   width: 24px;
   height: 24px;
 }
--- a/lib/gollum/public/gollum/stylesheets/template.scss.erb
+++ b/lib/gollum/public/gollum/stylesheets/template.scss.erb
@@ -175,3 +175,31 @@ nav.actions {
    }
 }
 
+/* `gollum`-provided CSS for the note and warning macros. */
+.flash.gollum-note::before {
+  content: url('data:image/svg+xml;utf8,<%= rocticon_css(:info) %>');
+  display: inline-block;
+  position: relative;
+  vertical-align: middle;
+  padding-right: 0.5em;
+  width: 24px;
+  height: 24px;
+}
+
+.flash.gollum-warning::before {
+  content: url('data:image/svg+xml;utf8,<%= rocticon_css(:alert) %>');
+  display: inline-block;
+  position: relative;
+  vertical-align: middle;
+  padding-right: 0.5em;
+  width: 24px;
+  height: 24px;
+}
+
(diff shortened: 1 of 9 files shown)


bartkamphorst:
CI Tests fail because the PR depends on the gollum-lib PR I mentioned. Tests pass locally.

lib/gollum/public/gollum/stylesheets/template.scss.erb
```diff
@@ -175,3 +175,31 @@ nav.actions {
    }
 }
 
+/* `gollum`-provided CSS for the note and warning macros. */
+.flash.gollum-note::before {
```
benjaminwil, line 179: I think because, for very long notes, it would be reasonable for the icon to still appear at the top left of the note-container, it would make sense to just manually tweak the `padding-top` or the `top` value of the `.flash` icon area. Below I'll paste some suggested CSS changes.

I think we can also reduce some of the repetition by styling `.flash` by itself, or using another class that is shared between both `.gollum-note` and `.gollum-warning`:

```suggestion
.flash::before {
  display: inline-block;
  position: relative;
  padding-top: 0.5em; /* this value may require some tweaking */
  padding-right: 0.5em;
  width: 24px;
  height: 24px;
}
.flash.gollum-note::before {
  content: url('data:image/svg+xml;utf8,<%= rocticon_css(:info) %>');
}

.flash.gollum-warning::before {
  content: url('data:image/svg+xml;utf8,<%= rocticon_css(:alert) %>');
}
```

bartkamphorst: Thanks @benjaminwil ! Makes sense to reduce repetition in this way. I wasn't quite getting the results I wanted with `padding-top` however, so I've now used `vertical-align: middle` instead. This is what the flash elements look like now:
Screenshot 2023-05-01 at 19 58 37

Still not perfect, but pretty good. Suggestions for further improvement welcome of course!

dometto:
I like where this is going, will be great for maintainability when octicons are fully de-entangled. :) I'll try to do a 6.0 release of `gollum-lib` as soon as possible, so that we can start merging this and other 6.0 `gollum` functionality.

dometto (approved):
Icons look good now, thanks for the restyling!

PR #1994, opened 2023-08-04
Fix colors in search results

bartkamphorst:
I just noticed that my search results weren't being highlighted properly. This is due to a change in Primer's color naming scheme some time ago. This PR resolves the issue.

Screenshot 2023-08-04 at 09 44 55
Screenshot 2023-08-04 at 09 44 39


lib/gollum/public/gollum/javascript/editor/gollum.editor.js
```diff
@@ -306,6 +306,11 @@
               formData = new FormData();
           formData.append('file', file);
           
+          if (Object.prototype.toString.call(file) != '[object File]') {
+            $editorBody.removeClass('uploading');
+            return;
+          }
+
```
benjaminwil, line 313: Looks like this commit sneaked in from another branch.

benjaminwil (approved):

lib/gollum/public/gollum/javascript/editor/gollum.editor.js
bartkamphorst: Thanks, just fixed this through a rebase.

dometto (approved):

dometto (approved):
Looks good, to take effect the static assets would have to rebuilt after this. However, that'll happen anyway before the 6.0 release, so your choice whether you want to do that in this PR or wait for it to happen down the line!

PR #1997, opened 2023-08-15
[WIP] Improve tabindex for navigation bar on Firefox.

bartkamphorst:
This PR improves gollum's tabbing behavior on Firefox on OS X. 

On current `master`, not all buttons from the navigation bar are "tabbable" in Firefox 116.0.1 on OS X.  

Screenshot 2023-08-14 at 16 10 44

The difference can be explained by looking at the underlying HTML: some buttons are actual `button`s (or divs) while others are `<a>` links that are styled as buttons. Those latter ones do not get focus, not even when `tabindex` is set explicitly to "0" (follow default tabbable order) or "1". 

Apparently, this behavior is caused by Firefox being true to [OS X preferences](https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox). It can be "solved" by changing those preferences in OS X and/or changing some setting in Firefox. Alternatively, we can get more uniform behavior by either:

1. Changing all `<a>` elements into proper `button` elements. Upside: uniformity. Downside: the buttons that used to be links will need some `onClick` behavior to mimic the link behavior (i.e., setting `window.location.href`). That means all buttons will require JS to work.
2. Wrapping some of the `<a>` text in a `div` with `tabindex="0"` set. Upside: links will continue to work as they did before. Downside: less uniformity, as technically the tab focus will shift to the internal `div` and not to the button itself. 

Screenshot 2023-08-15 at 13 42 36

I've implemented the second option, but I'm open to change course and go the first route. Let me know what you think! 

TODO:

- [ ] Test on Chrome
- [ ] Test on MS Edge

### DIFF
--- a/lib/gollum/templates/navbar.mustache
+++ b/lib/gollum/templates/navbar.mustache
@@ -10,39 +10,23 @@
     {{>mobilenav}}
   </div>
 
-  <div class="TableObject-item hide-sm hide-md">
-    <a class="btn btn-sm" id="minibutton-home" href="{{page_route}}">
-      Home
-    </a>
+  <div class="TableObject-item hide-sm hide-md" tabindex="0">
+    <a id="minibutton-home" class="btn btn-sm" href="{{page_route}}">Home</a>
   </div>
 
-  <div
-    class="TableObject-item TableObject-item--primary px-2"
-    {{^search}}style="visibility:hidden"{{/search}}
-  >
+  <div class="TableObject-item TableObject-item--primary px-2"
+    {{^search}}style="visibility:hidden"{{/search}} >
     {{>searchbar}}
   </div>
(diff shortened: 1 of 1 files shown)


bartkamphorst:
Thanks. Could I ask you to try this PR on Chrome and see how it behaves?

dometto:
I just tested on Chrome, and sadly there's a problem: it treats *both* the new `div` with tabindex elements *and* the button as distinct tab-able items. So this change decreases user experience on Chrome -- you have to double tab to move to the next button, and to the average user it's probably not clear why pressing tab doesn't move to the next button on the first try. :-(

bartkamphorst:
Replaced by #2003 .

PR #2000, opened 2023-08-28
Handle 409 gracefully

dometto:
A `DuplicatePageError` is thrown when attempting to change the file extension of an exiting page to that of another already exististing page. That is, when: 

* A.md exists
* A.txt exists
* And we attempt to rename A.txt to A.md in the editor

This PR handles the resulting exception gracefully and returns a `409` in that case. It also ensures the error message is shown in the dialog that is triggered when pressing 'Save'

### DIFF
--- a/lib/gollum/public/gollum/javascript/gollum.js.erb
+++ b/lib/gollum/public/gollum/javascript/gollum.js.erb
@@ -560,6 +560,9 @@ $(document).ready(function() {
           if (data.status == 412) {
             $('#gollum-editor-submit').attr('disabled', false)
             alert('Someone else has modified this page while you were editing it. Please store your version on disk outside of the browser, reload this page and reapply your modifications.');
+          } else if (data.status == 409) {
+            console.log(errorThrown)
+            alert('Error updating page: ' + textStatus + ' ' + errorThrown);
           } else {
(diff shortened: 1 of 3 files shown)


lib/gollum/public/gollum/javascript/gollum.js.erb
```diff
@@ -561,7 +561,7 @@ $(document).ready(function() {
             $('#gollum-editor-submit').attr('disabled', false)
             alert('Someone else has modified this page while you were editing it. Please store your version on disk outside of the browser, reload this page and reapply your modifications.');
```
dometto, line 562: We could also move this error message to `app.rb`, and get rid of the `if...else` structure here (just showing a generic alert that displays the server-generated response in all cases).

bartkamphorst: Fair enough. As long as the submit button is still disabled when one of these errors show up.

bartkamphorst (approved):
PR looks good to me.

lib/gollum/public/gollum/javascript/gollum.js.erb
dometto: Ah, I didn't notice that's unique behavior for the 412 case. I'll just leave it like this then!

PR #2003, opened 2023-09-06
Change links to buttons and use onclick to navigate. Improves keyboard navigation and accessibility.

bartkamphorst:
This PR is an alternative to #1997 . It works by replacing all button-styled `<a>`'s by actual buttons, and using `onclick` to navigate. 

Screenshot 2023-09-06 at 19 32 53


bartkamphorst:
Follow-up question: 

Do we think the Edit/Preview tabs should be keyboard tabbable as well?

This is the situation now:
Screenshot 2023-09-06 at 19 35 39

This is the situation when the `<a>` are changed here into buttons.
Screenshot 2023-09-06 at 19 35 30

It's hard to tell, but changing the `<a>`s into `<button>`s does not change the general appearance of the tabs.

dometto:
> Do we think the Edit/Preview tabs should be keyboard tabbable as well?

I'm perfectly happy to have that be the case!

dometto:
> Do we think the Edit/Preview tabs should be keyboard tabbable as well?

In fact I just noticed that it already behaves this way in Chrome.

dometto (approved):

bartkamphorst:
> If there's no reason to be afraid that this won't work e.g. on mobile I'm all for it!

From what I can tell by using FF `Responsive Design Mode`, it works just fine on mobile.

@benjaminwil Any suggestions for this PR or is it okay to be merged?

dometto:
I just noticed that on the editor view, it's difficult to tab down to the Footer tab below the editor because Ace traps the focus. It looks like ACE has found a workaround for this in recent versions:

benjaminwil (approved):
Thank you! This looks great.

As for updating ACE to improve the tab-trapping situation, perhaps that's best done separately? (I can imagine it opening other cans of worms, is all.)

dometto:
> As for updating ACE to improve the tab-trapping situation, perhaps that's best done separately? (I can imagine it opening other cans of worms, is all.)

Very fair! @bartkamphorst don't let me hold you up anymore :) (I see there's a merge conflict at the moment, though. :-()

dometto:
JRuby test failing due to https://github.com/ruby/psych/issues/655

Issue #2007, opened 2023-09-24
Running tests fails if git's `init.defaultbranch` is not set to `master`


purp:
✅  **Please read the [CONTRIBUTION GUIDELINES](https://github.com/gollum/gollum/blob/master/CONTRIBUTING.md) before submitting!**

**Describe the bug**
12 tests in `test_page_view.rb` fail with a `Rugged::ReferenceError: reference 'refs/heads/master' not found` error if you've set `init.defaultbranch` to anything other than `master`. 

This seems to be because they're cloning `test/examples/empty.git` which has no branches in its `refs/heads` which causes clone to use `init.defaultbranch` to create the first branch in the cloned test directory, and the tests somehow require that branch to be named `master`. Workaround is, of course, pull that out of your config or change the value to `master`.

**To Reproduce**
```sh
> git config --global --add init.defaultbranch sticky-toffee-pudding
> bundle exec rake
```
... and see many errors.

**Expected behavior**
Didn't expect the default branch choice to break tests.

**Environment Info**

```sh
❯ bin/gollum --versions
Gollum 5.3.0
Running on: arm64-darwin22 with Ruby version 3.1.4
Using:
rugged 1.7.1
gollum-rugged_adapter 3.0
gollum-lib 6.0
Markdown rendering gem: kramdown
Other rendering gems:
none
```

purp:
Had a look at several possible ways to fix and have no idea what's the right one. Workaround's annoying but good enough for what I came to do.

dometto:
Thanks for the report. This is actually an issue I ran into when trying to change the branch name for the gollum repository to `main`, it's helpful to have this issue as a reminder.

benjaminwil:
I've also run into this issue. To work around it, I set my local git config for the Gollum repo(s):

```
cd my-gollum-project
git config --local init.defaultbranch=master
```

But I think this isn't really a totally foolproof way of dealing with the issue. And I don't think it would be right to commit any git config to the repository.

That said, we could at least add that in the `CONTRIBUTING` instructions to the "Set up your development environment" section. What do you think?

dometto:
> That said, we could at least add that in the CONTRIBUTING instructions to the "Set up your development environment" section. What do you think?

That seems like a good temporary solution! Anyone have a moment to add this to CONTRIBUTING?

benjaminwil:
I’d be happy to do it in the next couple of days. I will also add a couple of sentences context about why this might be necessary in 2023.

PR #2009, opened 2023-09-28
Add to development environment setup instructions

benjaminwil:
In 2023, not all git users have `init.defaultbranch` set to `master` anymore, which causes the Gollum test suite to fail. We can work around this with the instructions in the commit body.

This relates to #2007.


dometto:
Thank you, this is really clear and helpful. You could add a line saying it's on our todo list to move to `main`?

benjaminwil:
@dometto Done. Thank you. 🙏

PR #2010, opened 2023-10-16
Improve and localize the "edit" view partial

benjaminwil:
This small change started to localize the "edit" view partial.

While I was looking at this partial I realized that a) it was indented in an unconventional way and b) its buttons had `href` attributes, which is not something that `<button>`s have in valid HTML documents.

### DIFF
--- a/lib/gollum/locales/en.yml
+++ b/lib/gollum/locales/en.yml
@@ -16,6 +16,10 @@ en:
     comparing_versions_of: Comparing versions of
     comparing_from: "Comparing %{before} to %{after}"
     revert: Revert Changes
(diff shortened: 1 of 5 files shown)


benjaminwil:
Not sure what the JRuby CI failures mean. Will look into it.

benjaminwil:
The CI failure seems related to https://github.com/ruby/psych/issues/655 but I'm unsure what in our CI configuration makes it unable to use v5.1.1.1 which should resolve the issue.

dometto (approved):
Looks good! And thanks for adding more capybara tests!

PR #2021, opened 2023-12-21
Add checksum for gollum.war to release

dometto:


dometto:
On reflection, adding a checksum *file* does not seem very helpful. :) Checksums are supposed to give some protection with regard to man-in-the-middle attacks, but that function is not performed if the checksum must itself be downloaded by the user. We could, however, try to add the checksum to the text body of the release?

I guess the most optimal solution would be to actually GPG sign a (tarball of) the `.war`. But I don't see any way of doing that in an Actions workflow. If anyone has any idea what is considered best practice here, I'd be much obliged!

dometto:
This is too complicated and arguably of little worth. Better to add the signature to the release manually for now.

PR #2028, opened 2024-01-23
Update gems

dometto:
Updating dependencies. The change from http response 302 to 303 by sinatra seems to be because 303 is the preferred way of redirecting after a post request (see e.g. [here](https://serverfault.com/questions/391181/examples-of-302-vs-303)).


gollum.gemspec
```diff
@@ -24,15 +24,16 @@ Gem::Specification.new do |s|
   s.add_dependency 'gollum-lib', '~> 6.0'
   s.add_dependency 'kramdown', '~> 2.3'
   s.add_dependency 'kramdown-parser-gfm', '~> 1.1.0'
-  s.add_dependency 'sinatra', '~> 3.0'
-  s.add_dependency 'sinatra-contrib', '~> 3.0'
+  s.add_dependency 'rackup', '~> 2.1'
+  s.add_dependency 'sinatra', '~> 4.0'
+  s.add_dependency 'sinatra-contrib', '~> 4.0'
```
bartkamphorst, line 29: What is it that we use `sinatra-contrib` for?

bartkamphorst (approved):

gollum.gemspec
dometto: For the `namespace` functionality: https://github.com/sinatra/sinatra/blob/main/sinatra-contrib/lib/sinatra/namespace.rb

Issue #2029, opened 2024-02-12
Move from RACK_ENV to APP_ENV


svoop:
Gollum is using `RACK_ENV` to set the environment (production, development) similar to `RAILS_ENV`. While this is quite a common pattern, `RACK_ENV` was [never intended for this](https://github.com/rack/rack/issues/1546).

Given other frameworks move towards `APP_ENV` instead (even Rails 7 honors this), how about making the change on Gollum as well?

A quick grep shows only two places where this is used:

* [x] [bin/gollum](https://github.com/gollum/gollum/blob/db4adc231c26d58bb7e42da45ffcd768990ba239/bin/gollum#L285C3-L285C33) –> read `APP_ENV`, fall back to `RACK_ENV` (with deprecation warning), fall back to `:production`
* [x] [test/helper.rb](https://github.com/gollum/gollum/blob/db4adc231c26d58bb7e42da45ffcd768990ba239/test/helper.rb#L31) -> given the above just use `APP_ENV` instead
* [x] Maybe also document `APP_ENV` in the README.

What do you think about this? Am I missing something and if not, would you consider such a PR for merge?

benjaminwil:
I think this change would make sense. I don't see anything stopping us from deprecating using `RACK_ENV` and removing it in the next major version or something like that.

Thanks for pointing out how we could make Gollum more conventional.

dometto:
I agree this would be a good idea. Let's add a deprecation notice!

dometto:
If I see correctly all the todos in the issues are completed now. Thanks again @svoop!

PR #2031, opened 2024-02-19
Deprecate RACK_ENV in favour of APP_ENV

svoop:
Say goodbye for the often misused `RACK_ENV` in favour of `APP_ENV` (like Sinatra does). A deprecation warning is printed if only `RACK_ENV` is set and Golllum has not yet reached version 6.

Related issue: #2029

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -205,7 +205,7 @@ MSG
     puts results.empty? ? 'none' : results
     exit 0
   end
-  
+
   opts.separator ''
   opts.separator '  Development:'
 
@@ -282,7 +282,12 @@ if options[:irb]
   end
 else
   require 'gollum/app'
-  Precious::App.set(:environment, ENV.fetch('RACK_ENV', :production).to_sym)
+  # TODO: Remove RACK_ENV fallback once gollum-7 is released (4 lines)
+  if ENV['RACK_ENV'] && !ENV['APP_ENV']
+    warn "[DEPRECATION] RACK_ENV will be ignored as of gollum-6.0.0, please use APP_ENV instead."
+    ENV['APP_ENV'] = ENV['RACK_ENV']
+  end
+  Precious::App.set(:environment, ENV.fetch('APP_ENV', :production).to_sym)
   Precious::App.set(:gollum_path, gollum_path)
   Precious::App.set(:wiki_options, wiki_options)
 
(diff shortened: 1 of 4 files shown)


bin/gollum
```diff
@@ -282,7 +282,12 @@ if options[:irb]
   end
 else
   require 'gollum/app'
-  Precious::App.set(:environment, ENV.fetch('RACK_ENV', :production).to_sym)
+  # TODO: Remove RACK_ENV fallback once gollum-6.0-0 is released (4 lines)
+  if Gollum::VERSION < "6.0.0" && ENV['RACK_ENV'] && !ENV['APP_ENV']
+    warn "[DEPRECATION] RACK_ENV will be ignored as of gollum-6.0.0, please use APP_ENV instead."
+    ENV['APP_ENV'] = ENV['RACK_ENV']
```
benjaminwil, line 288: I was surprised to learn that this string comparison works.

If Gollum was a project that had more pre-releases (i.e. `6.0.0.alpha`) I would reach for a more-expressive `Gem::Version` comparison:

```
Gem::Version.new(Gollum::VERSION) < Gem::Version.new("6.0.0")
```

But I manually tested that this does the right thing, and it does. So looks good to me.

benjaminwil (approved):
Thanks for this contribution.

I will defer to @dometto to approve and merge, as he has a clearer picture of what the 6.0.0 release looks like.

bin/gollum
svoop: Actually, I'm not really sure whether checking for the version there is a good idea at all. Might just as well without, provided someone keeps track of such post-release TODOs.

dometto: I think it could be alright to leave out the version check, if we open an issue to remove the deprecation notice. Even with the check present, we'll have to remember to remove the code af some point, so it doesn't make that much of a difference I think.

So I propose that we keep the deprecation notice (without the version check) until 7.0 and open an issue to remind ourselves to remove the relevant code in that release.

svoop: @dometto Sorry for the delay, this slipped off my desk somehow. I've removed the version check and just kept the `# TODO` which mentions the lines to be removed once gollum-7 is released.

dometto:
Thank you @svoop and sorry for the delay!

svoop:
Easy, thanks for the merge!

PR #2032, opened 2024-02-23
Docker semver supports

finagin:
Append docker semver support to using like:
```bash
docker pull gollumwiki/gollum
docker pull gollumwiki/gollum:latest
docker pull gollumwiki/gollum:5.3.2         # Concrete version
docker pull gollumwiki/gollum:5.3           # Concrete minor version with latest hotfixes
docker pull gollumwiki/gollum:5             # Concrete major version with latest minor updates and hotfixes
docker pull gollumwiki/gollum:db4adc2       # Concrete commit

docker pull ghcr.io/gollum/gollum:5.3       # Use GitHub Container Registry as alternative registry
```

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -13,21 +13,32 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - name: Set deploy name to master
-        if: github.ref == 'refs/heads/master'
-        run: |
-          echo "DEPLOY_NAME=gollumwiki/gollum:master" >> $GITHUB_ENV
-      - name: Set deploy name to release version
-        if: startsWith(github.ref, 'refs/tags/')
-        run: |
-          echo "DEPLOY_NAME=gollumwiki/gollum:${{ github.ref_name }}" >> $GITHUB_ENV
       - name: Check Out Repo
         uses: actions/checkout@v2
-      - name: Login
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            gollumwiki/gollum
+            ghcr.io/gollum/gollum
+          tags: |
+            type=semver,pattern={{version}}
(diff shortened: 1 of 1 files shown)


benjaminwil (commented):
This is a nice improvement. I don't know that Gollum uses the GitHub Container Registry at all right now, so we'll either want to set that up first or remove that part of the commit.

cc: @dometto since I think you'd be the person with the keys for the GHCR.

finagin:
> cc: @dometto since I think you'd be the person with the keys for the GHCR.

GitHub Actions have access to GitHub Container Registry

dometto:
Thanks for this PR!

We would have to check whether we could get free unlimited public downloads from ghcr. But also we'd have to find a way of letting users know that the container registry has changed from dockerhub to ghcr. I'm not against such a move but in the interest of time it might be easier to just stick with dockerhub for the moment. Or does only ghcr support semver?

benjaminwil:
This pull request _adds_ GHCR, but Dockerhub is still included here.

finagin:
@dometto, are u here?

dometto:
Sorry for the long wait. I believe pushing a container to ghcr is enough to setup the relevant Package, but we can just merge and see if action needs to be taken.

It looks like ghcr is free for public containers (this has either changed, or else it was not so clear on https://ghcr.io), so happy to add it here. Ready to merge as far as I'm concerned!

Thanks for the patience @finagin

dometto:
@finagin the ghcr and dockerhub packages were published successfully, but it appears the `master` tag is not being applied anymore to the latest image build: https://hub.docker.com/layers/gollumwiki/gollum/sha-4591bf7/images/sha256-01ade3a57181307edac1a0fa510a36e75f3a6322b1c3c4020665d5b562e2b414?context=explore

Any chance that can be added in addition to the semver labels?

finagin:
Hmm🤔 
I'll try to fix it this weekend

PR #2038, opened 2024-04-16
Update js dependencies

dometto:
In an effort not to create a million commits with newly compiled assets, I'm creating one PR in which are updated:

* KaTeX
* Mermaid
* ACE

With the exception of one commit containing the compiled assets, I'll split all commits so the history stays orderly,


benjaminwil (approved):
Wonderful ✨

PR #2040, opened 2024-04-16
Support for more languages: chinese

zoumingzhe:
This patch can support more languages, such as chinese.

```ruby
  ::I18n.default_locale = :cn
```

You just need to add a new line in the configuration to achieve the following effect.

image

I have added more content in the locales(gollum/lib/gollum/locales) to achieve better results.

I hope more people can contribute their mother tongue.

### DIFF
--- a/lib/gollum/templates/search.mustache
+++ b/lib/gollum/templates/search.mustache
@@ -3,10 +3,7 @@
     {{>navbar}}
 
     <h1 class="header-title text-center text-md-left py-4">
-      <span class="f1-light color-fg-subtle">
-        {{t.search_results_for}}
-      </span>
-      {{name}}
+      {{t.search_results_for}}
     </h1>
   </div>
 
@@ -41,7 +38,7 @@
 
 {{#no_results}}
   <p id="no-results">
-    {{t.no_results}} <strong>{{query}}</strong>.
+    {{t.no_results}}
   </p>
 {{/no_results}}
 
--- a/LATEST_CHANGES.md
+++ b/LATEST_CHANGES.md
@@ -11,6 +11,7 @@
 * Add support for downloading page sources with ?raw (@tstein).
 * Add openssh client to docker images for ssh: repo support. (@jagerkin).
 * Add support for mathematical typesetting using KaTeX (@dometto). Users can now choose between MathJax and KaTeX with the --math flag.
+* Add support for more languages (such as: Chinese).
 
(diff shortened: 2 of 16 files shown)


zoumingzhe:
[Issues](https://github.com/gollum/gollum/issues/2041#issue-2246451549) Some pages cannot support multiple languages

benjaminwil:
Thank you! I will review this more closely later.

I responded to your issue #2041 with a potential solution to the translating the remaining text :+1:

zoumingzhe:
> I responded to your issue #2041 with a potential solution to the translating the remaining text 👍

Thanks for the reply, the localization of the "Create New Page" and "Rename to" pages is not urgent because I have already completed the localization of the important pages.

I wonder if we need to add a language list menu so that users can switch the display language instead of forcing it to be specified in the configuration.

This might make a lot of sense. First, we can preview the localization effect more quickly. Secondly, services based on gollum wiki can serve different regions.

benjaminwil:
> I wonder if we need to add a language list menu so that users can switch the display language instead of forcing it to be specified in the configuration. 

This sounds great, but we could work on this and merge it separately. 👍 

> the localization of the "Create New Page" and "Rename to" pages is not urgent

I agree. I will review this pull request as it is now, then.

benjaminwil (approved):
I manually tested this by pulling down the branch and changing my locale to CN.

Thank you for this contribution!

lib/gollum/templates/search.mustache
```diff
@@ -41,7 +38,7 @@
 
 {{#no_results}}
   <p id="no-results">
-    {{t.no_results}} <strong>{{query}}</strong>.
+    {{t.no_results}}
```
benjaminwil, line 41: Removing the query from here has caused one test to fail.

What is the reason for removing `{{query}}` here? It can be helpful for the user to see what their previous search query was—in case they made a mistake.

zoumingzhe: The query is already in 'search_results_for', and it is unnecessary in 'no_results'.

image

benjaminwil: I see now. Thanks for explaining!

benjaminwil:
@dometto I would be happy to put together a release version for this change. Would love to connect with you about release workflow first, though.

benjaminwil:
Just wanted to note that this was included in the 6.0 release.

Issue #2042, opened 2024-04-16
Unnecessary content in the overview page


zoumingzhe:
1. Click on the overview button on the home page.
2. You will see the "Home" on this page. 

![image](https://github.com/gollum/gollum/assets/23275497/570e6090-4ece-4018-afc6-3251e8d1d1a9)

zoumingzhe:
This path is only useful in multi-level directories. If your git repository doesn't have any subdirectories, "Home" will confuse you.

Maybe we can modify it this way (not displaying anything).

![image](https://github.com/gollum/gollum/assets/23275497/5574a206-fe2d-4f70-b281-308a64cd7749)

zoumingzhe:
The patch has been merged. I will close it.

benjaminwil:
Thanks again!

PR #2043, opened 2024-04-16
Remove unnecessary content in overview page

zoumingzhe:
Link https://github.com/gollum/gollum/issues/2042


benjaminwil (approved):
Makes sense to me!

benjaminwil:
Thanks for the contribution.

It looks like this change has caused some tests to fail, because we previously expected the text "Home" to appear and now that does not happen.

Are you able to push a commit to remove the offending tests?

zoumingzhe:
> Are you able to push a commit to remove the offending tests?

I'm going to do this.

PR #2047, opened 2024-04-27
Re-read base_path after loading config file

leon-barrett:
Bug fix: Setting base_path in the config file didn't have any effect.

The config file was being loaded, but it wasn't being used to update the wiki_options local variable, so the config file wasn't affecting the base_path.

This change updates the wiki_options local variable after the config file is read. Now, the base_path can be set using the config file.

### DIFF
--- a/LATEST_CHANGES.md
+++ b/LATEST_CHANGES.md
@@ -16,3 +16,4 @@
 ## Fixes & Improvements
 
 * Fix (Docker image): add git configuration for `/wiki` as safe directory. #2006
+* Fix: use `base_path` as set in config file.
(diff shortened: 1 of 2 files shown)


benjaminwil:
Thanks for this contribution.

For posterity, I was able to manually reproduce this issue locally and verify that this resolves the issue.

Steps:

1. Create a Gollum configuration file containing:
   ```ruby
   # ~/my-wiki/config.rb
   
   Precious::App.set :wiki_options, {base_path: "test"}
   ```
2. Start the Gollum process (at the `master` ref) pointing to this file:
   ```
   bundle exec gollum ~/my-wiki --config ~/my-wiki/config.rb
   ```
3. Note that visiting `localhost:4567`, the base path does not include `/test`.
4. Pull and switch to the branch for this pull request and restart the Gollum process the same way.
5. Note that visiting `localhost:4567` injects the correct base path `/test`.

benjaminwil (approved):

PR #2058, opened 2024-06-30
fix: MapGollum should handle superfluous slashes in base_path

### LINKED ISSUE #2050: redirect error when use base_path option
**Describe the bug**

When set the 'base_path: "/wiki",' in config.rb, I cannot view page.

**To Reproduce**

1. view http://192.168.0.110:4567/wiki/
2. redirect to http://wiki:4567/

**Screenshots**

image
image

**Environment Info**

docker image tag 6.0.0

dometto:
Resolves #2050.

As reported in #2050 setting `base_path: /wiki` doesn't work, while `base_path: wiki` does work. It appears this is because the `MapGollum` class is setting the base route to `/#{base_path}`, so when passing in `/wiki`, the actual route being served is `//wiki`, not `/wiki`. By contrast, setting `--base-path /wiki` on the commandline works because the leading slash is being stripped [here[(https://github.com/gollum/gollum/blob/master/bin/gollum#L76C1-L77C1).

This PR attempts to solve the issue by handling extra slashes in the `MapGollum` class.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 3.1
+          ruby-version: 3.2.1
           bundler-cache: true
       - name: Install Chromedriver
         uses: nanasess/setup-chromedriver@v2
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ['2.7', '3.1', '3.2.1']
+        ruby: ['3.1.4', '3.2.1', '3.3']
     steps:
       - name: Install required system dependencies
         run: |
@@ -55,6 +55,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          rubygems: 3.5.14
           bundler-cache: true
       - name: Run tests
         run: bundle exec rake
@@ -79,7 +80,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1'
+          ruby-version: '3.2.1'
           bundler-cache: true
       - name: Run tests
         run: bundle exec rake test:capybara
--- a/lib/gollum/app.rb
+++ b/lib/gollum/app.rb
@@ -55,17 +55,19 @@ module Precious
 
   # For use with the --base-path option.
   class MapGollum
+  include Precious::Helpers
     def initialize(base_path)
+      base_route = "/#{remove_leading_and_trailing_slashes(base_path)}"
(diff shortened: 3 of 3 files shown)


dometto:
I've had to drop CI support for Ruby 3.1 to get the tests passing.

Apparently there is an issue with `psych` (again) that makes it incompatible with older versions of `rubygems`, so that you need either ruby >= 3.2, or to update rubygems manually on older rubies.

However, there's no room to run `gem update --system` in CI, because setup of Ruby and `bundle install` are both run as part of the `setup-ruby` GitHub Action. Before we run that action, there's no rubygems to update, and after that is too late (the error is thrown during `bundle install`).

@benjaminwil @bartkamphorst ok to drop tests for ruby < 3.2?

benjaminwil:
I am not opposed to dropping the tests for < 3.2 but would note that Ruby 3.1 will not be end-of-lifed until next year.

I quickly looked to see if there were any `setup-ruby` users with this issue, and it looks like there are some. And it looks like there's a solution, but I haven't tested.

The `setup-ruby` GitHub Action has been updated to let us [specify a RubyGems version explicitly](https://github.com/ruby/setup-ruby/pull/271). From the source code in `action.yml`:

> `rubygems`:
>
> The version of RubyGems to use. Either 'default', 'latest', or a version number (e.g., 3.3.5).  For 'default', no action is taken and the version of RubyGems that comes with Ruby by default is used. For 'latest', `gem update --system` is run to update to the latest RubyGems version. Similarly, if a version number is given, `gem update --system <version>` is run to update to that version of RubyGems.
>
> Defaults to 'default'.

From the comments on that pull request, an example configuration:

```
jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, macos-latest]
        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1]
    runs-on: ${{ matrix.os }}
    steps:
    - uses: actions/checkout@v2
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: ${{ matrix.ruby }}
        rubygems: MINIMUM_RUBYGEMS_VERSION_WHICH_SUPPORTS_PSYCH4
        bundler-cache: true
```

After reading some of the discussion around this, I would be hesitant to just use `latest` in case that breaks CI runs in the future and is hard to understand. But maybe you know, @dometto, what a good version lock could be now?

dometto:
Thanks @benjaminwil, I was looking for the possibility of setting a version in `setup-ruby` but didn't see it in the readme! That's a much nicer solution.

If I understand correctly, the `rubygems: ` option [noops](https://github.com/ruby/setup-ruby/pull/271/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R137) if the specified version is lower than the default version for the specified ruby version. So we could set `rubygems: 3.0.3` (which should work with psych), and for newer rubies, it should still use the default rubygems for that version.

dometto:
Took some fiddling to find the right `rubygems` versions because `3.3.26` works locally but throws the psych error in CI 🤷 

Found a working combination, can always revisit if it leads to issues later on.

PR #2060, opened 2024-07-07
fix #2059: Rack::Server class has moved to the rackup gem

### LINKED ISSUE #2059: Rack::Server error when using `--config config.rb`
**Describe the bug**
If I try to run `gollum` with the `--config config.rb` option, I get the following error

    $ gollum --config config.rb wikidata/
    /home/vagrant/.local/share/gem/ruby/3.1.0/gems/gollum-6.0.0/bin/gollum:299:in `<top (required)>': uninitialized constant Rack::Server (NameError)

        Rack::Server.new(:app => Precious::MapGollum.new(base_path), :Port => options[:port], :Host => options[:bind]).start
            ^^^^^^^^
    Did you mean?  TCPServer
            from /home/vagrant/.local/share/gem/ruby/3.1.0/bin/gollum:25:in `load'
            from /home/vagrant/.local/share/gem/ruby/3.1.0/bin/gollum:25:in `<main>'

**To Reproduce**
1. Installing gollum in a test VM (debian 12)

        sudo apt install ruby-rubygems git ruby-dev build-essential cmake pkgconf libssl-dev
        gem install --user-install gollum

2. Creating the git repository

        mkdir -vp my-gollum-wiki/wikidata
        cd my-gollum-wiki/wikidata/
        git init
        touch Home.md
        git add Home.md && git commit -m "initial commit"

3. Creating a config file with a minimal content

        $ cat config.rb
        wiki_options = {
        base_path: "/wiki"
        }
        Precious::App.set(:wiki_options, wiki_options)

   See "Description" above ↑

**Expected behavior**

if I understand the <https://github.com/gollum/gollum/wiki/Sample-config.rb> page correctly, if we define the configuration in `config.rb` and not in `config.ru`, we don't need to have a rack server.

Did I miss something ?

**Environment Info**

    $ gollum --versions
    Gollum 6.0.0
    Running on: x86_64-linux-gnu with Ruby version 3.1.2
    Using:
    rugged 1.7.2
    gollum-rugged_adapter 3.0
    gollum-lib 6.0
    Other rendering gems:
    none

    $ lsb_release -d
    No LSB modules are available.
    Description:    Debian GNU/Linux 12 (bookworm)

dometto:
Resolves #2059.

In `/bin/gollum` we try to instantiate the `Rack::Server` class, but it turns out [that class has been moved](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#binrackup-rackserver-rackhandlerand--racklobster-were-moved-to-a-separate-gem) to a different gem (`rackup`, which we already depend on).

### DIFF
--- a/bin/gollum
+++ b/bin/gollum
@@ -293,9 +293,9 @@ else
   if base_path.nil?
     Precious::App.run!(options)
   else
-    require 'rack'
+    require 'rackup'
 
-    # Rack::Handler does not work with Ctrl + C. Use Rack::Server instead.
-    Rack::Server.new(:app => Precious::MapGollum.new(base_path), :Port => options[:port], :Host => options[:bind]).start
+    # Rackup::Handler does not work with Ctrl + C. Use Rackup::Server instead.
+    Rackup::Server.new(:app => Precious::MapGollum.new(base_path), :Port => options[:port], :Host => options[:bind]).start
   end
 end
--- a/gollum.gemspec
+++ b/gollum.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'gollum-lib', '~> 6.0'
   s.add_dependency 'kramdown', '~> 2.3'
   s.add_dependency 'kramdown-parser-gfm', '~> 1.1.0'
+  s.add_dependency 'rack', '>= 3.0'
   s.add_dependency 'rackup', '~> 2.1'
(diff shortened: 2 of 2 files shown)


benjaminwil:
This change looks good to me 👍 But I wonder if, to be safe, we should also ensure the Gemspec required Rack > v3, so we don't end up with some users getting errors due to having Rack v2 installed.

PR #2085, opened 2024-12-18
Resolve JRuby CI failures

benjaminwil:
This pull request collects some changes to make JRuby CI runs pass. (I may throw in some other CI quality-of-life improvements.)

## jar-dependencies version constraint

There is currently an issue in JRuby 9.4.9.0 because of the open-ended version requirement of `jar-depedencies` in the `psych` gem. The issue results in Gollum and its test suite being completely broken with this `Gem::LoadError`:

```
Gem::LoadError: You have already activated jar-dependencies 0.4.1,
but your Gemfile requires jar-dependencies 0.5.1. Since 
jar-dependencies is a default gem, you can either remove your
dependency on it or try updating to a newer version of bundler
that supports jar-dependencies as a default gem.
```

The issue is discussed here: https://github.com/jruby/jruby/issues/8488

And it will be resolved when 9.4.10.0 is released.

If your Gollum project is affected by this issue, you can add a `"jar-dependencies", "< 0.5"` version constraint to your project's Gemfile.

## Weird test failure

A new, weird test failure appeared for JRuby CI runs only. Changing how the test is written was the easiest fix.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,6 +11,7 @@ on:
       - '**.md'
       - '**.txt'
       - 'contrib/**'
+
 jobs:
   jruby_build:
     name: JRuby (${{ matrix.ruby }})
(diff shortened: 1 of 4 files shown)


benjaminwil:
One new test failure. I can't reproduce it locally yet. 😞

benjaminwil:
Not adding anything to the latest changes document here, because none of this effects application code.

Issue #2084, opened 2024-12-17
Official docker image for 6.0+ not working due to /wiki directory access error


Hooverdan96:
**Describe the bug**
Attempting to run official docker image with tags >=6.0 as well as `latest` or `master` results in "fatal" error during startup.

**To Reproduce**
Steps to reproduce the behavior, tested with `latest` (currently 6.0.1), `master` and `6.0`
1. start docker container within empty directory (no `.git` file present) using:
```
docker run --rm -p 4568:4567 -v $(pwd):/wiki gollumwiki/gollum:6.0
```
2. See error during execution of docker command
```
Unable to find image 'gollumwiki/gollum:6.0' locally
6.0: Pulling from gollumwiki/gollum
Digest: sha256:dc7bd2a46c83a189fb4bc3c1ab1fbc1ba28254ed4f019935c26527ea574668c8
Status: Downloaded newer image for gollumwiki/gollum:6.0
fatal: Cannot access work tree '/wiki': Permission denied
[2024-12-17 22:13:04] INFO  WEBrick 1.8.1
[2024-12-17 22:13:04] INFO  ruby 3.1.6 (2024-05-29) [x86_64-linux-musl]
== Sinatra (v4.0.0) has taken the stage on 4567 for production with backup from WEBrick
[2024-12-17 22:13:04] INFO  WEBrick::HTTPServer#start: pid=1 port=4567
```

3. navigate to `host-ip address`:4568 and observe web page error:
<img width="787" alt="image" src="https://github.com/user-attachments/assets/b88e7cdb-1615-46ba-9c76-4b1e2f111da7" />

**Expected behavior**
When using e.g. version `5.3.3`  in the same above command, the gollum page comes up, no errors. The message log in that version indicates that the git initialization is successful:
```
Initialized empty Git repository in /wiki/.git/
```
while with 6.0+ it seems to run into the authorization problem.

**Environment Info**
see above
Always include the output of the `gollum --versions` command (NB: `versions` in plural).

dometto:
Thanks for the bug report. Unfortunately I can't reproduce this. Might the issue be related to your Docker setup locally? I'm thinking the volume mount of the directory might not be working correctly for some reason. What happens if you try the following:

1. Run gollum while giving the docker container a specific name: `docker run --name gollum-test --rm -p 4568:4567 -v $(pwd):/wiki gollumwiki/gollum:6.0`
2. While the container is running, from a different terminal, run `docker exec gollum-test touch /wiki/test`

[...]

PR #2086, opened 2024-12-20
Introduce directory permission check to docker-run.sh script (#2084)

Hooverdan96:
1. Please clearly explain the purpose of your PR. Let us know if it introduces a new feature, an enhancement, or fixes a bug.
Enhancement Proposal:
For docker images starting with 6.0 a non-root user needs the correct permissions to the mapped host volume:
`-v /path/to/git/directory/on/host:wiki`. If these are not set correctly, the docker container will continue to run, but the web page will not work.
This introduces a short check at the beginning of docker-run.sh to check for existence and writeability of the volume. If true, continue with the script, if false, exit. For both conditions issue a message describing success or failure.

2. Please link any issues that your PR addresses.
Inspired by #2084

3. If your PR changes any of the CSS or JavaScript under `lib/gollum/public`, don't forget to update the static asssets as explained [here](https://github.com/gollum/gollum/blob/master/CONTRIBUTING.md#updating-static-assets).
No changes to CSS/JavaScript

Added to [LATEST_CHANGES.md](https://github.com/gollum/gollum/blob/master/LATEST_CHANGES.md)

### DIFF
--- a/LATEST_CHANGES.md
+++ b/LATEST_CHANGES.md
@@ -3,3 +3,4 @@
 ## Fixes & Improvements
 * Fixed a typo in the deprecation message for `--mathjax` (@DavidForster)
 * Update Mermaid to 11.4
+* Add wiki permission check to docker run script
--- a/docker-run.sh
+++ b/docker-run.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# Check if /wiki directory exists and is writable
+if [ ! -d "/wiki" ] || [ ! -w "/wiki" ]; then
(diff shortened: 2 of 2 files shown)


dometto:
Thanks for helping out with this!

I'm wondering if actually exiting is too drastic a measure when the directory is not writable, since Gollum also has a read-only mode. Just throwing the error message will be a huge help to users (and to us, if people do file bug reports).

Hooverdan96:
yes, that makes sense. I will adjust it so it will just throw the message and then continue.

dometto:
Thanks @Hooverdan96!

benjaminwil:
Awesome! Thank you @Hooverdan96 💯

PR #2089, opened 2024-12-22
Update Ruby in Dockerfile

dometto:
Update the version of ruby used in the docker image from `3.1` to `3.3`. We are already testing Gollum against 3.3, so this should be fine, and potentially improve performance a bit.


benjaminwil:
Thank you!

PR #2092, opened 2024-12-29
Improve the write test in docker-run.sh to ensure actual write permissions on the mounted volume.

bartkamphorst:
This PR solves the following issue, namely that checking for write permissions on the running docker container with `[ ! -w /wiki ]` would not be accurate when write permissions on the local volume were not granted to the docker user. 

The scenario is this: a local directory, say `wikidir` has owner `user` and permissions `600` and is mounted in docker. Docker is told that the mounted volume directory is owned by `www-dir` with permissions `600`. However, user `www-dir` is not allowed to actually write to the volume due to the local permissions on the host. The `-w` check only looks at permissions, which _seem_ to allow writing to the mounted volume when it is in fact not allowed, while the check in this PR actually attempts a write action.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -28,7 +28,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Cache Docker layers
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: /tmp/.buildx-cache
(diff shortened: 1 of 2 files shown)


dometto:
@benjaminwil would you have a moment for a quick look at this PR? @bartkamphorst worked on it together so would appreciate another perspective, but if you're too busy ATM we can just merge!

PR #2094, opened 2025-01-07
Update sassc-embedded

ntkme:
This PR updates `sassc-embedded` to latest version. Also:

- Remove `sassc` to speed up `bundle install` for development, as latest `sassc-embedded` already bundles necessary code from `sassc`.
- Add options to silence `import` and `global-builtin` deprecations.

### DIFF
--- a/lib/gollum/assets.rb
+++ b/lib/gollum/assets.rb
@@ -19,6 +19,19 @@ module Precious
       env.js_compressor  = ::Precious::Assets::JS_COMPRESSOR if defined?(::Precious::Assets::JS_COMPRESSOR)
       env.css_compressor = :sassc
 
+      options = {
+        sass_config: {
+          quiet_deps: true,
+          silence_deprecations: [
+            'global-builtin',
+            'import'
+          ]
+        }
+      }
+
+      env.register_transformer 'text/sass', 'text/css', Sprockets::SasscProcessor.new(options)
+      env.register_transformer 'text/scss', 'text/css', Sprockets::ScsscProcessor.new(options)
+
       env.context_class.class_eval do
         def base_url
           self.class.class_variable_get(:@@base_url)
--- a/Gemfile
+++ b/Gemfile
@@ -19,10 +19,7 @@ group :test do
 end
 
 group :development do
-  unless RUBY_PLATFORM.match(/linux-musl$/)
-    gem 'sassc',  '~> 2.4'
-    gem 'sassc-embedded', '~> 1.54'
-  end
+  gem 'sassc-embedded', '~> 1.80'
 end
 
 gemspec


lib/gollum/assets.rb
```diff
@@ -19,6 +19,19 @@ def self.sprockets(dir = File.dirname(File.expand_path(__FILE__)))
       env.js_compressor  = ::Precious::Assets::JS_COMPRESSOR if defined?(::Precious::Assets::JS_COMPRESSOR)
       env.css_compressor = :sassc
 
+      options = {
+        sass_config: {
+          quiet_deps: true,
+          silence_deprecations: [
```
bartkamphorst, line 25: What are these deprecation messages saying? Shouldn't we address them rather then ignoring the messages?

bartkamphorst (commented):
Thanks for contributing @ntkme ! 

Great to see that 'sassc-embedded' has been maintained and that `linux-musl` is now supported as well.

In general I'd say the PR is looking good, I just have a question about the deprecation messages, see the comment around l.25.

ntkme:
@bartkamphorst I fixed all deprecation warnings. Can you please take a look?

Note that `quiet_deps: true` is to silence warnings from dependencies that we cannot change, e.g. `@primer/css`.

lib/gollum/assets.rb
```diff
@@ -44,3 +40,16 @@ def base_url
     end
   end
 end
+
+module Sprockets
```
bartkamphorst, line 44: I don't think I understand the reasoning behind this monkeypatch. Can you explain why we need this?

ntkme: It's a bit complicated.

First of all, `//= require` is highly discouraged for modern sass because `//= require` embed the compiled child file directly in the parent. This has two problems:
1. It cause child files to be processed more than once
2. It completely breaks source map location
3. It slows down the I/O performance of sass backend

So I changed all `//= require` to `@use`.

The second problem is that `@use` cannot load `template.sass.erb`, because `@use` is directly processed by sass backend, which does not know how to process `erb`. From what I can see the only use case for `erb` is this `rocticon_css` function. Therefore, I replaced the `erb` with a custom sprocket `sassc` function, which is now directly invoked by sass backend.

ntkme: In other words, this is the preferred way of calling ruby function from sass, instead of using `erb`.

It effectively replaces this erb helper (the helper is also used by js code, so it's kept there): https://github.com/gollum/gollum/blob/476c7de094732115c553337eff487750724eeae9/lib/gollum/views/helpers.rb#L78-L83

ntkme: Reference: https://github.com/rails/sprockets/blob/v4.2.1/lib/sprockets/sassc_processor.rb#L89-L100

dometto: thanks for the explanation @ntkme! That makes sense.

We'd like to revisit our workflow for building gollum's SASS a little before merging this. At the moment we are asking people to add the precompiled and minimized CSS/JS to their PRs, but that is obviously suboptional for a few reasons. We would prefer to have a workflow for where only changes to the SASS/JS templates are directly contributed, and a build step is triggered to precompile from the new source.

Let's see if we get around to doing this this week. Or if you feel like it and have time, you're welcome to contribute a Github Actions file that achieves this, as well. :)

ntkme: @dometto Added a commit to automate that in github actions. You can see that it creates a commit authored by `github-actions[bot]` in this PR. Also updated some outdated actions that are broken.

.github/workflows/precompile-assets.yaml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          yq 'del(.. | select(has("mtime")).mtime)' lib/gollum/public/assets/.sprockets-manifest-*.json | tee /tmp/.sprockets-manifest.json
+      - name: Hash assets
+        id: hash-after
+        run: |
+          echo "hash=${{ hashFiles('lib/gollum/public/assets/**/*', '!lib/gollum/public/assets/.sprockets-manifest-*.json', '/tmp/.sprockets-manifest.json') }}" | tee -a "$GITHUB_OUTPUT"
+      - name: Commit
+        if: steps.hash-before.outputs.hash != steps.hash-after.outputs.hash
+        run: |
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
```
dometto, line 40: `41898282` is just a hardcoded number to make the email more unique, or is there something here that I'm missing? :)

lib/gollum/assets.rb
dometto: Thanks a lot for the contribution, looks great! Small question in the review.

.github/workflows/precompile-assets.yaml
```diff
@@ -0,0 +1,44 @@
+name: Precompile Assets
+on:
+  push:
+    branches:
+      - '**'
```
dometto, line 5: Wondering if it might be better to run this action only on pushes on `master`. @ntkme @bartkamphorst any thoughts?

.github/workflows/precompile-assets.yaml
ntkme: That is the user id of GitHub Actions bot user. By using this email the commit will show as authored by the bot user.

.github/workflows/precompile-assets.yaml
ntkme: Updated to `master` only and remove the pull_request trigger. - Technically it's possible to checkout `refs/pull/<id>/head` and commit to PR, but turns out `actions/checkout` will checkout PR as detached HEAD by default, so that will require some additional setup which might not be worth it.

Regarding master vs all branches, I think it does not matter that much, as the hash check is designed to only commit when assets actually change. - If we update a branch and then merge to master, it won't commit on master the second time as assets will not change at that time.

ntkme: I think we should run it both before and after, so that you will have test status on every single commit on master.  I added a trigger for doing that.

dometto: Hmm, I think `docker-deploy` will now run twice -- once called from the `precompile-assets` step, and once from the push to `master`. Do you see any way to avoid this?

Additionally, I think we want to also precompile assets in the `docker-test` workflow, so that when a PR changes assets, the tests are ran against the compiled changes.

Thanks so much for your help! This will make it a lot easier to maintain our JS/CSS dependencies.

ntkme: Honestly, GitHub hosted runners are free so I don’t see running twice before and after having any kind of problem. It will give the commit history a check mark on every commit which I think is actually better.

dometto: That's a fair point, although I was mostly thinking about double pushes to DockerHub. I'm not entirely sure how this works though; I guess maybe it won't re-push images with a digest that already exists on dockerhub?

We can also just merge as is, and adjust if necessary.

ntkme: Yes, docker digest will change given image content will be different. Also, docker tags are basically mutable pointers to digest.

dometto:
Seeing one test failure annoyingly: https://github.com/gollum/gollum/actions/runs/14223947849/job/40082135911?pr=2094

ntkme:
This also happens randomly on my local, seems to be a flaky test.

dometto:
🎆  thanks so much @ntkme!

gollum.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'terser', '~> 1.1'
 
   # = MANIFEST =
-  s.files = %w[
+  s.files = Dir['lib/gollum/public/assets/**/*'] + %w[
```
benjaminwil, line 50: 😌

PR #2108, opened 2025-04-08
Update docker-deploy.yml: don't trigger on push to master

dometto:
After merging https://github.com/gollum/gollum/pull/2094, the precompile-assets workflow will trigger docker-deploy. To avoid duplicate runs of this workflow, we can remove the push-to-master trigger here.

So at present the workflow logic looks as follows:

* Push to `master`: triggers `precompile-asset`, which triggers `docker-deploy`
* Push a new tag: triggers `docker-deploy`, but not `precompile-assets`.
* Push to a different branch: nothing happens. :)

Would appreciate a quick check if this makes sense!

### DIFF
--- a/.github/workflows/docker-deploy.yml
+++ b/.github/workflows/docker-deploy.yml
@@ -1,8 +1,9 @@
 name: Deploy docker
 on:
   push:
-    branches:
-      - master
+    # This workflow will be triggered by the `precompile-assets` workflow on pushes to `master`
+    # branches:
+    #   - master
     tags:
       - "*"
     paths-ignore:


benjaminwil (approved):
I wasn't following along on the new precompile assets workflow work, so thanks for explaining the different workflow run cases.

I feel like there may be an opportunity to rename or restructure these workflows in the future... at a big-picture level, I am not sure why precompiling assets should trigger a Docker deploy. It is only when I look closely at the workflows that it starts to make sense.

I think the more semantic root workflow would be the "release" named workflow. We need Docker deploys and precompiled assets for releases, and between releases we have the unstable release which is the head of the `master` branch.

dometto:
Merging this even though we still have flaky capybara tests, so further merges to fix the tests don't trigger unnecessary docker-deploy actions.

PR #2126, opened 2025-08-22
Add yarn.lock to precompile assets workflow

dometto:
I *think* we want to also commit changes to `yarn.lock` that result from running the precompile workflow. Sanity check appreciated :)


benjaminwil (approved):
I think this could be nice to have. 👍

Issue #2128, opened 2025-08-22
Rework workflows for asset compilation and docker deploy


dometto:
At the moment, a push to `master` will result in the `precompile-assets` workflow being called. This workflow in turn calls `docker-deploy`, which will build a docker image and deploy it to dockerhub. However, there are at least two issues with this:

* since `precompile-assets` can succeed independently of `docker-test`, this means `docker-deploy` is run even when `docker-test` fails.
* because of the way we are calling `docker-deploy` (as a command within `precompile-assets`), we do not get to see a status mark (green/succeeded or red/failed) for the `docker-deploy` workflow.

What we want ideally:

- push to master: precompile assets -> run tests
  * if tests successful: commit precompiled assets, docker deploy
  * if tests fail: don't commit precompiled assets, don't run docker deploy
- see if tests and deploy were succeeded in the checkmarks for the latest commit

Any help appreciated!

PR #2141, opened 2025-08-22
Improve asset compilation and docker deploy workflow

dometto:
See https://github.com/gollum/gollum/issues/2128

This PR:

* addresses a bug in the `precompile-assets` workflow, which meant that an asset compuilation commit was always generated and pushed, even when there was not difference in the assets.
* merges the `docker-test` and `docker-deploy` workflows
* calls the `precompile-assets` workflow as a requirement in the new `docker-test-deploy` workflow

Although it is not so intuitive that `precompile-assets` is called from within `docker-test-deploy`, it does achieve what we want:

- docker-test-deploy is called on PR, new tag, and push to `master` events
  * in all cases, tests against the dockerized instance will be run
  * only in case of a push to `master`: run `precompile-assets`, and push the new docker image after it has been built and tests are successful (this ensures the newly compiled assets commit is included in the new image).


dometto:
@benjaminwil this is still not ideal but maybe better? If you have a more 'semantic' solution in mind that we could implement easily that would be great, but if you think this is at least good enough we can go with this for the time being!

Btw, sorry for the sudden large amount of open PRs, that's a consequence of activating dependabot for version updates and not just for security alerts. :)

PR #2142, opened 2025-08-22
Improve docker deply and assets compilation workflows

dometto:
See https://github.com/gollum/gollum/issues/2128

This PR:

* addresses a bug in the `precompile-assets` workflow, which meant that an asset compuilation commit was always generated and pushed, even when there was not difference in the assets.
* merges the `docker-test` and `docker-deploy` workflows
* calls the `precompile-assets` workflow as a requirement in the new `docker-test-deploy` workflow

Although it is not so intuitive that `precompile-assets` is called from within `docker-test-deploy`, it does achieve what we want:

- docker-test-deploy is called on PR, new tag, and push to `master` events
  * in all cases, tests against the dockerized instance will be run
  * only in case of a push to `master`: run `precompile-assets`, and push the new docker image after it has been built and tests are successful (this ensures the newly compiled assets commit is included in the new image).


dometto:
Annoyingly, the new `docker-test-deploy` workflow won't run on this PR because it is not yet in `master`, but locally I can run it with https://github.com/nektos/act -- I get failures because of missing secrets, but at least the syntax should be ok.

benjaminwil (approved):
Looks great. Thanks for doing this thankless work!

